PR #132, opened 2014-08-11
Switching selectPlaylist's fallback chain to bandwidth over resolution

bparadie:
This PR is a proposal for changing the priority from selectPlaylist's fallback chain from to resolution over bandwidth to bandwidth over resolution.

This might be best explained by using a real life example. BigBucksBunny has two channels:
```
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=560000,RESOLUTION=560x320
chunklist_w1591926520_b560000.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=280000,RESOLUTION=280x160
chunklist_w1591926520_b280000.m3u8
```

One serves high bandwidth at 560x320, the other one serves low bandwidth at 280x160. The current implementation of selectPlaylist decides that smaller resolutions, i.e. 600x300 is closer to 280x160 than to 560x320 and therefore picks the low bandwidth channel even though the network conditions allow for high bandwidth presentation.

As a result we'll get this result for 600x300 using the 280000 bandwidth channel:

original small

instead of this one, which is of higher quality using the 560000 bandwidth channel:

flashls small

I am arguing that bandwidth should always take precedence over resolution when selecting playlists.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -282,7 +282,7 @@ videojs.Hls.prototype.selectPlaylist = function () {
   }
 
   // fallback chain of variants
-  return resolutionBestVariant || bandwidthBestVariant || sortedPlaylists[0];
+  return  bandwidthBestVariant || resolutionBestVariant || sortedPlaylists[0];
 };
 
 /**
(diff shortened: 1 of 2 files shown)


dmlap:
This was fixed via #243

PR #149, opened 2014-09-12
mute video according to setting

orcaman:
the HLS player ignores the muted setting. this commit fixes it.


gkatsev:
html5 tech mutes the video but the controls don't update. Flash tech doesn't mute the video: http://jsbin.com/hiyemiwaci

dmlap:
This appears to be an issue with the Flash tech. Refiled in the main video.js repo.

PR #189, opened 2014-11-20
remove event handlers from player before switching src

hc2p:
When you switch sources you get into trouble with still bound event-handlers. After switching src, on every "timeupdate" the fillBuffer-function gets called, pointing to a gone playlist.

[See this example](http://jsbin.com/jixizi) to highlight the issue. (Skip close to the end of the first video to make it switch to the next)

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -61,7 +61,7 @@ videojs.Hls.prototype.src = function(src) {
 
   // if there is already a source loaded, clean it up
   if (this.src_) {
-    this.resetSrc_();
+    this.reset_();
   }
 
   this.src_ = src;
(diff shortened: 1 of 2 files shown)


hc2p:
@dmlap yes. Following my instructions you end up with an `Uncaught TypeError: Cannot read property 'segments' of undefined` error

dmlap:
@hc2p ok, thanks. I'll re-try and see if I can trigger it.

dmlap:
@hc2p I dug into this and felt like we needed to change the lifecycle of the playlist loader to simplify the process a bit. I think the underlying issue in this PR is resolved via #223. Thank you very much for the PR! (and sorry it took so long for me to get around to reviewing it) If you try out v0.11.1 and see this issue still, please let us know.

hc2p:
@dmlap thanks a lot for looking into this. I'll give v0.11.1 a shot.

PR #190, opened 2014-11-27
fix Live stream will stop playing

dista:
If mediaSource.endOfStream() is called in http hls LIVE mode,
in https://github.com/videojs/video-js-swf/blob/8c03e650b9be9bccd9ecef9e7520def56cde1e5f/src/com/videojs/providers/HTTPVideoProvider.as

```
case "NetStream.Buffer.Empty":
                    // should not fire if ended/paused. issue #38
                    if(!_isPlaying){ return; }

                    // reaching the end of the buffer after endOfStream has been called means we've
                    // hit the end of the video
                    if (_ending) {
                        _ending = false;
                        _isPlaying = false;
                        _isPaused = true;
                        _hasEnded = true;
                        _model.broadcastEvent(new VideoPlaybackEvent(VideoPlaybackEvent.ON_STREAM_CLOSE, {info:e.info}));
                        _model.broadcastEventExternally(ExternalEventName.ON_PAUSE);
                        _model.broadcastEventExternally(ExternalEventName.ON_PLAYBACK_COMPLETE);

                        _startOffset = 0;
                        _pausedSeekValue = 0;
                        break;
                    }
```

It will issue ON_PAUSE and ON_PLAYBACK_COMPLETE event to this plugin, and eventually cause
player in bad state.

So for hls LIVE mode, endOfStream should not be called

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -828,6 +828,7 @@ videojs.Hls.translateMediaIndex = function(mediaIndex, original, update) {
     return videojs.Hls.setMediaIndexForLive(update) + 1;
   }
 
+  // sync
   return translatedMediaIndex;
 };
 
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -828,7 +828,6 @@ videojs.Hls.translateMediaIndex = function(mediaIndex, original, update) {
     return videojs.Hls.setMediaIndexForLive(update) + 1;
   }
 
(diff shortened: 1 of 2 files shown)


dmlap:
@gkatsev is going to manually merge this one to clean up the commit history a bit.

gkatsev:
Merged in via c54a20335133a2d8a1eef2cde80e04d77376c3df

PR #195, opened 2014-12-11
Use TARGETDURATION to identify live start point

bc-bbay:
Live start time is now calculated using 3x TARGETDURATION instead of a static 30 seconds.


dmlap:
@bc-bbay do you need help with the rebase?

gkatsev:
Merged in via 3a3037d02e0bb9e37d03f2a8451fcb8a6b138d7b

PR #198, opened 2015-01-06
Scale the initial bandwidth estimate

dmlap:
Multiply the bandwidth measurement from the master playlist by five to account for the higher ratio of latency to throughput for this request relative to segments. Using the bandwidth number directly was almost always resulting in very low initial bandwidth estimates and poorer quality startup than necessary. The scaling factor was obtained by testing a number of videos from a high throughput/low latency connection as an upper bound and the same connection throttled to "DSL" levels with Network Link Conditioner for the lower bound. Update the playlist switching simulator to apply initial bandwidth estimates and reduce the simulation duration a bit so that the early behavior is more visible.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -138,7 +138,14 @@ videojs.Hls.prototype.handleSourceOpen = function() {
     };
 
     oldMediaPlaylist = this.playlists.media();
-    this.setBandwidth(this.playlists);
+    // the bandwidth estimate for the first segment is based on round trip
+    // time for the master playlist. the master playlist is almost always
+    // tiny so the round trip time is dominated by latency and so the
+    // computed bandwidth is much lower than steady-state. to account for
+    // this, we scale the bandwidth estimate from the master playlist.
+    this.setBandwidth({
+      bandwidth: this.playlists.bandwidth * 5
+    });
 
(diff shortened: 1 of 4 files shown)


src/videojs-hls.js
```diff
@@ -138,7 +138,14 @@ videojs.Hls.prototype.handleSourceOpen = function() {
     };
 
     oldMediaPlaylist = this.playlists.media();
-    this.setBandwidth(this.playlists);
+    // the bandwidth estimate for the first segment is based on round trip
+    // time for the master playlist. the master playlist is almost always
+    // tiny so the round trip time is dominated by latency and so the
+    // computed bandwidth is much lower than steady-state. to account for
+    // this, we scale the bandwidth estimate from the master playlist.
+    this.setBandwidth({
```
gkatsev: does `setBandwidth` expect only an object with a bandwidth property?

dmlap: It expects roundTripTime, bandwidth, and bytesRecieved. BytesReceived are defaulted to zero if not supplied. roundTripTime isn't actually used anywhere and setBandwidth() is happy to set it to undefined.

gkatsev: So, basically it just cares about bandwidth. LGTM, then.

PR #199, opened 2015-01-07
Allow a user-provided initial bandwidth estimate

dmlap:
If the downstream developer has a better way of estimating bandwidth for the initial playlist, use that value.


dmlap:
Reviewed by @seniorflexdeveloper

Issue #202, opened 2015-01-16
EXT-X-KEY URI processing not to spec.


stevemayhew:
The spec ([HTTP Live Streaming - Section  3.1](https://tools.ietf.org/html/draft-pantos-http-live-streaming-13#section-3.1) says URI should be relative to the playlist:

>  "A URI in a Playlist, whether it is a URI line or part of a tag, MAY
>   be relative.  Relative URIs MUST be resolved against the URI of the
>   Playlist file that contains it."

Here is a very short sample m3u8 that demos the issue: http://flashls-test.s3.amazonaws.com/7_copy.m3u8

And here is a branch with a proposed fix:  https://github.com/stevemayhew/videojs-contrib-hls/tree/fix-keyurl

I'm happy to add unit tests and submit the pull request with a little guidance as to which specific JS modules/sample playlist I should add them (I'm guessing the m3u8_test.js?)

Thanks!

dmlap:
@stevemayhew nice find! Can you open up a PR so we can comment on the change more easily?

stevemayhew:
Sure.  Will do

stevemayhew:
@dmlap The first checkin failed jsHint, that is fixed.  The top of tree for the branch passes all the build tests (on my mac).   Here's the pull request.  Let me know if/where I should add tests:
https://github.com/videojs/videojs-contrib-hls/pull/203

Also I'm looking at an issue with video/audio sync for the same stream.  I'll try to extract the transport streams that cause the issue.

dmlap:
Fixed via https://github.com/videojs/videojs-contrib-hls/pull/203

PR #203, opened 2015-01-19
Fix for EXT-X-KEY URI not handling relative URL (Issue #202)

stevemayhew:
Here's a proposed fix for handling the URI if it's relative to the playlist (on the same level as segment).

All it needs is some unit test cases, the existing test cases pass.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -487,7 +487,7 @@ videojs.Hls.prototype.resolveSegmentUrl = function(segmentRelativeUrl) {
   } else {
     segmentUrl = resolveUrl(resolveUrl(this.src_, this.playlists.media().uri || ''), segmentRelativeUrl);
   }
-  return segmentUrl
+  return segmentUrl;
 };
 
 /*
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -474,13 +474,20 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
   }
 
   // resolve the segment URL relative to the playlist
-  if (this.playlists.media().uri === this.src_) {
-    segmentUri = resolveUrl(this.src_, segment.uri);
-  } else {
-    segmentUri = resolveUrl(resolveUrl(this.src_, this.playlists.media().uri || ''), segment.uri);
-  }
+  segmentUri = this.resolveSegmentUrl(segment.uri);
 
   this.loadSegment(segmentUri, offset);
+};
+
+videojs.Hls.prototype.resolveSegmentUrl = function(segmentRelativeUrl) {
+  var segmentUrl;
+    // resolve the segment URL relative to the playlist
+  if (this.playlists.media().uri === this.src_) {
+    segmentUrl = resolveUrl(this.src_, segmentRelativeUrl);
+  } else {
+    segmentUrl = resolveUrl(resolveUrl(this.src_, this.playlists.media().uri || ''), segmentRelativeUrl);
+  }
+  return segmentUrl
 };
 
 /*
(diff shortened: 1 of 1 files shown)


stevemayhew:
The build is not failing with my last checkin?  Why does this message persist?  Is there something else I need to do.

gkatsev:
Looks like there's a test error now https://travis-ci.org/videojs/videojs-contrib-hls/builds/47580832
The issue is that on travis we run phantomjs whereas locally, we use karma to run it in chrome.

stevemayhew:
Ah, that is this one:


 <>Running "karma:phantomjs" (karma) task
 <>PhantomJS 1.9.8 (Linux) ERROR
 <>	TypeError: 'null' is not an object (evaluating 'toolbar.appendChild')
 <>	at /home/travis/build/videojs/videojs-contrib-hls/node_modules/qunitjs/qunit/qunit.js:2534

Looks like I need to fire up my linux VM?  We have had lots of trouble with PhantomJS tests, particularly on Windows.

Gary, 
I am working on a video/audio sync issue now too…  Will put up the bug and sample playlist shortly.
Can you please connect me the person who manages the VideoJS effort at Bightcove.  The company I am working with is in the end game of evaluating video playback tech for HLS solutions.  There are a couple of flash only solutions as well as VisualOn in the running.

I’m a VideoJS fan though, because:
1) It follows HTML5 (so it falls back to native Safari for example)
2) lack of need for plugins is huge for us (FlashPlayer is really part of the browser now)
3) Much nicer to work in JS then AS3 (IMNSHO)

Would love to contribute materially to the project, but my management needs to get the warm and fuzzy to narrow in the bet on the winning horse.

You have heard of the company, they are a household name, a bit of Googling and you’ll figure it out ;-).

PS hope you guys are digging out ok out there!

Best Regards,
Steve

Steve Mayhew
Skype: stevemayhew
Twitter: stevemayhew
Mobile: 650-619-2529

stevemayhew:
Ah, that is this one:

 <>Running "karma:phantomjs" (karma) task
 <>PhantomJS 1.9.8 (Linux) ERROR
 <>	TypeError: 'null' is not an object (evaluating 'toolbar.appendChild')
 <>	at /home/travis/build/videojs/videojs-contrib-hls/node_modules/qunitjs/qunit/qunit.js:2534

Looks like I need to fire up my linux VM?  We have had lots of trouble with PhantomJS tests, particularly on Windows.

I'll try and see if I can reproduce it.

src/videojs-hls.js
```diff
@@ -474,13 +474,20 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
   }
 
   // resolve the segment URL relative to the playlist
+  segmentUri = this.resolveSegmentUrl(segment.uri);
+
+  this.loadSegment(segmentUri, offset);
+};
+
+videojs.Hls.prototype.resolveSegmentUrl = function(segmentRelativeUrl) {
```
dmlap: I'm wondering if there's a better name for this method now that we'll be using it for more than just segment URL resolution.

stevemayhew: Yeah, me too.  How about:
1. resolvePlaylistRelativeUri
2. playlistUriToUrl
3. resolveUrl (might be hard to search for vs similar names)

Personally I like option 2 best, we are resolving a URI in the playlist (which may be a URL) to a URL.  PS, for people that always confuse the two (like me ;-)).   Here's a simple non-geekly explanation of the difference that is pretty accurate [URL vs URI](https://danielmiessler.com/study/url_vs_uri/)

dmlap:
Merged via 66510e67aad74474b6ca44fb98d580f1ee008c63.

@stevemayhew had a bit of free time so I added unit tests. Let me know if you don't think they cover the case we were mishandling properly. Thank you for the contribution!

PR #205, opened 2015-01-22
Fixing discontinuity seek

bc-bbay:
On seeking to a discontinuity point, the buffer would not refill and
the player would stop working. This commit fixes that issue, and
adjusts the test to catch the previous failure.

### DIFF
--- a/test/videojs-hls_test.js
+++ b/test/videojs-hls_test.js
@@ -1186,7 +1186,7 @@ test('waits until the buffer is empty before appending bytes at a discontinuity'
   var aborts = 0, setTime, currentTime, bufferEnd;
 
   player.src({
-    src: 'disc.m3u8',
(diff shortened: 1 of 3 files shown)


test/videojs-hls_test.js
```diff
@@ -1261,14 +1261,14 @@ test('clears the segment buffer on seek', function() {
   standardXHRResponse(requests.pop());
 
   // play to 6s to trigger the next segment request
-  currentTime = 6;
+  currentTime = 1;
```
dmlap: The comment above this line is now incorrect.

dmlap:
I think we should have two test-cases for these scenarios: the original test which seeked after playing through some portion of the video and your updated test that seeks directly to a discontinuity. Is there a reason the old test is no long valuable I'm missing?

Issue #211, opened 2015-02-11
h264 / AAC streams not synchronized


ronhopper:
We are attempting to play streams where the audio / video streams start at slightly different times. The playback causes an audio lag problem. I logged the first PTS from each stream (h.264 and AAC) and noticed that they differ by about 760ms.

    DEBUG: aac-stream starting pts_offset = 39399181.35555556
    DEBUG: h264-stream starting pts_offset = 39399940.35555556

Since the setNextTimeStamp function in h264-stream.js and aac-stream.js are recording different offsets on the first call, it seems to maintain the 760ms lag throughout playback.

I modified the code to ignore the pts_offset (just ignore the stream info and set the offset to zero) and then playback was synchronized.

It seems to me that the correct answer would be to use the same offset for both streams, rather than calculate them independently, but I don't fully understand the implications of that.

Here is an example playlist for reference:
https://dev-recordings.boxcast.com/recordings/20140319-202927-scaqvitikqvzctev/all.m3u8

ronhopper:
FYI -- I'd be happy to implement and submit a pull request, if someone can confirm desired behavior. Thanks.

dmlap:
That is an interesting problem. It does sound like the current behavior is not correct to me. If you'd be willing to take a shot at a PR, that would be great!

dmlap:
Merged into master now, will go out with the next update.

dmlap:
For future reference: fixed via d94275ef8cf27ae6d227d20ccb9abb9e9cb499ae

PR #213, opened 2015-02-12
Synchronize timestamp offsets between h.264 and AAC streams

ronhopper:
Solves Issue https://github.com/videojs/videojs-contrib-hls/issues/211

When the audio and video streams start at slightly different times, the existing code captures different PTS offsets. This causes an audio lag problem during playback.

In this pull request, rather than capturing the PTS offset independently in the AAC and h.264 streams, the first stream within the TS segment initializes the offset for both streams, which is then set via the new function `setTimeStampOffset`. This is clearer than using self-modifying functions within each stream independently and synchronizes audio/video streams that do not share the same absolute starting time.


ntadej:
This fixes audio-video sync with our streams without any noticeable regressions.

heff:
@szatmary, any chance we could get you to take a quick look at this and verify that it's a safe change?

szatmary:
I see nothing obviously wrong

dmlap:
Merged via d94275ef8cf27ae6d227d20ccb9abb9e9cb499ae

PR #214, opened 2015-02-12
Add MP3 audio support

rdgRmCore:
Based on the previous work of @raytiley.

### DIFF
--- a/test/flv-tag_test.js
+++ b/test/flv-tag_test.js
@@ -61,14 +61,14 @@ test('audio type handled correctly', function(){
   var mp3Tag = new FlvTag(FlvTag.AUDIO_TAG, undefined, FlvTag.MP3_AUDIO), 
       aacTag = new FlvTag(FlvTag.AUDIO_TAG, undefined, FlvTag.AAC_AUDIO); 
 
-  equal(12, mp3Tag.length, 'mp3 tag length is 12');
-  equal(13, aacTag.length, 'aac tag length is 13');
+  equal(mp3Tag.length, 12, 'mp3 tag length is 12');
+  equal(aacTag.length, 13, 'aac tag length is 13');
 
   mp3Tag.finalize();
   aacTag.finalize();
   equal(mp3Tag.bytes[11], 0x2F, 'mp3 is set to 44khz 16 bit stereo');
   equal(aacTag.bytes[11], 0xAF, 'aac is set to 44khz 16 bit stereo');
-  equal(aacTag.bytes[12], 0x01, 'aac extra data is 0x00');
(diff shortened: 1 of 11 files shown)


gkatsev:
Thanks! We'll get it reviewed when we get the change.
Any chance you could add some tests that test this new functionality? That would be awesome and help this get merged in quicker.

rdgRmCore:
@gkatsev 

I will give testing some thought.

Right now I have tested it manually with an actual transport stream.

I would have to come up to speed on your unit testing framework.

rdgRmCore:
@gkatsev 

I added a unit test.

test/flv-tag_test.js
```diff
@@ -57,4 +57,18 @@ test('writeBytes grows the internal byte array dynamically', function() {
   }
 });
 
+test('audio type handled correctly', function(){
+  var mp3Tag = new FlvTag(FlvTag.AUDIO_TAG, undefined, FlvTag.MP3_AUDIO), 
+      aacTag = new FlvTag(FlvTag.AUDIO_TAG, undefined, FlvTag.AAC_AUDIO); 
+
+  equal(12, mp3Tag.length, 'mp3 tag length is 12');
```
gkatsev: The expected value should be the second item with the actual value being the first.

rdgRmCore:
@gkatsev Is there any news or updates regarding this pull request?

forbesjo:
We've done a lot of refactoring, could you open a PR against mux.js? Sorry about taking so long to get back to you.

JacobRichardt:
Did this ever get merged? I'm having some trouble getting mp3 audio to play.

GerardVirga:
We are seeing this same problem with live streams published from FMLE: H264 video with MP3 audio (44htz)  to Wowza 4.6.x. When the live video streams from FMLE include MP3 audio there is no sound. When published with AAC from FFMPEG or Wirecast the sound renders. We understand that this is a legacy application but want to head off any future support issues if FMLE or an MP3 codec is used in the future. 

The he playlist shows MP3 as as mp4a.40.34. Which should be right? I think this ticket deals with Audio only but it sounds like a variation of the same problem. I also think the same problem exists in HLS.js when we test with the demo player. 

We are still learning the framework and protocol for bugs so forgive us if we missed something posted in the wrong place.

GerardVirga:
I also wanted to add that the use case here is less FMLE but more important for users of vMix or older Tricasters which still automate FMLE behind the scene.

anthropos9:
Has there been any progress on this issue? 

I have a project that is pulling an m4u8 playlist with mp3s from a WowzaStreamingEngine/4.7.0 server, it reads the playlist, but when it tries to play an individual mp3 it just hangs.

Any update would be helpful. Thanks.

PR #215, opened 2015-02-13
Allow override of automatic variant playlist selection logic

ronhopper:
In order to provide a quality selector button, there must be some way for an external plugin to choose between automatic variant playlist selection (the current behavior) and manual selection.

This PR provides a method (`hls.setVariant`) to forcibly override the default behavior.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -356,12 +356,22 @@ videojs.Hls.prototype.dispose = function() {
 };
 
 /**
+ * Sets the variant playlist that should be used.
+ * @param variant (optional) {int or string} the index or key of the variant to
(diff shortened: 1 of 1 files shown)


dmlap:
Our intent was that if you wanted to provide custom playlist selection logic, you could just override selectPlaylist():

```js
  player.hls.selectPlaylist = function() {
    return userSelectedVariantPlaylist;
  };
```

Is there a reason I'm not thinking of why that wouldn't work?

ronhopper:
That sound's reasonable. Let me try it that way and, assuming it works, I'll close this out.

ronhopper:
I was able to do this the way that you suggested. Thanks!

dmlap:
No problem. Glad it worked for you!

PR #216, opened 2015-02-23
Support custom-IV when decrypting segments

rajkosto:
The m3u8 parser was parsing the IV attribute, but didn't account for the case when they are prefixed by 0x, this fixes that, and also uses the IV if it's present
With this change, nginx-rtmp aes encrypted hls streams work.

### DIFF
--- a/src/m3u8/m3u8-parser.js
+++ b/src/m3u8/m3u8-parser.js
@@ -442,7 +442,7 @@
                 method: entry.attributes.METHOD || 'AES-128',
                 uri: entry.attributes.URI
               };
-	      
+
               if (entry.attributes.IV !== undefined) {
                 key.iv = entry.attributes.IV;
               }
--- a/src/m3u8/m3u8-parser.js
+++ b/src/m3u8/m3u8-parser.js
@@ -440,9 +440,12 @@
               // setup an encryption key for upcoming segments
               key = {
                 method: entry.attributes.METHOD || 'AES-128',
-                uri: entry.attributes.URI,
-                iv: entry.attributes.IV || null
+                uri: entry.attributes.URI
               };
+	      
+	      if (entry.attributes.IV !== undefined) {
+	      	key.iv = entry.attributes.IV
+	      }
             },
             'media-sequence': function() {
               if (!isFinite(entry.number)) {
--- a/src/m3u8/m3u8-parser.js
+++ b/src/m3u8/m3u8-parser.js
@@ -305,8 +305,9 @@
         event.attributes = parseAttributes(match[1]);
         // parse the IV string into a Uint32Array
         if (event.attributes.IV) {
-          if (event.attributes.IV.substring(0,2) == '0x')
+          if (event.attributes.IV.substring(0,2) === '0x') {	
             event.attributes.IV = event.attributes.IV.substring(2);
+          }
 
           event.attributes.IV = event.attributes.IV.match(/.{8}/g);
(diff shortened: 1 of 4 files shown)


test/manifest/encrypted.json
```diff
@@ -30,11 +33,21 @@
...
+        "iv": null
       },
       "uri": "http://media.example.com/fileSequence53-A.ts"
     },
     {
+      "duration": 14,
+      "key": {
+        "method": "AES-128",
+        "uri": "https://priv.example.com/key.php?r=54",
+        "iv": new Uint32Array([0, 0, 331, 3063767524])
```
dmlap: I'm not completely sure why this text is highlighted in red in github. A text encoding issue, maybe?

gkatsev: This is a JSON file. You can't have js expressions in it.

src/m3u8/m3u8-parser.js
```diff
@@ -436,7 +440,8 @@
               // setup an encryption key for upcoming segments
               key = {
                 method: entry.attributes.METHOD || 'AES-128',
-                uri: entry.attributes.URI
+                uri: entry.attributes.URI,
+                iv: entry.attributes.IV || null
```
dmlap: For most of the other properties in this parser, we haven't added key/values for attributes that weren't present in the m3u8.

rajkosto: How would you have a Uint32Array in json then ? I just made changes untill it passed all your automated tests, change it to fit your style however you like.

test/manifest/encrypted.json
rajkosto: How would you have a Uint32Array in json then ? I just made changes untill it passed all your automated tests, change it to fit your style however you like.

src/m3u8/m3u8-parser.js
rajkosto: entry.attributes.IV is there and was working, other than the whole 0x prefix thing, but it needs to propagate to the decryptor somehow.

dmlap: @rajkosto this is a great place for the key when it's present in the manifest. If it's not present though, we've been signaling that by not including the property name. Could you update this so that `entry.attributes.IV` is only set if a custom IV is present in the m3u8?

test/manifest/encrypted.json
dmlap: Ok, that makes sense. Any reason you went with `Uint32Array` instead of `Uint8Array` here? I'd be cool with pulling this in if you could change the file extension of the examples to `.js`. We'll just have to be strict about the sort of expressions we let in here in the future.

rajkosto:
OK how about instead of having someone who is unfamiliar with JS and your build process massage it into whatever code style guidelines you have, just do that yourselves ?

test/manifest/encrypted.json
rajkosto: I just made it pass your test, if you want, you can remove the iv test completely, but then you have a missing test. The way you test is deep-comparing objects parsed from the m3u8 with the ones from the json file, i cannot nor should i rename the json file. I don't know of how you would have a byte/int array in a json file with it remaining JSON compliant either. You usually base64 encode binary data in JSON but then the deep-compare would fail.

dmlap:
I want to go through and rename our expected parsed output from .json to .js so that we don't confuse anyone who takes a look over the test cases later. Other than that, this looks great. Thanks!

dmlap:
Merged. @rajkosto thanks again for the patch!

PR #221, opened 2015-03-02
In-Band Metadata

dmlap:
HTTP Live Streaming supports inclusion of [timed metadata](https://developer.apple.com/library/ios/documentation/AudioVideo/Conceptual/HTTP_Live_Streaming_Metadata_Spec/HTTP_Live_Streaming_Metadata_Spec.pdf) in [ID3 format](http://id3.org/id3v2.3.0), carried in the MPEG-2 transport stream. Expose that metadata as cues on an [in-band metadata text track](https://html.spec.whatwg.org/multipage/embedded-content.html#texttrack).


src/videojs-hls.js
```diff
@@ -74,6 +74,38 @@ videojs.Hls.prototype.src = function(src) {
...
+      metadataStream = tech.segmentParser_.metadataStream,
+      textTrack;
+
+    metadataStream.on('data', function(metadata) {
+      var i, frame, time, hexDigit;
+
+      // create the metadata track if this is the first ID3 tag we've
+      // seen
+      if (!textTrack) {
+        textTrack = tech.player().addTextTrack('metadata', 'Timed Metadata');
```
gkatsev: would we want to make sure this method exists before using? This would mean that hls will no longer be usable with versions of vjs below 4.12.

src/videojs-hls.js
```diff
@@ -74,6 +74,38 @@ videojs.Hls.prototype.src = function(src) {
...
+
+        // build the dispatch type from the stream descriptor
+        // https://html.spec.whatwg.org/multipage/embedded-content.html#steps-to-expose-a-media-resource-specific-text-track
+        textTrack.inBandMetadataTrackDispatchType = videojs.Hls.SegmentParser.STREAM_TYPES.metadata.toString(16);
```
gkatsev: FYI, emulated tracks currently ignore this property.

src/videojs-hls.js
```diff
@@ -74,6 +74,38 @@ videojs.Hls.prototype.src = function(src) {
...
+        for (i = 0; i < metadataStream.descriptor.length; i++) {
+          hexDigit = ('00' + metadataStream.descriptor[i].toString(16)).slice(-2);
+          textTrack.inBandMetadataTrackDispatchType += hexDigit;
+        }
+      }
+
+      for (i = 0; i < metadata.frames.length; i++) {
+        frame = metadata.frames[i];
+        time = metadata.pts / 1000;
+        textTrack.addCue(new window.VTTCue(time, time, frame.value || frame.url));
```
gkatsev: I was going to mention that on iOS6 or 7, we only have TextTrackCue and not VTTCue but then I realized this doesn't apply to iOS.

src/videojs-hls.js
dmlap: From my reading, it's entirely informational at this point. I included it just for completeness sake.

src/videojs-hls.js
dmlap: VTTCue is the only concrete implementation of TextTrackCue that I'm aware of. I would have preferred using a more generic cue type for this but I don't think that's supported anywhere.

src/videojs-hls.js
gkatsev: Yeah, it seemed that way. One of the reasons why I chose not to implement it. Just wanted to call it out directly.

src/videojs-hls.js
gkatsev: VTTCue is what we should be using unless we're on that specific version of iOS.

dmlap: Ok, cool.

src/videojs-hls.js
dmlap: Added a feature detect before the metadata event handler a couple lines up.

gkatsev:
LGTM

src/videojs-hls.js
```diff
@@ -74,6 +74,44 @@ videojs.Hls.prototype.src = function(src) {
...
+      // create the metadata track if this is the first ID3 tag we've
+      // seen
+      if (!textTrack) {
+        textTrack = tech.player().addTextTrack('metadata', 'Timed Metadata');
+
+        // build the dispatch type from the stream descriptor
+        // https://html.spec.whatwg.org/multipage/embedded-content.html#steps-to-expose-a-media-resource-specific-text-track
+        textTrack.inBandMetadataTrackDispatchType = videojs.Hls.SegmentParser.STREAM_TYPES.metadata.toString(16);
+        for (i = 0; i < metadataStream.descriptor.length; i++) {
+          hexDigit = ('00' + metadataStream.descriptor[i].toString(16)).slice(-2);
```
gkatsev: Apparently, it needs to be uppercase ascii

PR #223, opened 2015-03-11
Clean up the playlist loader when sources change

### LINKED ISSUE #133: paused videos stop buffering
The `fillBuffer` function is called on the `timeupdate` event. Since `timeupdate` doesn't fire when the video is paused, pausing causes the player to stop further buffering of data.

### LINKED ISSUE #160: timeupdate event is not triggered while a video is paused
Because the current time is set in `drainBuffer` and it doesn't trigger `timeupdate` event, this event is not triggered when a video is paused. If the video is playing, the callback of setInterval triggers `timeupdate` every 250 millisecond so it won't be a big problem.
```js
  if (segment.discontinuity) {
    if (event.type !== 'waiting') {
      return;
    }
    this.sourceBuffer.abort();
    // tell the SWF where playback is continuing in the stitched timeline
    this.el().vjs_setProperty('currentTime', segmentOffset * 0.001);
  }
```

Therefore I would like to trigger `timeupdate` right after `vjs_setProperty` is called, however, it will be an infinite loop because `drainBuffer` is one of a callback for `timeupdate`.

So I came up with an idea below.
Like `manualTimeupdatesOn` in the original `video.js`, implement `setInterval` to trigger `timeupdate` in `videojs-hls` and call all callbacks including `drainBuffer` in the interval. And also set `featuresTimeupdateEvents` to false to avoid original `manualTimeupdatesOn` from being called.
```js
videojs.Hls.prototype.trackCurrentTime = function(){
  if (this.currentTimeInterval) { this.stopTrackingCurrentTime(); }
  this.currentTimeInterval = setInterval(vjs.bind(this, function(){
    this.player().trigger('timeupdate');
    this.fillBuffer();
    this.drainBuffer();
  }), 250);
};
```
Then `timeupdate` can be triggered right after `vjs_setProperty` is called in `drainBuffer`.
But I feel this solution is a bit redundant...

Could you tell me your opinion?

dmlap:
Make sure the playlist loader is immediately disposed when the source is changed instead of waiting for sourceopen. Previously, it was possible that the source would be changed and fillBuffer() would run before the new set of playlists had been loaded. In that case, HLS would download a segment from the old playlist and block drainBuffer() from making progress because playlist.segments in the segmentBuffer element would be undefined. Switch from using timeupdate to trigger buffering to using straight window.timeout which fixes #160 and fixes #133.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -10,6 +10,9 @@ var
   // a fudge factor to apply to advertised playlist bitrates to account for
   // temporary flucations in client bandwidth
   bandwidthVariance = 1.1,
+
+  // the amount of time to wait between checking the state of the buffer
+  bufferCheckInterval = 500,
   keyXhr,
   keyFailed,
   resolveUrl;
@@ -36,6 +39,11 @@ videojs.Hls = videojs.Flash.extend({
     this.currentTime = videojs.Hls.prototype.currentTime;
     this.setCurrentTime = videojs.Hls.prototype.setCurrentTime;
 
+    // periodically check if new data needs to be downloaded or
+    // buffered data should be appended to the source buffer
+    this.segmentBuffer_ = [];
+    this.startCheckingBuffer_();
+
     videojs.Hls.prototype.src.call(this, options.source && options.source.src);
   }
 });
@@ -49,7 +57,10 @@ videojs.Hls.GOAL_BUFFER_LENGTH = 30;
 videojs.Hls.prototype.src = function(src) {
   var
     tech = this,
+    player = this.player(),
+    settings = player.options().hls || {},
     mediaSource,
+    oldMediaPlaylist,
     source;
 
   // do nothing if the src is falsey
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -10,6 +10,9 @@ var
   // a fudge factor to apply to advertised playlist bitrates to account for
   // temporary flucations in client bandwidth
   bandwidthVariance = 1.1,
+
+  // the amount of time to wait between checking the state of the buffer
+  bufferCheckInterval = 500,
```
gkatsev: why 500?
This PR unhooks listening to timeupdate to run fillbuffer right? That used to run at either 250 or less.

dmlap: I was thinking 250ms was a bit more frequent than we actually needed, given we call fillBuffer() explicitly whenever we take an action that has a big effect on the buffer.

gkatsev:
Does this actually fix #160? I don't see any new triggering of `timeupdate`.

src/videojs-hls.js
```diff
@@ -377,12 +384,7 @@ videojs.Hls.prototype.cancelSegmentXhr = function() {
  * Abort all outstanding work and cleanup.
  */
 videojs.Hls.prototype.dispose = function() {
-  var player = this.player();
-
-  // remove event handlers
-  player.off('timeupdate', this.fillBuffer);
-  player.off('timeupdate', this.drainBuffer);
-  player.off('waiting', this.drainBuffer);
+  this.stopCheckingBuffer_();
```
gkatsev: Are there any other cases where we would want to call `stopCheckingBuffer`?

dmlap:
Ah, I guess you're right this wouldn't address firing a timeupdate, though it does fix the fact that seeking while paused wouldn't trigger buffering immediately. Should we address the timeupdate-on-seek issue in the Flash tech?

gkatsev:
It definitely makes it possible for us to fire timeupdates now and not worry about infinite loops.

gkatsev:
Assuming the flash tech won't break if we fire a timeupdate from it on seek that's fine, though, we still would need to fire a timeupdate at a discontinuity in this project.

gkatsev:
Do we want a test for checking that `startCheckingBuffer_` is called on init and `stopCheckingBuffer_` is called on dispose?

src/videojs-hls.js
dmlap: We could try to be more clever about when buffer checking was in effect but trying to be clever *was* the reason for these issues in the first place. This iteration does some extra work when no video is loaded or buffering has completed but it makes the tech lifecycle more consistent. The extra checks don't even show up in the profiler for me so I feel like this is reasonable.

gkatsev: Definitely not talking about "clever" places to call it but rather if there were other obvious places aside from `dispose` that we would want to stop buffering. I can't really think of any.

dmlap:
@gkatsev I added a test case to show that buffer checking happens on init and we do have a test that checks event handlers on dispose(). Sound good?

gkatsev:
That should check that waiting is removed but nothing is verifying that the timeout has been cleared.
Otherwise, LGTM.

gkatsev:
(Also, make sure that #160 stays open when this gets merged in)

gkatsev:
LGTM.

PR #224, opened 2015-03-12
added prepublish hook to have a concatenated version after install

quarkus:
just added a prepublish hook to have a local concatenated / minified version available after install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
 - '0.10'
-before_script:
-- npm install -g grunt-cli
+install:
+- npm install -g grunt-cli && npm install
 - if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then curl https://gist.githubusercontent.com/santiycr/5139565/raw/sauce_connect_setup.sh | bash; fi
 notifications:
   hipchat:
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,7 @@ dist/*
 /node_modules/
 *~
 *.iml
+*.ipr
+*.iws
 *.swp
 tmp/**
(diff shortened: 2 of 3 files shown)


dmlap:
Good thinking! I have no idea why travis is trying to run prepublish as part of the CI build, though.

gkatsev:
I believe it's a bug with npm. We probably want to update our travis config to use a newer version of npm as well as using newer versions of node/iojs.

gkatsev:
A workaround is to change the `"before_script"` property in `.travis.yml` to be `install` and append `npm install` to it.
```yaml
before_script:
- npm install -g grunt-cli
```
should turn into:
```yaml
install:
- npm install -g grunt-cli && npm install
```
See my changes here: https://github.com/videojs/videojs-contrib-ads/pull/72

gkatsev:
@quarkus hey, want to make the change to the `.travis.yml` file?

quarkus:
@gkatsev sorry .. i have been pretty busy over the last days. is the change correct ? i am not totally sure what i´m doing ere ;)

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
 - '0.10'
-before_script:
-- npm install -g grunt-cli
+install:
+- npm install -g grunt-cli && npm install
```
gkatsev: perfect.

gkatsev:
LGTM

gkatsev:
Also, I just realized there's a better way of doing it. Instead of changing the `install` script in the `travis.yml`, we can just declare `grunt-cli` as a devDependency.

dmlap:
@quarkus just tried this out with the 0.13 release and it worked like a charm. Thank you!

PR #225, opened 2015-03-13
Validate existence of h264Frame where needed

ntadej:
Sometimes HLS crashes due to not existing h264Frame. Validate it before writing data to it and it plays/works properly afterwards.

### DIFF
--- a/src/h264-stream.js
+++ b/src/h264-stream.js
@@ -192,7 +192,7 @@
         } else {
           // If we get here we have 00 00 00 or 00 00 01
           if (data[offset + 2] === 1) {
-            if (offset > start) {
+            if (this._h264Frame && offset > start) {
               this._h264Frame.writeBytes(data, start, offset - start);
             }
             this._state = 3;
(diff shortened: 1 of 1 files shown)


dmlap:
@ntadej thanks for the PR! I think in this case that something is going wrong at another point in the parsing process that we should probably get to the bottom of. Do you have a reduced test case available that causes this to occur?

ntadej:
Our streams do not start properly with your HLS implementation do to missing ADTS headers at the beginning. This may cause the issues, but our streams play completely fine on iOS, Android and other devices.

We are preparing a test stream for you (I plan to open another bug, as there are some problems with quality switching and general playback after 0.11.1 -> 0.11.2 update). Is there a preferred email to send the URL to you, as we can't make it public at the moment?

dmlap:
@ntadej email me at dlapalomento at brightcove dot com.

ntadej:
I've sent you the url on email and opened #236. Note that playback on 0.11.2 works again after updating to Video.js 4.12.5 but quality switching still has problems as described in #236.

ntadej:
We investigated why this is necessary and found out that H264 data is not aligned inside the TS packet. It seems to be a valid case to have this `h264Frame` validation and data before first frame can be ignored.

forbesjo:
We think this is fixed with HLS 1.0, could you test this in the development branch?

PR #229, opened 2015-03-16
Live start time tweaks

bc-bbay:
Moving live start time calculation earlier in the code flow to eliminate a race condition, repeating it on first play to create a smooth experience, and making sure currentTime is updated to match these changes.


src/videojs-hls.js
```diff
@@ -239,7 +244,14 @@ videojs.Hls.prototype.src = function(src) {
   });
 };
 
-videojs.Hls.setMediaIndexForLive = function(selectedPlaylist) {
+/* Returns the media index for the live point in the current playlist, and updates
+   the current time to go along with it.
+ */
+videojs.Hls.getMediaIndexForLive = function(selectedPlaylist) {
+  if (!selectedPlaylist.segments) {
```
dmlap: Should we be calling this function with a half-loaded playlist now?

src/videojs-hls.js
```diff
@@ -279,6 +291,11 @@ videojs.Hls.prototype.play = function() {
     this.mediaIndex = 0;
   }
 
+  if (this.playlists.media() && !this.el().classList.contains('vjs-has-started')) {
```
dmlap: We don't support IE8 in this project but you should get into the habit of using `this.hasClass()` for maximum compatibility.

gkatsev: I think classList actually exists in all the browsers we care to support by this project.

src/videojs-hls.js
```diff
@@ -973,6 +990,20 @@ videojs.Hls.getMediaIndexByTime = function(playlist, time) {
   return -1;
 };
 
+videojs.Hls.prototype.setCurrentTimeByMediaIndex = function(playlist, mediaIndex) {
```
dmlap: Please document this function. Also, we should denote it as private (append an `_`) unless there's a good reason it should be exposed.

dmlap: Also, "set" in the name implies some sort of mutation to me. I like side-effect free but the name should change.

src/videojs-hls.js
dmlap: For consistency with the rest of video.js, I think it's still better to use hasClass(). Plus, it's actually less typing here.

src/videojs-hls.js
bc-bbay: Never in production, in theory, but we do it all the time in the unit tests. This is the breakage I was seeing after my first two commits.

src/videojs-hls.js
gkatsev: Yep, consistency is good.
I'd love to move everything to classList, though.

bc-bbay: Will update, thanks.

test/videojs-hls_test.js
```diff
@@ -1720,6 +1739,11 @@ test('calling play() at the end of a video resets the media index', function() {
   player.hls.ended = function() {
     return true;
   };
+
+  player.hasClass = function(classname) {
```
dmlap: Why is this necessary?

src/videojs-hls.js
dmlap: Should those tests be invoking this method without segments?

dmlap:
@bc-bbay can you append an `_` to the methods added in this commit? I want to be clear they're not intended for external use.

src/videojs-hls.js
bc-bbay: Yeah, in some cases, that's the point.

test/videojs-hls_test.js
bc-bbay: hasClass isn't defined in this scope, but you recommended using it in a previous PR review, so this addition keeps the test working.

bc-bbay:
Will do.

PR #230, opened 2015-03-19
Bitrate switching docs

dmlap:
Expand on bitrate switching documentation.


README.md
```diff
@@ -40,6 +40,10 @@ HLS stream, code against the regular HTML5 video APIs, and create a
 fast, high-quality video experience across all the big web device
 categories.
 
+Check out the [full documentation](docs/) for details on how HLS works
```
bclwhitaker: just links to docs folder?

gkatsev:
Nice art :)
LGTM

README.md
dmlap: We should make an index page of some sort for docs but my main point with this PR was to expand on the adaptive switching info.

docs/bitrate-switching.md
```diff
@@ -0,0 +1,40 @@
...
+
+If you're the visual type, the whole process is illustrated
+below. Whenever a new segment is downloaded, we calculate the download
+bitrate based on the size of the segment and the time it took to
+download:
+
+![New bitrate info is available](bitrate-switching-1.png)
+
+First, we filter out all the renditions that have a higher bitrate
+than the new measurement:
```
gquinones: We should never discard the lowest available rendition based on bitrate. Otherwise we're left with no renditions to play at all. The question is: which is better, playing a rendition that will buffer due to bandwidth limits, or not play at all? In my opinion it is better to at least try to play because bandwith limits could be temporary or they could be caused by our bitrate sampling limitations.

gkatsev: It should theoretically be re-added via the player sizing, but yeah, I agree.

dmlap: "Always be playing" is my motto :) I skipped over it for simplicity sake in this write-up but we will use the first listed encoding if all the available encodings are filtered out by other parts of the process.

README.md
gkatsev: Given that we only have a few articles, I feel like that ToC should live in the README.

docs/bitrate-switching.md
gquinones: Choose the simplest description that explains the process, but no simpler. :)

dmlap: Added a line at the end to clarify the worst-case scenario.

README.md
dmlap: That's a good point. There's a bunch in the mp2t-to-mp4 branch that I'd like to move into master someday...

PR #231, opened 2015-03-20
Fixing bug that started playback at 2nd segment

bc-bbay:
Bugfix and unit test

### DIFF
--- a/test/videojs-hls_test.js
+++ b/test/videojs-hls_test.js
@@ -1754,7 +1754,12 @@ test('drainBuffer will not proceed with empty source buffer', function() {
   player.hls.segmentBuffer_ = [{mediaIndex: 0, playlist: newMedia, offset: 0, bytes: [0,0,0]}];
 
   player.hls.drainBuffer();
-  deepEqual(player.hls.segmentBuffer_,compareBuffer,'playlist remains unchanged');
+
+  /* Normally, drainBuffer() calls segmentBuffer.shift(), removing a segment from the stack.
+   * Comparing two buffers to ensure no segment was popped verifies that we returned early
+   * from drainBuffer() because sourceBuffer was empty.
+   */
+  deepEqual(player.hls.segmentBuffer_, compareBuffer, 'playlist remains unchanged');
 
   player.hls.playlists.media = oldMedia;
 });
--- a/test/videojs-hls_test.js
+++ b/test/videojs-hls_test.js
@@ -1724,6 +1724,41 @@ test('calling play() at the end of a video resets the media index', function() {
   strictEqual(player.hls.mediaIndex, 0, 'index is 1 after the first segment');
 });
 
+test('drainBuffer will not proceed with empty source buffer', function() {
+  var oldMedia, newMedia, compareBuffer;
+  player.src({
+    src: 'https://example.com/encrypted-media.m3u8',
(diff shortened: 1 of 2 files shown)


bclwhitaker:
:+1:

test/videojs-hls_test.js
```diff
@@ -1724,6 +1724,41 @@ test('calling play() at the end of a video resets the media index', function() {
...
+  var oldMedia, newMedia, compareBuffer;
+  player.src({
+    src: 'https://example.com/encrypted-media.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(player);
+
+  oldMedia = player.hls.playlists.media;
+  newMedia = {segments: [{
+    key: {
```
dmlap: Are the key properties important to this test?

test/videojs-hls_test.js
```diff
@@ -1724,6 +1724,41 @@ test('calling play() at the end of a video resets the media index', function() {
...
+  player.hls.playlists.media = function() {
+    return newMedia;
+  };
+
+  player.hls.sourceBuffer = undefined;
+  compareBuffer = [{mediaIndex: 0, playlist: newMedia, offset: 0, bytes: [0,0,0]}];
+  player.hls.segmentBuffer_ = [{mediaIndex: 0, playlist: newMedia, offset: 0, bytes: [0,0,0]}];
+
+  player.hls.drainBuffer();
+  deepEqual(player.hls.segmentBuffer_,compareBuffer,'playlist remains unchanged');
```
dmlap: Please add whitespace between arguments to functions.

test/videojs-hls_test.js
```diff
@@ -1724,6 +1724,41 @@ test('calling play() at the end of a video resets the media index', function() {
...
+    },
+    uri: 'http://media.example.com/fileSequence53-B.ts'
+  }]};
+  player.hls.playlists.media = function() {
+    return newMedia;
+  };
+
+  player.hls.sourceBuffer = undefined;
+  compareBuffer = [{mediaIndex: 0, playlist: newMedia, offset: 0, bytes: [0,0,0]}];
+  player.hls.segmentBuffer_ = [{mediaIndex: 0, playlist: newMedia, offset: 0, bytes: [0,0,0]}];
```
dmlap: If you could add a comment explaining why this comparison proves that drainBuffer() isn't dropping segments, I think that would make this test easier to understand.

bc-bbay:
Making the segment a key makes the bug-infected code throw an error sooner. It gets to the segmentBuffer.shift() without having to override other function calls that happen in drainBuffer() and are irrelevant to the test, and I think it's the lesser of two evils.

test/videojs-hls_test.js
bc-bbay: Making the segment a key makes the bug-infected code throw an error sooner. It gets to the segmentBuffer.shift() without having to override other function calls that happen in drainBuffer() and are irrelevant to the test, and I think it's the lesser of two evils.

test/videojs-hls_test.js
bc-bbay: Will do.

test/videojs-hls_test.js
bc-bbay: Will do.

PR #243, opened 2015-04-15
Resolution best variant plus one.

gkatsev:
If we have a perfect resolution match for playlist and player, use that
playlist. Otherwise, get the next highest resolution matched playlist,
even if bigger than the player.

The question here is when exactly do we want the `resolutionPlusOne` to take precedence?

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -427,7 +427,6 @@ videojs.Hls.prototype.selectPlaylist = function () {
     i = sortedPlaylists.length,
     variant,
     oldvariant,
-    ret,
     bandwidthBestVariant,
     resolutionPlusOne,
     resolutionBestVariant;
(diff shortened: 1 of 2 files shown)


dmlap:
Test case?

Would it simplify things a bunch to always use resolutionPlusOne? I'm tempted to do that all the time if so.

gkatsev:
Yep, working on a test case.
It definitely would simply things to always use `resolutionPlusOne`, though, I think, keeping the exact match is a good compromise. The main issue I think is a scenario where the resolution under the player size is just *slightly* below it where-as the `resolutionPlusOne` is significantly bigger.
I.e., player size is `300x150` but `resolutionPlusOne` is `1920x1080` and the resolution under ends up being `264x149`. I guess it would be best to take distance into account but not sure it is worth it.

gkatsev:
I updated tests.
Also, just realized that with this we would never end up choosing the audio-only rendition in just about any case. Thoughts, @dmlap?

dmlap:
It seems like audio-only would still be chosen in low bandwidth scenarios, which sounds right. LGTM.

gkatsev:
Yes. That does sound right, actually.

PR #244, opened 2015-04-15
Use presentation time stamp to calculate playlist duration

bc-bbay:


src/videojs-hls.js
```diff
@@ -736,8 +736,14 @@ videojs.Hls.prototype.drainBuffer = function(event) {
   this.segmentParser_.flushTags();
 
   tags = [];
+  segment.preciseTimestamp = -1;
```
gkatsev: why not default it `null`? Then below on line `890` you only need to check for truthyness.

src/videojs-hls.js
```diff
@@ -881,7 +887,19 @@ videojs.Hls.getPlaylistDuration = function(playlist, startIndex, endIndex) {
 
   for (; i >= startIndex; i--) {
     segment = playlist.segments[i];
-    dur += (segment.duration !== undefined ? segment.duration : playlist.targetDuration) || 0;
+    if (segment.preciseTimestamp && segment.preciseTimestamp >= 0) {
+      var offset = 0;
```
gkatsev: could we calculate the `preciseDuration` statically instead of having to re-calculate it every time?
Or perhaps, we could store it into a `preciseDuration` after the first calculation.

gkatsev: also, var should be declared at the top of the scope :)

gkatsev:
LGTM. @dmlap?

src/videojs-hls.js
```diff
@@ -736,8 +736,14 @@ videojs.Hls.prototype.drainBuffer = function(event) {
   this.segmentParser_.flushTags();
 
   tags = [];
+  segment.preciseTimestamp = null;
+
   while (this.segmentParser_.tagsAvailable()) {
     tags.push(this.segmentParser_.getNextTag());
+    if (!tags[tags.length - 1].counted && tags[tags.length-1].pts) {
```
dmlap: Are you doing this inside the loop in case PTS values are not ordered ascending? I don't think that's necessary.

src/videojs-hls.js
```diff
@@ -881,7 +888,26 @@ videojs.Hls.getPlaylistDuration = function(playlist, startIndex, endIndex) {
 
   for (; i >= startIndex; i--) {
     segment = playlist.segments[i];
-    dur += (segment.duration !== undefined ? segment.duration : playlist.targetDuration) || 0;
+    if (segment.preciseDuration || segment.preciseTimestamp) {
```
dmlap: Can you calculate the segment duration immediately following the muxing so you don't have to check whether it needs to be calculated or not every time the duration is computed?

dmlap:
This could use a test case. Also, will the precise durations get carried along when live playlists reload? I just ran into a problem with keys and playlist reloading in HLSe :)

src/videojs-hls.js
```diff
@@ -719,10 +734,26 @@ videojs.Hls.prototype.drainBuffer = function(event) {
   this.segmentParser_.flushTags();
 
   tags = [];
+
   while (this.segmentParser_.tagsAvailable()) {
     tags.push(this.segmentParser_.getNextTag());
   }
 
+  if (tags.length > 0) {
```
dmlap: This section could use some commenting about what is going on and why.

src/videojs-hls.js
```diff
@@ -894,7 +925,11 @@ videojs.Hls.getPlaylistDuration = function(playlist, startIndex, endIndex) {
 
   for (; i >= startIndex; i--) {
     segment = playlist.segments[i];
-    dur += (segment.duration !== undefined ? segment.duration : playlist.targetDuration) || 0;
+    if (segment.preciseDuration) {
+      dur += segment.preciseDuration;
```
dmlap: More concisely (though maybe not more readably): `dur += segment.preciseDuration || segment.duration || playlist.targetDuration || 0`

dmlap:
It looks like this PR has picked up some commits that have been merged into master. Can you rebase it agains the current head to clean things up?

PR #245, opened 2015-04-16
enable support for "ffmpeg -reset_timestamps 1" segments

dkuryakin:
This patch allow you to play segments with "ffmpeg -reset_timestamps 1". It's very useful in case if you want to generate dynamic random-ordered m3u8 lists from predefined set of segments and play it in gapless maner.

Usage example:

var player = videojs('video', {
    hls: {
        resetTimestamps: true
    }
});
player.play();

### DIFF
--- a/src/segment-parser.js
+++ b/src/segment-parser.js
@@ -124,7 +124,9 @@
       return tag.finalize();
     };
 
-    self.parseSegmentBinaryData = function(data) { // :ByteArray) {
+    self.parseSegmentBinaryData = function(data, segmentOffset) { // :ByteArray) {
+      segmentOffset = segmentOffset ? segmentOffset : 0;
+      
       var
         dataPosition = 0,
         dataSlice;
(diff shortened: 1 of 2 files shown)


dmlap:
@dkuryakin this setting in ffmpeg causes TS timestamps to always reset across segments? Do you have a link to the docs for it?

dkuryakin:
Yeah, this option allow TS segments to be shuffled in random order. So, it's possible to generate valid  dynamic m3u8 playlists.

I find following links:
http://stuff.onse.fi/man?program=ffmpeg-formats&section=1
https://searchcode.com/codesearch/view/37369513/

Example of ffmpeg command I use:
ffmpeg -y -i $INPUT -vcodec libx264 -acodec aac -r 30 -g 300 -sc_threshold 30 -keyint_min 3 -force_key_frames "expr:gte(t,n_forced*10)" -flags -global_header -f segment -segment_list stream.m3u8 -segment_time 10 -segment_format mpegts -profile:v baseline -segment_list_type m3u8 -strict experimental -segment_list_entry_prefix segments/ -reset_timestamps 1 segments/%03d.ts

forbesjo:
You should probably make every segment a discontinuity if you want to do this.

PR #249, opened 2015-04-18
Trigger NetStreamAppendBytesAction.END_SEQUENCE when the mediasource buf...

bclwhitaker:
...fers empty. Requires a corresponding pull in videojs-contrib-media-sources.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -267,9 +267,16 @@ videojs.Hls.prototype.handleSourceOpen = function() {
   // construct the video data buffer and set the appropriate MIME type
   var
     player = this.player(),
-    sourceBuffer = this.mediaSource.addSourceBuffer('video/flv; codecs="vp6,aac"');
+    mediaSource = this.mediaSource,
+    sourceBuffer = mediaSource.addSourceBuffer('video/flv; codecs="vp6,aac"');
 
   this.sourceBuffer = sourceBuffer;
+
+  // signal whenever the buffer flushes to the swf
+  this.sourceBuffer.addEventListener('updateend', function() {
+    mediaSource.endOfStream();
+  });
+
   sourceBuffer.appendBuffer(this.segmentParser_.getFlvHeader());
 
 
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -267,9 +267,16 @@ videojs.Hls.prototype.handleSourceOpen = function() {
...
     player = this.player(),
-    sourceBuffer = this.mediaSource.addSourceBuffer('video/flv; codecs="vp6,aac"');
+    mediaSource = this.mediaSource,
+    sourceBuffer = mediaSource.addSourceBuffer('video/flv; codecs="vp6,aac"');
 
   this.sourceBuffer = sourceBuffer;
+
+  // signal whenever the buffer flushes to the swf
+  this.sourceBuffer.addEventListener('updateend', function() {
+    mediaSource.endOfStream();
```
dmlap: Won't this trigger endOfStream() after every segment finishes?

src/videojs-hls.js
```diff
@@ -671,7 +678,6 @@ videojs.Hls.prototype.loadSegment = function(segmentUri, offset) {
       offset: offset,
       bytes: new Uint8Array(this.response)
     });
-    player.trigger('progress');
```
dmlap: Why did you remove this?

test/videojs-hls_test.js
```diff
@@ -1734,7 +1733,7 @@ test('calls ended() on the media source at the end of a playlist', function() {
   requests[0].response = new ArrayBuffer(17);
   requests.shift().respond(200, null, '');
 
-  strictEqual(endOfStreams, 1, 'ended media source');
+  strictEqual(updateendListener, 1, 'ended media source');
```
dmlap: `updateend` should fire after every appendBuffer() operation has completed, not the end of the stream. I think this test is asserting on the wrong thing now.

src/videojs-hls.js
bclwhitaker: Yes, it was.  Most of this logic has now been moved into the mediaSources project via https://github.com/videojs/videojs-contrib-media-sources/pull/24

src/videojs-hls.js
bclwhitaker: progress events are already firing from videojs because the bufferedPercent is changing.  Since all the other changes in this pull has been moved into mediaSources I opened a separate pull request to remove just this event #252.  Firing progress here was sometimes causing the buffered region in the player to misreport.

Issue #218, opened 2015-02-23
Video stutters when the next downloaded segment is being decrypted


rajkosto:
Non-encrypted HLS livestream test: http://sshnuke.net/teststream.php?03ej74H0LvDFBefzDCY4rPbtWYYbgdIHIGFA9t0s%2FcGhI%3D
Encrypted HLS livestream test: http://sshnuke.net/teststream.php?9E81QtnIG1LuyLhKsOq0RMYwb%2FOi739KRh8%2BO6FNX34epYcZXfUzuQEPVTw4zt%2BU0j

Same delivery method, same .m3u8 and .ts file content (other than one having the key header and the other being encrypted with a constant key in a normal file)
Best way to notice this is to view in chrome, go into Developer Tools->Rendering tab on the bottom, and check 'Show paint rectangles'. Then watch the green tint on the black borders around the stream. The encrypted one will fade out (indicating no updates happening) when the aes decrypt method is being called (which you can confirm by CPU profiling in chrome) on a newly-downloaded segment file.
The non-encrypted one will be rock-solid except when switching qualities.
The CPU is definitely not being overloaded, and the bitrate of the stream is only around 2.3mbps for the highest quality (the stuttering can be seen on all qualities, actually), the problem  is that the decrypt method is being called in a blocking fashion which prevents the video from being updated at the same time.
Other players such as flashls based ones do not have any issue with this.

dmlap:
Thank you very much for the test streams! We'll take a look.

dmlap:
I'm looking into this now. My hypothesis was swapping byte order before and after decrypting was tons of overhead and taking up the bulk of the time. My initial profiling in Chrome indicates a couple things:

1. Decryption is at least 10x as expensive as transmuxing right now
1. Byte-order swapping, CBC, and pkcs7 do not play a significant role in decryption time
1. AES.decrypt() takes up 2/5 of decryption time in my test runs

All of which seem to invalidate my initial hunch. Assuming AES.decrypt() is the culprit, I'm going to explore a couple options:

- Using TypedArrays instead of vanilla arrays during decryption
- Finding a way to decrypt asynchronously (web workers, incremental decryption, etc)

Still investigating.

dmlap:
I ended up building v8 and using the debugging console to try out a bunch of crazy ideas. For future reference:

1. Inlining AES.decrypt() and hoisting loop invariants (the key, table references, etc.) can shave about 5% off decryption time
[...]

PR #251, opened 2015-04-21
Optimize decryption

dmlap:
For #218. Use TypedArrays instead of DataViews because they weren't getting optimized very well. Use Int32Arrays instead of Uint32Arrays to avoid v8 deoptimization at runtime when negative values were encountered. Swap byte order in the main loop instead of before-hand. These changes don't get decrypt running in less than 16ms on my machine but they do speed things up an order of magnitude or so. Additional work to make the decryption asynchronous appears necessary.

### DIFF
--- a/src/decrypter.js
+++ b/src/decrypter.js
@@ -38,7 +38,18 @@
 (function(window, videojs, unpad) {
 'use strict';
 
-var AES, decrypt;
+var AES, AsyncStream, Decrypter, decrypt, ntoh;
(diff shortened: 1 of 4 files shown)


dmlap:
This is still a work in progress. I will be adding the async changes here as well.

dconnolly:
Overall LGTM.

dmlap:
@dconnolly thanks for the review!

PR #252, opened 2015-04-22
Dont fire progress events from hls. 

bclwhitaker:
videojs uses the bufferedPercent from the swf and fires it already.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -671,7 +671,6 @@ videojs.Hls.prototype.loadSegment = function(segmentUri, offset) {
       offset: offset,
       bytes: new Uint8Array(this.response)
     });
-    player.trigger('progress');
     tech.drainBuffer();
 
     tech.mediaIndex++;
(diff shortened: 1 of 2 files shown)


bclwhitaker:
The video.js Flash swf updates the buffered progress on the control bar without this line.  

When the video was looping, I recall that updating progress here was causing the buffered progress not to be correct (i.e. it wasn't extending all the way to the end of the video even though it had buffered it all). As a result, the playback scrubber would pass the buffered area without actually buffering making the buffered region representation unreliable.

dmlap:
Could that be an issue with the playback scrubber and not this event? I feel that triggering `progress` here is consistent with the event definition, i.e. new video data has been downloaded.

bclwhitaker:
I think that definition assumes that downloading it means its added to the buffer, which in this case isn't true until its sent to the swf.

dmlap:
That definition doesn't count buffering in JS :) I think I understand your perspective on this one but I think what we're currently doing is reasonable, too. I'm closing this one for now but if there's a more fundamental issue with firing progress here, please reopen it.

PR #260, opened 2015-05-06
Use presentation time stamp to calculate playlist duration

bc-bbay:
This code uses the presentation timestamp of the ts segment to calculate its exact duration, since this may differ by fractions of a second from what is reported. Using the exact, calculated 'preciseDuration' allows for smoother seeking and calculation of the total playlist duration, which previously (especially in short videos) was reported erroneously and made the play head overrun the end of the progress bar.


dmlap:
LGTM

PR #262, opened 2015-05-07
Id3 enhancements

dmlap:
Rebased #258. Added a bit more first-class parsing for PRIV frames and exposed the bytes of all frames on the generated VTTCue so that they could be parsed upstream if we don't have explicit support for a particular frame type. Added buffering in metadata stream and fixed an issue in packet parsing so that ID3 tags that span multiple TS packets are reformed before being parsed.

### DIFF
--- a/src/metadata-stream.js
+++ b/src/metadata-stream.js
@@ -6,23 +6,24 @@
 (function(window, videojs, undefined) {
   'use strict';
   var
-    // return the string representation of the specified byte range,
-    // interpreted as UTf-8.
-    parseUtf8 = function(bytes, start, end) {
+    // return a percent-encoded representation of the specified byte range
+    // @see http://en.wikipedia.org/wiki/Percent-encoding
+    percentEncode = function(bytes, start, end) {
       var i, result = '';
       for (i = start; i < end; i++) {
         result += '%' + ('00' + bytes[i].toString(16)).slice(-2);
       }
-      return window.decodeURIComponent(result);
+      return result;
+    },
+    // return the string representation of the specified byte range,
+    // interpreted as UTf-8.
+    parseUtf8 = function(bytes, start, end) {
+      return window.decodeURIComponent(percentEncode(bytes, start, end));
     },
     // return the string representation of the specified byte range,
     // interpreted as ISO-8859-1.
     parseIso88591 = function(bytes, start, end) {
-      var i, result = '';
-      for (i = start; i < end; i++) {
-        result += '%' + ('00' + bytes[i].toString(16)).slice(-2);
-      }
-      return window.unescape(result);
+      return window.unescape(percentEncode(bytes, start, end));
     },
     tagParsers = {
       'TXXX': function(tag) {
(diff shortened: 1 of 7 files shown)


dmlap:
@oldcookie: I modified the format you had used to expose PRIV tags a bit so that it would allow contrib-hls users to get at the data for any ID3 frame. So instead of `cue.text` being a binary string, you could get at a Uint8Array of a frame through `cue.frame.data` and the private data portion of a PRIV frame at `cue.frame.privateData`. How does that sound?

gkatsev:
LGTM

oldcookie:
Looks good, thanks!

forbesjo:
:+1:

src/metadata-stream.js
```diff
@@ -6,12 +6,24 @@
 (function(window, videojs, undefined) {
   'use strict';
   var
```
mmcc: This is way out of the scope of this review, so no changes requested here, just curious. Is there a reason for this beyond style choice?

gkatsev: the `var` thing? Yeah, it just lines up the names with one (2-space) indent.

mmcc: Well I see why `var` is there, I'm more asking about the benefits of having an 80 line multiple var statement. Not trying to be contrary, I've just seen this pattern pop up in a few different projects recently.

gkatsev: it's a mishmash of several rules we've been following. Like only declaring vars at the top and using function expressions instead of function declarations.

mmcc: Those are all great rules by themselves, but I think the combination makes for a breadcrumb trail made of commas and pain. For what it's worth, I really like Ben Alman's [take on multiple var statements](http://benalman.com/news/2012/05/multiple-var-statements-javascript/). Feels roughly similar to where we've landed in VJS core?

> My rule is as-follows: I’ll put multiple comma-separated declarations in a single var statement, but only if they don’t have assignments. Assignments must be in their own var statement. Like this:
```javascript
var foo, bar, baz;
var abc = 1;
var def = 2;
```

Again, this probably isn't the place for this conversation, but as someone not familiar with this style, this is ripe for massive wtf moments. Anecdotally, I had one of those massive wtf moments when trying to help someone out internally with BCC code that uses this same style.

Somewhat related: are we planning on using [standard](https://github.com/videojs/standard) across contrib plugins as well?

`/me climbs off soap box`

dmlap: @mmcc I'm with you-- I definitely want to adopt [standard](https://github.com/videojs/standard) across all the contrib plugins. We haven't had a chance to convert this project over, however, so I'm following the current conventions until we can do that sweep.

gkatsev: Yep, part of the idea for forking standard to videojs-standard was that we could have it included in all our projects.

mmcc: :+1: Sounds good. Call me crazy, but purely stylistic changes can probably sit on the back burner for a little while. :)

I'm down to help with those when the time comes, it'd be nice to get a better feel for the internals of some of these contrib plugins.

Issue #233, opened 2015-03-23
Support preload settings


dmlap:
HLS always behaves as if the preload attribute is set to "auto". It is useful in some settings to have support for the other values as well.

sinr0202:
I'm also looking for a solution to this problem. Seems like adding preload="false" only copies the attribute to the resulting element.

PR #265, opened 2015-05-11
Respect the preload-none option

bc-bbay:
Don’t start downloading segments if preload is set to “none”

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -143,9 +143,6 @@ videojs.Hls.prototype.src = function(src) {
 
     setupEvents = function() {
       this.fillBuffer();
-
-
-
       player.trigger('loadedmetadata');
     };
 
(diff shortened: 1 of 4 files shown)


src/videojs-hls.js
```diff
@@ -558,6 +555,11 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
     segment,
     segmentUri;
 
+  // if preload is set to "none", do not download segments until playback is requested
+  if (!player.hasClass('vjs-has-started') && player.options.preload === "none") {
```
dmlap: Should the second condition be `player.preload() === 'none'`? `options` are set on player creation and (I believe) ignored from that point out.

dmlap:
Nice!

dmlap:
Related: #233

shaharmor:
+100 ! this is very disturbing!

dmlap:
@bc-bbay I just tried this out on a player and it appears preload() is (incorrectly) returning a boolean. Can you double check this is working as you expect?

bc-bbay:
You're right, preload was a boolean in the SWF. I've updated that project and this pull now depends on this one in video-js-swf: https://github.com/videojs/video-js-swf/pull/155

bclwhitaker:
LGTM

PR #267, opened 2015-05-14
Discontinuity insertion race

### LINKED ISSUE #266: PTS-based duration doesn't account for discontinuities
The duration of the previous segment is used to calculate the precise duration of the current segment based on the PTS values in the generated FLV tags. If the previous segment crosses over a timestamp discontinuity, the calculation will be inaccurate.

dmlap:
If segments are delivered very quickly, it's possible that a segment after a discontinuity will be ready before the previous segment is finished getting processed by the media source. Our current mechanism for signalling discontinuities is synchronous so this could lead to us injecting a discontinuity into the middle of the first segment instead of at the end. Instead, adopt a workflow more closely aligned to how real SourceBuffers work and don't append additional bytes until the previous append has been fully processed.

Also includes a fix for precise segment duration after discontinuities. Fixes #266

### DIFF
--- a/src/flv-tag.js
+++ b/src/flv-tag.js
@@ -377,7 +377,10 @@ hls.FlvTag.durationFromTags = function(tags) {
     return 0;
   }
 
-  var first = tags[0], last = tags[tags.length - 1], frameDuration;
+  var
+    first = tags[0],
+    last = tags[tags.length - 1],
+    frameDuration;
 
   // use the interval between the last two tags or assume 24 fps
   frameDuration = last.pts - tags[tags.length - 2].pts || (1/24);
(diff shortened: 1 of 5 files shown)


src/flv-tag.js
```diff
@@ -358,4 +363,26 @@ hls.FlvTag.frameTime = function(tag) {
...
+ * @param tags {array} the FlvTag objects to query
+ * @return the number of milliseconds between the display time of the
+ * first tag and the last tag.
+ */
+hls.FlvTag.durationFromTags = function(tags) {
+  if (tags.length < 2) {
+    return 0;
+  }
+
+  var first = tags[0], last = tags[tags.length - 1], frameDuration;
```
gkatsev: each of these should be on new lines.

gkatsev:
Other than the minor comment, LGTM

PR #270, opened 2015-05-18
Always write out metadata at the start of segments and refactor of h264stream

gkatsev:
In some cases (most likely badly encoded video), segments didn't start
with a keyframe so they ended up getting squished because flash didn't
know how to play them.

### DIFF
--- a/src/h264-stream.js
+++ b/src/h264-stream.js
@@ -283,7 +283,9 @@
           newExtraData = new H264ExtraData();
         }
 
-        if (h264Frame.keyFrame) {
+        // Check if keyframe and the length of tags.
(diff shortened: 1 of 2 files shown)


dmlap:
@gkatsev is there an easy way to test this?

gkatsev:
Not sure. I have a stream that shows this issue currently.
Working on the failing test currently as well.

gkatsev:
Looks like the current test is failing because it is entering into this section, so, might be able to duplicate the currently failing test into a working test for this option.

gkatsev:
I have an update to fix the currently broken test. I don't currently have a test that checks the functionality in this PR because it's impossible to stub out the methods that we care about, currently.

PR #272, opened 2015-05-18
Refactor h264stream

gkatsev:


src/segment-parser.js
```diff
@@ -214,6 +214,16 @@
         // adaptation_field_control, whether this header is followed by an
         // adaptation field, a payload, or both
         afflag = (data[offset + 3] & 0x30 ) >>> 4,
+        adaptation_field_length,
+        afftemp,
+        discontinuity_indicator,
+        random_access_indicator,
```
gkatsev: we only really care about the `random_access_indicator` currently, so, I could remove the others if needed.

src/segment-parser.js
```diff
@@ -247,7 +257,27 @@
...
+          elementary_stream_priority_indicator = (afftemp & 0x20) >>> 5;
+          PCR_flag  = (afftemp & 0x10) >>> 4;
+          OPCR_flag = (afftemp & 0x08) >>> 3;
+          splicing_point_flag = (afftemp & 0x04) >>> 2;
+          transport_private_data_flag = (afftemp & 0x02) >>> 1;
+          adaptation_field_extension_flag  = afftemp & 0x01;
+
+          if (random_access_indicator === 1) {
+            if (h264Stream._h264Frame)
+              h264Stream._h264Frame.keyFrame = true;
```
gkatsev: I'm not certain whether this is a good way of doing it. Another idea I have is something like `h264Stream.setNextFrameToKeyFrame()` or something.
However, this does work.

src/segment-parser.js
gkatsev: [According to szatmary, the RAI means keyrame](http://stackoverflow.com/a/18748654)

gkatsev:
Looking into build failure.

src/h264-extradata.js
```diff
@@ -0,0 +1,213 @@
+(function() {
```
dmlap: What changed in here?

gkatsev: I moved the `H264ExtraData` object into its own file instead of having it live inside of `h264-stream.js`

src/segment-parser.js
dmlap: I know this code is very inconsistent about variable naming but we should stick with camel case.

src/segment-parser.js
```diff
@@ -247,7 +257,28 @@
...
+          afftemp = data[offset + 1];
+
+          discontinuity_indicator = (afftemp & 0x80) >>> 7;
+          random_access_indicator = (afftemp & 0x40) >>> 6;
+          elementary_stream_priority_indicator = (afftemp & 0x20) >>> 5;
+          PCR_flag  = (afftemp & 0x10) >>> 4;
+          OPCR_flag = (afftemp & 0x08) >>> 3;
+          splicing_point_flag = (afftemp & 0x04) >>> 2;
+          transport_private_data_flag = (afftemp & 0x02) >>> 1;
+          adaptation_field_extension_flag  = afftemp & 0x01;
```
dmlap: Why do you parse these additional fields?

src/segment-parser.js
```diff
@@ -247,7 +257,28 @@
...
+
+          if (random_access_indicator === 1) {
+            if (h264Stream._h264Frame) {
+              h264Stream._h264Frame.keyFrame = true;
```
dmlap: What about if we made this a getter/setter method so it was a little clearer there are side effects going on behind the scenes?

src/segment-parser.js
gkatsev: I literally copy/pasted this from the spec :), but yes, sticking with camel case is fine.

src/segment-parser.js
gkatsev: Purely for completeness. I wanted to parse it out to compare it with other output but we don't really need them here yet (maybe ever) but at least now they'll be in the history.

src/segment-parser.js
gkatsev: Getter/setter for `h264Frame` or for `keyFrame`?
My other comment went away when I fixed the test but I said I was thinking about maybe having something like `h264Stream.setNextFrameToKeyFrame()`.

src/segment-parser.js
dmlap: Cool. I do think we should nix them before merging.

src/segment-parser.js
dmlap: I was thinking of turning `keyFrame` into a combo getter/setter but `setNextFrameToKeyFrame` seems like a legit option, too.

src/h264-extradata.js
dmlap: In the future, please try to split refactor PRs from functionality-changing PRs so it's clearer what needs to be reviewed.

src/segment-parser.js
gkatsev: Updated to use `h264Stream.setNextFrameKeyFrame()`.

src/h264-stream.js
```diff
@@ -241,235 +34,244 @@
...
+        this._h264Frame = new FlvTag(FlvTag.VIDEO_TAG);
+        this._h264Frame.pts = this._next_pts;
+        this._h264Frame.dts = this._next_dts;
 
-        // finishFrame may render h264Frame null, so we must test again
-        if (!h264Frame) {
-          h264Frame = new FlvTag(FlvTag.VIDEO_TAG);
-          h264Frame.pts = next_pts;
-          h264Frame.dts = next_dts;
+        if (this._nextFrameKeyFrame) {
```
gkatsev: when creating a new h264Frame, check to see if it'll be a keyframe.

src/h264-stream.js
```diff
@@ -241,235 +34,244 @@
...
-            }
 
-            if (length > 1 && data[offset] === 0 && data[offset + 1] === 1) {
-              // ?? 00 | 00 01
-              if (nalUnitSize >= 2 && h264Frame.negIndex(2) === 0) {
-                h264Frame.length -= 2; // 00 00 : 00 01
-              } else {
-                h264Frame.length -= 1; // 00 : 00 01
-              }
+  H264Stream.prototype.setNextFrameKeyFrame = function() {
```
gkatsev: where the function is defined.

src/h264-stream.js
```diff
@@ -241,235 +34,246 @@
...
 
-      oldExtraData = new H264ExtraData(), // :H264ExtraData
-      newExtraData = new H264ExtraData(), // :H264ExtraData
+    this._nalUnitType = -1; // :int
 
-      nalUnitType = -1, // :int
+    this._state = 0; // :uint;
 
-      state; // :uint;
+    this._nextFrameKeyFrame = false;
```
gkatsev: defaulting `nextFrameKeyFrame` to false.

dmlap:
LGTM

Issue #236, opened 2015-04-02
Quality switching repeats chunks


ntadej:
We have some problems with our live streams on quality switching. Some chunks get repeated after switch.

Details what happens on quality switch:
 - all already downloaded chunks from previous quality play to the end
 - video freezes but audio repeats from one or two chunks from new quality
 - after that audio/video are in sync again and video plays normally

Note that our streams might not have ADTS header available at start (error Assertion failed: Error no ATDS header found).

Our streams do not have audio and video aligned. PR #213 helps with that and does not affect quality switching.

I have sent the private test URL to email to @dmlap.

dmlap:
#213 has been merged and some additional fixes have been merged in the meantime. Are you still seeing this issue in master, @ntadej?

ntadej:
Unfortunatelly, yes. There don't seem to be any visible improvements. In some cases it even took longer for playback to recover to normal state.

PR #274, opened 2015-05-20
Live playlist loading improvements

ntadej:
We managed to fix the problems described in #236. This PR consists of two commits.

The repetition of chunks was caused because media index didn't need to be translated when changing quality. The playlist was the same and translation should not occur.

The second commit fixes crazy loading of playlists and chunks after several quality changes as timeouts accumulated and produced more and more requests.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -94,6 +94,7 @@
 
           // refresh live playlists after a target duration passes
           if (!loader.media().endList) {
+            window.clearTimeout(mediaUpdateTimeout);
             mediaUpdateTimeout = window.setTimeout(function() {
               loader.trigger('mediaupdatetimeout');
             }, refreshDelay);
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -1035,7 +1035,7 @@ videojs.Hls.translateMediaIndex = function(mediaIndex, original, update) {
   // bitrate switches should be happening here.
   translatedMediaIndex = (mediaIndex + (original.mediaSequence - update.mediaSequence));
 
-  if (translatedMediaIndex >= update.segments.length || translatedMediaIndex < 0) {
+  if (translatedMediaIndex > update.segments.length || translatedMediaIndex < 0) {
     // recalculate the live point if the streams are too far out of sync
     return videojs.Hls.getMediaIndexForLive_(update) + 1;
   }


src/playlist-loader.js
```diff
@@ -94,6 +94,7 @@
 
           // refresh live playlists after a target duration passes
           if (!loader.media().endList) {
+            window.clearTimeout(mediaUpdateTimeout);
```
dmlap: This had to be a huge pain to track down. Thank you!

src/videojs-hls.js
```diff
@@ -1030,6 +1030,11 @@ videojs.Hls.translateMediaIndex = function(mediaIndex, original, update) {
     // let the media index be zero when there are no segments defined
     return 0;
   }
+  
+  if (original.mediaSequence === update.mediaSequence) {
+	// playlist sequence didn't change
+	return mediaIndex;
```
dmlap: Isn't this equivalent to the [computation of translatedMediaIndex](https://github.com/ntadej/videojs-contrib-hls/blob/playlist-fix/src/videojs-hls.js#L1041)? If `original.mediaSequence === update.mediaSequence`, then:

```
(mediaIndex + (original.mediaSequence - update.mediaSequence))
=> (mediaIndex + (0))
=> mediaIndex
```
The only case where that value is unused is if the resulting index is outside the scope of the current playlist. Were you seeing situations where `translatedMediaIndex === update.segments.length`?

ntadej: Yes, the problem is when `translatedMediaIndex === update.segments.length`. The index gets increased and falls out of the playlist.

dmlap: Ok, that makes sense. It seems to me that the conditional is wrong. If we made the test `translatedMediaIndex >= update.segments.length + 1`, would that resolve the issue?

ntadej:
Updated PR according to your comments.

src/videojs-hls.js
```diff
@@ -1035,7 +1035,7 @@ videojs.Hls.translateMediaIndex = function(mediaIndex, original, update) {
   // bitrate switches should be happening here.
   translatedMediaIndex = (mediaIndex + (original.mediaSequence - update.mediaSequence));
 
-  if (translatedMediaIndex >= update.segments.length || translatedMediaIndex < 0) {
+  if (translatedMediaIndex > update.segments.length || translatedMediaIndex < 0) {
```
bc-bbay: This change doesn't feel right. Segments is a 0-indexed array, so update.segments.length is an out-of-bounds element. Why is this necessary?

dmlap: When all the segments currently available in a live playlist have been downloaded, mediaIndex will point to one beyond the end. There's nothing wrong with this situation, we're just waiting for a playlist refresh. It's incorrect to reset mediaIndex back to the live point if this occurs.

PR #282, opened 2015-05-27
Trim whitespace in playlist

forbesjo:
Fix from #279. 

> If playlists contain "\r\n" line endings or even whitespace lines the m3u8 parser falsely detects those lines as "empty" segment URI-s. This results in faulty duration calculation and possible failure at playback start. By trimming line data before feeding it into the parser this issue is fixed.


dmlap:
LGTM

PR #284, opened 2015-05-27
Ignore nit

dmlap:
Cherry-picked and rebased version of the NIT fix in #279. Also, added handling for PCR PIDs so that a bunch of spurious warnings aren't generated.

@mikrohard I updated your change so that console warnings wouldn't be generated for NIT packets. This still look good to you?


mikrohard:
It looks good as far as I'm able to say be just looking at it. Will do a runtime test tomorrow...

mikrohard:
I can confirm that it works the way it should. Great job!

PR #285, opened 2015-05-27
Quicker playlist switching

dmlap:
If bandwidth drops off precipitously, it's not a good idea to fetch a higher quality segment while waiting for the new media playlist to load. Delay segment loading instead. Less fetching will be done in parallel after this change but:
 - the player will adjust to dramatic bandwidth drops more quickly
 - the player will upshift more quickly when bandwidth returns


imbcmdth:
LGTM :tongue:

PR #290, opened 2015-05-29
Don't add metadata at random access indicators

### LINKED ISSUE #289: Tearing/melting when scrubbing in Firefox
Seen using the code after Gary's fix post 0.16.0.
While scrubbing an HLS video in Firefox 38 the frames melt together for a second, I am able to reproduce this 100% of the time.
I'm not seeing this issue in Chrome.

may 29 2015 11-08

may 29 2015 11 15

dmlap:
A metadata tag injected into the generated FLV when a random access indicator was encountered in the adaptation field of a TS packet causes Firefox 38 on OSX to apply frame diffs against the wrong keyframe. This leads to ugly artifacts for a few seconds after seeking. Now metadata tags are only generated at the start of a segment and IDRs. Fixes #289.

### DIFF
--- a/src/h264-stream.js
+++ b/src/h264-stream.js
@@ -48,11 +48,9 @@
 
     this._state = 0; // :uint;
 
-    this._nextFrameKeyFrame = false;
-
     this.tags = [];
-
   };
+
   //(pts:uint):void
   H264Stream.prototype.setTimeStampOffset = function(pts) {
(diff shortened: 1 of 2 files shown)


forbesjo:
:+1:

gkatsev:
Weird. I know if there's anyone we could ask about why this could happen.

PR #293, opened 2015-06-01
Get the switching simulator working again

dmlap:
Mock out metadata stream support. Comment javascript simulation parameters. Simulation results currently always buffer the first segment because of the bandwidth scaling factor applied to deal with real-world discrepancies between m3u8 download times and steady-state downloads. Behavior after the first segment still looks good.


forbesjo:
:+1:

PR #294, opened 2015-06-01
Fix seek after end

bc-bbay:
Seeking after ‘ended’ event would cause the player to begin playback at
the beginning, but count up from the seek point timestamp. Now it will
play from the seek point.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -318,7 +318,13 @@ videojs.Hls.prototype.setupMetadataCueTranslation_ = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    this.mediaIndex = 0;
+    if (this.currentTime() === 0 || this.currentTime() === undefined || this.currentTime() === this.duration()) {
+      this.mediaIndex = 0;
+      this.setCurrentTime(0);
+    } else {
+      this.mediaIndex = videojs.Hls.getMediaIndexByTime(this.playlists.media(), this.currentTime());
+      this.setCurrentTime(this.currentTime());
+    }
   }
 
   // we may need to seek to begin playing safely for live playlists
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -318,17 +318,15 @@ videojs.Hls.prototype.setupMetadataCueTranslation_ = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    // When the video ends and the 'ended' event fires, current time is set to 0. If the video is in this state, or the
-    // current time is equal to the duration (ie, play head is at the end position), reset the video to the beginning on
-    // a play event.
-    if (this.currentTime() === 0 || this.currentTime() === this.duration()) {
-      this.mediaIndex = 0;
-      this.setCurrentTime(0);
-    // Otherwise, if the user has seeked to another point in the video after the 'ended' event, it is likely the case
-    // that they wish to restart playback from that point, so play from the play head's position.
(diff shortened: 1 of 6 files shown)


src/videojs-hls.js
```diff
@@ -286,7 +286,13 @@ videojs.Hls.prototype.handleSourceOpen = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    this.mediaIndex = 0;
+    if (this.currentTime() === 0 || this.currentTime() === undefined || this.currentTime() === this.duration()) {
```
dmlap: Why are two cases necessary here?

src/videojs-hls.js
```diff
@@ -286,7 +286,18 @@ videojs.Hls.prototype.handleSourceOpen = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    this.mediaIndex = 0;
+    // When the video ends and the 'ended' event fires, current time is set to 0. If the video is in this state, or the
+    // current time is equal to the duration (ie, play head is at the end position), reset the video to the beginning on
+    // a play event.
+    if (this.currentTime() === 0 || this.currentTime() === this.duration()) {
```
dmlap: If currentTime is zero, would this cause unnecessary re-buffering?

Are you sure that `currentTime === duration`? Could currentTime be 9.9999 and duration be 10, for instance? We support m3u8 annotations (`#ZEN-TOTAL-DURATION`) which might not be perfectly accurate.

src/videojs-hls.js
```diff
@@ -286,7 +286,18 @@ videojs.Hls.prototype.handleSourceOpen = function() {
...
+      this.mediaIndex = 0;
+      this.setCurrentTime(0);
+    // Otherwise, if the user has seeked to another point in the video after the 'ended' event, it is likely the case
+    // that they wish to restart playback from that point, so play from the play head's position.
+    } else {
+      this.mediaIndex = videojs.Hls.getMediaIndexByTime(this.playlists.media(), this.currentTime());
+      this.setCurrentTime(this.currentTime());
```
dmlap: With this `if-else`, the player will always seek on play and since we don't support in-buffer seeking, preloading would not provide any benefit.

dmlap: Argh, missed the `if (this.ended())` on line 288. Nevermind this comment.

src/videojs-hls.js
dmlap: My reading of [the ended property](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-ended) seems to imply that `currentTime` should never be zero at this point. I think we should fix that rather than have all of these cases to handle here.

dmlap:
@bc-bbay this PR looks like it was improperly rebased. Can you clean up the commits?

src/videojs-hls.js
```diff
@@ -318,7 +318,16 @@ videojs.Hls.prototype.setupMetadataCueTranslation_ = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    this.mediaIndex = 0;
+    // If this.lastSeekedTime_ is defined, a seek has happened after playback ended, as it is set undefined at video end.
+    // We should begin playback from that point.
+    if (this.lastSeekedTime_ !== undefined) {
+      this.mediaIndex = videojs.Hls.getMediaIndexByTime(this.playlists.media(), this.currentTime());
```
dmlap: Doesn't setCurrentTime() already handle this?

PR #295, opened 2015-06-02
Implement seekable

dmlap:
Override the Flash tech's seekable method to take into account live playlists.


dmlap:
This is dependent on https://github.com/videojs/video.js/commit/1c35bfafc2d6d65f10f86c0f5918e07e3fd4c4e0

src/playlist.js
```diff
@@ -0,0 +1,115 @@
...
+   * for seeking
+   */
+  seekable = function(playlist) {
+    var startOffset, targetDuration;
+    // without segments, there are no seekable ranges
+    if (!playlist.segments) {
+      return videojs.createTimeRange();
+    }
+    // when the playlist is complete, the entire duration is seekable
+    if (playlist.endList) {
```
gkatsev: could it be that you get an `endList` in a sliding window manifest? In which case, the start range shouldn't be zero.

gkatsev: Or would this only come up in the DVR/VOD case only?

gkatsev:
Otherwise, LGTM

src/playlist.js
dmlap: `endList` is the signal that the playlist is complete and there will be no further updates, i.e. it's now a VOD playlist.

gkatsev: How do we know that a sliding window has ended, assuming it isn't turning into a VOD playlist?

forbesjo:
LGTM

PR #298, opened 2015-06-07
getMediaIndexByTime was returning -1 at playlist boundaries

bc-bbay:
It should correctly return the first or last segment in the playlist.


dmlap:
I think we need to find some way to unify this with the duration calculation happening in [the Playlist module](https://github.com/videojs/videojs-contrib-hls/blob/mediaindex-edge-cases/src/playlist.js) (recently moved out of the main file). Thoughts?

bc-bbay:
Obsolete by #308

PR #299, opened 2015-06-08
Don't preload live videos

dmlap:
If play() was called before the media playlist has been downloaded, when the playlist finally was received we would still update media index to the live point to start buffering. We would have skipped the code that adjusted currentTime however, so playback would seem to begin at zero instead of seekable().end(0). That would cause media-timeline relative positions to be off by the live window length and things like cue points would fire 60 seconds (for example) after they should have. Now, don't preload live video at all and just seek to live immediately on play, updating the currentTime at that point.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -129,16 +129,6 @@ videojs.Hls.prototype.src = function(src) {
       });
     }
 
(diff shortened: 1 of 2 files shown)


imbcmdth:
LGreatTM :tongue:

PR #302, opened 2015-06-10
Account for expired segments when seeking

dmlap:
Deprecate getMediaIndexByTime and replace it with a PlaylistLoader.getMediaIndexForTime_ that considers expired content in live playlists. Fix an issues that would allow the return value to be less than zero or greater than the index of the last available media segment. Currently, this code does not take into account rounding of segment durations in HLS v3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-hls",
-  "version": "0.17.1",
+  "version": "0.17.2-0",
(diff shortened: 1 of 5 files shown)


forbesjo:
:+1:

gkatsev:
LGTM

dmlap:
I'm going to rebase this one to remove the build commit.

PR #308, opened 2015-06-15
Live seeking

dmlap:
- Fix an issue where getMediaIndexByTime was returning -1 at playlist boundaries
- Account for expired segments when seeking 
- Fix a condition that would delay playback startup for live streams


gkatsev:
LGTM

bc-bbay:
Looks great! Thanks a lot, David!

PR #309, opened 2015-06-15
Add github-releaser config

dmlap:
Setup grunt-github-releaser for subsequent releases. Remove a leading newline that caused chg to mis-parse the changelog for 0.17.2.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,11 +3,12 @@
 var basename = require('path').basename;
 
 module.exports = function(grunt) {
+  var pkg = grunt.file.readJSON('package.json');
 
   // Project configuration.
   grunt.initConfig({
     // Metadata.
-    pkg: grunt.file.readJSON('package.json'),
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -118,6 +119,23 @@ module.exports = function(grunt) {
...
+          password: process.env.VJS_GITHUB_TOKEN
+        },
+        release: {
+          'tag_name': 'v' + pkg.version,
+          name: pkg.version,
+          body: require('chg').find(pkg.version).changesRaw
+        }
+      },
+      files: {
+        'dist': ['videojs.hls.min.js']
```
gkatsev: Don't we generate an unminified version of the concatenated file? Why not make it available as well?

dmlap: We should. I was shooting for MVP with this PR because the releaser was a bit finicky. Hopefully it will be easy to add the unminified version to this once we get the process working smoothly.

Issue #310, opened 2015-06-16
loadstart event should always come before loadedmetadata event


vasklund:
The `loadstart` event does not always happen before the `loadedmetadata` event, which is the expected behavior. This causes problems for example when trying to override the playlist selection, as discussed in #277. You can see the problem in action at http://output.jsbin.com/haheco/1/ (I can reproduce the problem around 50% of the time).

Today `loadstart` is triggered in Flash ([HTTPVideoProvider:initNetConnection](https://github.com/videojs/video-js-swf/blob/v4.7.0/src/com/videojs/providers/HTTPVideoProvider.as#L434)), and `loadedmetadata` is triggered in both HLS JS ([playlists.on('loadedmetadata') handler](https://github.com/videojs/videojs-contrib-hls/blob/v0.17.2/src/videojs-hls.js#L116)) and Flash ([HTTPVideoProvider:onMetaData](https://github.com/videojs/video-js-swf/blob/v4.7.0/src/com/videojs/providers/HTTPVideoProvider.as#L650)). My understanding is that we want to trigger `loadstart` right before we start downloading the playlists at [videojs-hls.js#L108](https://github.com/videojs/videojs-contrib-hls/blob/v0.17.2/src/videojs-hls.js#L108), but that would mean moving that logic from Flash to JS. It would make sense to only trigger `loadedmetadata` from JS too, as the event triggers twice today.

I'm not sure what other implications/issues apart from #277 there are, but nonetheless this seems like it could cause more problems down the road. I'm not sure how this problem should be approached (it seems code changes are needed both in Flash and contrib HLS), but I wanted to highlight the issue and also ask for feedback in case I'm misunderstanding something.

vasklund:
PR #311 shows a naive approach to solving the issue described above - it's most certainly not a good solution.

dcrockwell:
+1 this is blocking #277, which in turn is blocking our team.

vasklund:
@dcrockwell: There's a **very** crude monkey-patch which works as a short-term workaround if you [really need it](https://gist.github.com/vasklund/2b06aa4ea6595938df51). Notice [line 63](https://gist.github.com/vasklund/2b06aa4ea6595938df51#file-hls-quality-hack-js-L63). Maybe there are better workarounds, but this is what we are forced to used at the moment.

dcrockwell:
@vasklund excellent! Thank you very much.

dmlap:
Fixed in #356

PR #311, opened 2015-06-16
Issue 310 fix example

vasklund:
Waiting for the Flash `loadstart` event solves #310, but adds a delay to the first segment download (in my case around 600 ms, is probably too long).


forbesjo:
Issue is fixed already, closing.

PR #313, opened 2015-06-16
Add the native source handler pattern

dmlap:
The HLS tech should be a source handler itself but until that transition is complete, have it mimic the behavior of the Flash and HTML techs. Create a fallback "native" source handler that will apply the standard behavior when invoked to handle a source. Use player.hls.src_ to detect whether a source is loaded in fillBuffer so that the async nature of currentSrc doesn't delay segment fetching.


imbcmdth:
LGTM :toilet:

dmlap:
This change that caused this was backed out in video.js 4.12.10, so I'm closing this.

PR #307, opened 2015-06-15
pts undefined test unit

vdeshpande:
first commit - apologies if things are messy

PR #321, opened 2015-06-22
Calculate duration by-need from PTS values

dmlap:
Use the time range covered by the union of audio and video track PTS values when calculating duration. For non-live content, this matches the NetStream's duration calculation to within a millisecond or so. Make sure that segment PTS information is preserved when live playlists are refreshed. Remove FlvTag.durationFromTags in favor of Playlist.duration() since the latter handles durations across segments and discontinuities.

Previously, we were not accounting for audio and video tracks which have slightly different durations. We also would introduce tiny discrepancies because we were simply summing per-segment durations and there is a non-zero and variable duration between the last frame of one segment and the first of another. This PR also fixes an issue that caused segment PTS values to be overwritten whenever a live playlist was refreshed, leading to jittery duration calculations for live videos.

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -21,7 +21,7 @@
    * index.
    */
   segmentsDuration = function(playlist, startSequence, endSequence) {
-    var targetDuration, i, segment, expiredSegmentCount, result = 0;
+    var targetDuration, i, j, segment, endSegment, expiredSegmentCount, result = 0;
 
     startSequence = startSequence || 0;
     i = startSequence;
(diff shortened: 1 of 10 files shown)


src/playlist.js
```diff
@@ -36,9 +36,27 @@
     // accumulate the segment durations into the result
     for (; i < endSequence; i++) {
       segment = playlist.segments[i - playlist.mediaSequence];
-      result += segment.preciseDuration ||
```
gkatsev: so, `preciseDuration` gets subsumed into the new duration calculation stuff?

src/playlist.js
```diff
@@ -36,9 +36,27 @@
...
-                segment.duration ||
-                targetDuration;
+
+      // when PTS values aren't available, use information from the playlist
+      if (segment.minVideoPts === undefined) {
```
gkatsev: `minVideoPts` is a function.

src/playlist.js
```diff
@@ -36,9 +36,27 @@
...
+        result += segment.duration ||
+                  targetDuration;
+        continue;
+      }
+
+      // find the last segment with PTS info and use that to calculate
+      // the interval duration
+      for(j = i; j < endSequence - 1; j++) {
+        endSegment = playlist.segments[j - playlist.mediaSequence + 1];
+        if (endSegment.maxVideoPts === undefined ||
```
gkatsev: as is `maxVideoPts`

gkatsev: never mind, I see now.

src/playlist.js
gkatsev: never mind

gkatsev:
LGTM, though, there's an error.
Also, would this affect #307? Seems like the duration and pts updates might change whether we get the issue or not.

dmlap:
Was the error you noted related to maxVideo/AudioPts or something else?

Yes, I believe this change would invalidate any condition that would cause #307.

dmlap:
Missed the Travis issue. I'll update this PR

src/playlist.js
dmlap: Yep. `preciseDuration` didn't account for time-shifted audio and video tracks or the gaps between PTS values across multiple segments. That could add up to tens of milliseconds of error per segment.

dmlap:
Travis was failing because of the incompatibility with video.js 4.12.9.

PR #324, opened 2015-06-24
Reuse the merge commit for the changelog update

dmlap:
Modify the accept-as-is contrib.json task to add the changelog entry to a custom merge commit.


dmlap:
This still requires some testing on my part but I wanted to open up the PR for early feedback.

contrib.json
```diff
@@ -336,21 +336,22 @@
...
+     "accept-as-is": {
       "steps": [
         { "prompt": "text", "id": "prNum",                             "desc": "What is the the pull request number?" },
         { "get": "{{meta.urls.repo_api}}/pulls/{{prNum}}",             "desc": "Get the PR information", "id": "pr" },
         { "get": "{{meta.urls.repo_api}}/pulls/{{prNum}}/commits",     "desc": "Get the PR commits to access author info", "id": "prCommits" },
         [ "git checkout -b {{pr.user.login}}-{{pr.head.ref}} {{pr.base.ref}}", "Create a new branch for merging the changes" ],
-        [ "git pull --rebase {{pr.head.repo.ssh_url}}",                        "Merge and rebase the changes"],
-        [ "grunt test",                                                        "Run tests to make sure they still pass" ],
+        [ "git pull {{pr.head.repo.ssh_url}} {{pr.head.ref}}",                 "Merge in the changes"],
+        [ "git rebase {{pr.base.ref}}",                                        "Rebase against the latest changes in the base branch"],
```
gkatsev: `git pull` does a `fetch` followed by a `merge`, meaning this `rebase` won't be necessary. That's what it was previously `git pull --rebase`.

contrib.json
```diff
@@ -336,21 +336,22 @@
...
+        [ "git checkout {{pr.base.ref}}",                                      "Check out the base branch" ],
+        { "prompt": "confirm",                                         "desc": "Does everything look ok?" },
         { "prompt": "text", "id": "line",                              "desc": "Describe this change in one line" },
+        [ "git merge --no-ff {{pr.user.login}}-{{pr.head.ref}} -m {{line}}. Closes #{{prNum}}", "Merge the changes" ],
```
gkatsev: why not move this one down below `git add CHANGELOG`? Then `git commit --amend` shouldn't be necessary.

dmlap: I didn't realize merge wouldn't complain about that. Good suggestion!

contrib.json
dmlap: I was trying to ensure the change was applied cleanly on top of HEAD, not merged. My understanding is this after-the-fact rebase accomplishes that. You don't think so?

gkatsev: I would just do `git pull --rebase` in the same command.

contrib.json
dmlap: Turns out this would put the changelog commit before the body of the patch, which isn't what we're shooting for. Sticking with `--amend` for now.

PR #326, opened 2015-06-25
Use getComputedStyle instead of player.width

gkatsev:
When upswitching, we try to find the best resolution based on bandwidth
and based on the player dimensions. However, player.width will return
the set width value of the player (https://github.com/videojs/video.js/issues/2282) instead of the current value which means that in fullscreen we don't think we the player changes size and so we don't upswitch.


dmlap:
LGTM

Issue #319, opened 2015-06-19
[live] Seeking forward beyond the 'seekable' window throws an error 


psharmz:
I have been testing seek events and currentTime using the APIs for live streams. 
Chrome and ff handle certain situations differently from Safari

STR:
1. load up a player with a live stream (this is the one I used: http://bcoveliveios-i.akamaihd.net/hls/live/215102/master_english/398/master.m3u8?playerId=255166148200&lineupId=&affiliateId=&pubId=2402232199001&videoId=2552000984001)
2. open up the dev console
3. Play the video
4. Seek in the live stream: in the console set vid.currentTime(x) where x greater than the upper limit of the 'seekable' window
5. Play the video via the UI or vid.play() in the console

In Safari:
The live video doesn't seek to the time you specified since it can't seek that far ahead. Instead it starts  playback from the upper limit of the 'seekable' window.

In Chrome and FF:
Playback stops. An error is thrown in the console: 
```
TypeError: tags[i] is undefined index.js.handlebars:11393:11
```

PR #327, opened 2015-06-25
Seek to a safe position when resuming live playlists

### LINKED ISSUE #318: [live] Resuming playback after 'seekable' duration surpasses currentTime results in spinner
I have been testing seek events and currentTime using the APIs for live streams. 
Chrome and ff handle certain situations differently from Safari

STR:
1. load up a player with a live stream (this is the one I used: http://bcoveliveios-i.akamaihd.net/hls/live/215102/master_english/398/master.m3u8?playerId=255166148200&lineupId=&affiliateId=&pubId=2402232199001&videoId=2552000984001)
2. open up the dev console
3. Pause the video via the UI or vid.pause() in the console
4. Pause until the seekable window surpasses currentTime 
5. Play the video via the UI or vid.play() in the console

In Safari:
The live video doesn't resume playback from where you paused. It starts resumed playback from the lower end of the 'seekable' window. 

In Chrome and FF:
You get the spinner. Playback does not resume unless you seek forward in time. ( this is done in the console by updating currentTime: ex: vid.currentTime(300) ) 

No errors are thrown in the console but eventually you get the video timeout error

dmlap:
If the player has fallen out of the live window, seek back to the earliest safe position when playback resumes. If the player is paused too far ahead of the live point, seek back to the latest safe position when resuming. Fixes #318 and #319.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -365,6 +365,13 @@ videojs.Hls.prototype.setCurrentTime = function(currentTime) {
     return 0;
   }
 
+  // clamp seeks to the available seekable time range
+  if (currentTime < this.seekable().start(0)) {
+    currentTime = this.seekable().start(0);
+  } else if (currentTime > this.seekable().end(0)) {
+    currentTime = this.seekable().end(0);
+  }
+
   // save the seek target so currentTime can report it correctly
   // while the seek is pending
   this.lastSeekedTime_ = currentTime;
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -321,12 +321,21 @@ videojs.Hls.prototype.play = function() {
     this.mediaIndex = 0;
   }
 
-  // seek to the latest safe point in the media timeline when first
-  // playing live streams
-  if (this.duration() === Infinity &&
-      this.playlists.media() &&
-      !this.player().hasClass('vjs-has-started')) {
-    this.setCurrentTime(this.seekable().end(0));
+  // we may need to seek to begin playing safely for live playlists
+  if (this.duration() === Infinity) {
+
+    // if this is the first time we're playing the stream or we're
+    // ahead of the latest safe playback position, seek to the live
+    // point
+    if (!this.player().hasClass('vjs-has-started') ||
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -321,12 +321,21 @@ videojs.Hls.prototype.play = function() {
...
+    // if this is the first time we're playing the stream or we're
+    // ahead of the latest safe playback position, seek to the live
+    // point
+    if (!this.player().hasClass('vjs-has-started') ||
+        this.currentTime() > this.seekable().end(0)) {
+      this.setCurrentTime(this.seekable().end(0));
+
+    } else if (this.currentTime() < this.seekable().start(0)) {
+      // if the viewer has paused and we fell out of the live window,
+      // seek forward to the earliest available position
```
gkatsev: in this case, do we want to seek to the earliest position or to the "3 segments behind live"/"live point" position?

src/videojs-hls.js
```diff
@@ -321,12 +321,21 @@ videojs.Hls.prototype.play = function() {
...
-      !this.player().hasClass('vjs-has-started')) {
-    this.setCurrentTime(this.seekable().end(0));
+  // we may need to seek to begin playing safely for live playlists
+  if (this.duration() === Infinity) {
+
+    // if this is the first time we're playing the stream or we're
+    // ahead of the latest safe playback position, seek to the live
+    // point
+    if (!this.player().hasClass('vjs-has-started') ||
+        this.currentTime() > this.seekable().end(0)) {
```
gkatsev: what about storing seekable end in a variable?

src/videojs-hls.js
dmlap: This way the viewer misses the minimum amount of content relative to when they paused the broadcast. The choice is a bit subjective but I think this is defensible.

src/videojs-hls.js
dmlap: I don't believe calculating seekable end is expensive enough to worry about caching.

dmlap:
Offline approval from @gkatsev

PR #332, opened 2015-06-30
Fix grunt-protractor-runner dependency

forbesjo:
I've added an option to grunt-protractor-runner to install the selenium drivers during the protractor task instead of always installing with grunt-protractor-runner. I added this option for situations where a user would not want to install the drivers in the CI environment.


apadhye:
:+1:

gkatsev:
LGTM

PR #333, opened 2015-06-30
Don't prepublish in Travis

forbesjo:
The Travis's Sauce Connect doesn't get to start between npm install and prepublish. Only prepublish locally.


gkatsev:
Sure. :P

Issue #314, opened 2015-06-17
Duration accuracy


dmlap:
When attempting to translate ID3 tags specified at PTS values to playback positions, I realized that the trustworthiness of segment duration calculations can have a big impact on the viability of many implementation strategies. I have done some research into different methods for calculating duration and playback position and will collect that information here.

dmlap:
I ran all of the bipbop segments through ffmpeg/ffprobe and output their durations. I compared our duration with the duration reported by the tool and in each case, we differed by 33ms. This happens to be the PTS interval between the last two tags, which we use to estimate the frame display duration.  Related, the frame duration we measure, 33.3555... ms, is very close to the average framerate (tbr) reported by ffprobe, 29.97 or (1000/29.97) = 33.3667...ms.

The next question was whether the time value reported by Flash matched the numbers I saw using our segment parser and ffprobe. In the three cases I tested, the video duration reported by Flash matched the ffprobe duration. That is, it was 33ms shorter than the duration number we were calculating.

I put together a small playlist and checked the time reported by Flash when it completed to make sure that the final frame duration wasn't re-introduced when the segments were stitched. Strangely enough, Flash reported a total duration of 29.996s, while ffprobe and segment parsing came up with ~30.21s. As a final sanity check, I used high resolution timers to measure the duration of the time between `playing` and `ended` for my short playlist, which reported 29.957s.

dmlap:
Interestingly, the m3u8 for the short playlist I used for testing included duration numbers that were closer to the duration reported by Flash and timers. On a hunch, I decided to try calculating the duration numbers for the audio and video streams separately instead of mixing them together. As it turns out, the elementary streams in those files don't have matching durations. The minimum duration between the audio and video streams matches up precisely with the durations the segmenter wrote out in the M3U8. For this playlist, summing the durations in the segmenter gives a duration of 29.9334s.

dmlap:
## Summary
                  | contrib-hls | ffprobe | Flash | m3u8 | minimum stream duration
-------------- | -------------- | --------- | -------- | ------- | -----------------------------------
single TS | 10.10580 | 10.07 | 10.072 | 9.98458 | 9.98458
short m3u8 | 30.21 | 30.21 | 29.996 | 29.9334 | 29.9334

[...]

PR #339, opened 2015-07-06
More duration updates

dmlap:
Pulled in #279 so that seeking "between segments" does not cause an error. Updated duration calculation so that PTS-based results are used for the largest span of segments possible. Previously, PTS values were only used to calculate duration when they formed a continuous interval. This is a lot more conservative than necessary because PTS values are directly comparable unless there is a discontinuity between their containing segments. This method should result in more accurate calculations and get rid of some unnecessary segment iteration.

### DIFF
--- a/src/m3u8/m3u8-parser.js
+++ b/src/m3u8/m3u8-parser.js
@@ -375,7 +375,8 @@
 
     // the manifest is empty until the parse stream begins delivering data
     this.manifest = {
-      allowCache: true
+      allowCache: true,
+      discontinuityStarts: []
     };
 
     // update the manifest with the m3u8 entry from the parse stream
@@ -513,6 +514,7 @@
             },
             'discontinuity': function() {
               currentUri.discontinuity = true;
+              this.manifest.discontinuityStarts.push(uris.length);
             },
             'targetduration': function() {
               if (!isFinite(entry.duration) || entry.duration < 0) {
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -5,39 +5,17 @@
   'use strict';
 
   var DEFAULT_TARGET_DURATION = 10;
-  var duration, seekable, segmentsDuration;
+  var accumulateDuration, duration, seekable, segmentsDuration;
 
-  /**
-   * Calculate the media duration from the segments associated with a
-   * playlist. The duration of a subinterval of the available segments
-   * may be calculated by specifying a start and end index.
-   *
-   * @param playlist {object} a media playlist object
-   * @param startSequence {number} (optional) an inclusive lower
-   * boundary for the playlist.  Defaults to 0.
-   * @param endSequence {number} (optional) an exclusive upper boundary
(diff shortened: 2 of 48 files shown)


dmlap:
Also, closes #314

src/m3u8/m3u8-parser.js
```diff
@@ -375,7 +375,8 @@
 
     // the manifest is empty until the parse stream begins delivering data
     this.manifest = {
-      allowCache: true
+      allowCache: true,
+      discontinuityStarts: []
```
gkatsev: so `discontinuityStarts` end up being a list of indexes pointing at segments that begin a discontinuity?

src/videojs-hls.js
```diff
@@ -1148,14 +1160,15 @@ videojs.Hls.getMediaIndexByTime = function() {
...
+  var index, time = 0, segment;
 
   if (!playlist.segments || mediaIndex === 0) {
     return 0;
   }
 
   for (index = 0; index < mediaIndex; index++) {
-    time += playlist.segments[index].duration;
+    segment = playlist.segments[index];
+    time += segment.preciseDuration || segment.duration || playlist.targetDuration || 0;
```
gkatsev: I thought we don't have `preciseDuration` anymore?

src/playlist.js
```diff
@@ -5,7 +5,130 @@
...
+   * segments to examine.
+   * @param end {number} an exclusive upper boundary for the segments
+   * to examine.
+   * @param includeTrailingTime {boolean} if false, the interval between
+   * the final segment and the subsequent segment will not be included
+   * in the result
+   * @return {number} the duration between the start index and end
+   * index in seconds.
+   */
+  accumulateDuration = function(playlist, start, end, includeTrailingTime) {
```
gkatsev: so all this stuff tries to find those overlapping ranges of audio and video durations and take the greatest of them?

src/videojs-hls.js
dmlap: I believe this function got re-added somewhere along the way. It should be removed entirely.

src/playlist.js
dmlap: Basically, yes. It tries to find the longest span of segments that have PTS info available at the start and end. It does this by:

1. Splitting the interval into ranges that do not have internal discontinuities
1. For each of these ranges, finding the earliest and latest segment with PTS info
 1. Adding up the differences between the earliest PTS and the latest PTS in each range
 1. Adding up the M3U8 based durations for any segments in the range that aren't in the interval above

src/m3u8/m3u8-parser.js
dmlap: Yes. It makes splitting intervals into continuous spans a lot simpler.

src/videojs-hls.js
gkatsev: looks good other than this.

PR #345, opened 2015-07-09
Reset time to 0 on play after end instead of just mediaIndex

bc-bbay:
Replaces #294 (bad rebase)

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -318,8 +318,8 @@ videojs.Hls.prototype.setupMetadataCueTranslation_ = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    // If this.lastSeekedTime_ is defined, a seek has happened after playback ended, as it is set undefined at video end.
-    // We should begin playback from that point.
+    // lastSeekedTime_ is set undefined when the video ends. It changes to 'null' when the user seeks.
+    // If it hasn't changed after the video ended, that means we need to reset the current time to 0 before playing.
     if (this.lastSeekedTime_ === undefined) {
       this.setCurrentTime(0);
     }
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -318,7 +318,11 @@ videojs.Hls.prototype.setupMetadataCueTranslation_ = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    this.mediaIndex = 0;
+    // If this.lastSeekedTime_ is defined, a seek has happened after playback ended, as it is set undefined at video end.
+    // We should begin playback from that point.
+    if (this.lastSeekedTime_ === undefined) {
+      this.setCurrentTime(0);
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -950,6 +954,7 @@ videojs.Hls.prototype.drainBuffer = function(event) {
   // the playlist
   if (this.duration() !== Infinity && mediaIndex + 1 === playlist.segments.length) {
     this.mediaSource.endOfStream();
+    this.lastSeekedTime_ = undefined;
```
dmlap: Why is this line necessary?

test/videojs-hls_test.js
```diff
@@ -2254,6 +2260,8 @@ test('calling play() at the end of a video resets the media index', function() {
 
   player.play();
   strictEqual(player.hls.mediaIndex, 0, 'index is 1 after the first segment');
+
+  player.el().querySelector('.vjs-tech').vjs_getProperty = function() {};
```
dmlap: This can be rewritten a bit more succinctly as `player.tech.el().vjs_getProperty`

src/videojs-hls.js
```diff
@@ -318,7 +318,11 @@ videojs.Hls.prototype.setupMetadataCueTranslation_ = function() {
  */
 videojs.Hls.prototype.play = function() {
   if (this.ended()) {
-    this.mediaIndex = 0;
+    // If this.lastSeekedTime_ is defined, a seek has happened after playback ended, as it is set undefined at video end.
+    // We should begin playback from that point.
```
dmlap: The code below makes sense but this comment confuses me.

src/videojs-hls.js
bc-bbay: So that the condition above returns true, and we know to reset. It's not wiped out automatically at the end of playback, but continues to store the last seeked value (if the user ever seeked at all).

src/videojs-hls.js
bc-bbay: I'll reword it.

test/videojs-hls_test.js
bc-bbay: Will update

src/videojs-hls.js
bc-bbay: Sorry, you're right, that's both the correct behavior and what is actually happening. But null != undefined and still affects the reset at post-end playback.

dmlap: In that case, I think we should be consistent about what value we choose for `lastSeekedTime_` to indicate "unset". What do you think about initializing it to `null` in the init function and then just checking for null?

bc-bbay: Would we then not unset to null after a seek completes? Because the question I'm asking is, "has there been a seek since the video ended", and if video ending and seek completing do the same thing, I can't get the answer.

dmlap: If you seek once the video has ended, `ended()` should be false.

PR #346, opened 2015-07-13
Don't assume media sequence starts at zero

dmlap:
Zero is the default when unspecified but it's legal to have a one-indexed VOD media playlist. When the media sequence is non-zero, assume that should be the basis for all sequence calculations unless a different basis was explicitly specified. Fix Playlist.seekable so that it doesn't attempt to estimate the duration of expired segments.


gkatsev:
LGTM

PR #348, opened 2015-07-13
Duration should work when video or audio is missing

### LINKED ISSUE #341: Streaming fails for videos with no audio with 0.17.3 release
Hello, I tried to update videojs-hls but I am hitting a problem with files that do not include audio tracks. I isolated a simple test case that compares the behavior with a working version of videojs-hls (0.11) and the latest one.

Below is a screnshot of the stack trace I see:

screenshot 2015-07-07 17 56 24

dmlap:
If audio or video data is missing for an HLS stream, duration calculations should just use the info that is available. Fixes #341.

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -5,7 +5,23 @@
   'use strict';
 
   var DEFAULT_TARGET_DURATION = 10;
-  var accumulateDuration, ascendingNumeric, duration, intervalDuration, rangeDuration, seekable;
+  var accumulateDuration, ascendingNumeric, duration, intervalDuration, optionalMin, optionalMax, rangeDuration, seekable;
+
+  // Math.min that will return the alternative input if one of its
+  // parameters in undefined
+  optionalMin = function(left, right) {
+    left = isFinite(left) ? left : Infinity;
+    right = isFinite(right) ? right : Infinity;
+    return Math.min(left, right);
+  };
+
+  // Math.max that will return the alternative input if one of its
+  // parameters in undefined
+  optionalMax = function(left, right) {
+    left = isFinite(left) ? left: -Infinity;
+    right = isFinite(right) ? right: -Infinity;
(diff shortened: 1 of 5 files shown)


gkatsev:
LGTM

test/videojs-hls_test.js
```diff
@@ -1044,6 +1050,64 @@ test('records the min and max PTS values for a segment', function() {
...
+  tags.push({ pts: 10, bytes: new Uint8Array(1) });
+  standardXHRResponse(requests.pop()); // segment 0
+
+  equal(player.hls.playlists.media().segments[0].minVideoPts, 0, 'recorded min video pts');
+  equal(player.hls.playlists.media().segments[0].maxVideoPts, 10, 'recorded max video pts');
+  strictEqual(player.hls.playlists.media().segments[0].minAudioPts, undefined, 'min audio pts is undefined');
+  strictEqual(player.hls.playlists.media().segments[0].maxAudioPts, undefined, 'max audio pts is undefined');
+});
+
+test('records PTS values for video-only segments', function() {
```
bclwhitaker: audio-only

dmlap: Oops

src/playlist.js
```diff
@@ -91,7 +107,8 @@
     // available PTS information
     for (left = range.start; left < range.end; left++) {
       segment = playlist.segments[left];
-      if (segment.minVideoPts !== undefined) {
+      if (segment.minVideoPts !== undefined ||
```
FDiskas: Still getting error `TypeError: Cannot read property 'minVideoPts' of undefined`

dmlap: The entries of a media playlist should be segment objects-- `[undefined]` is not a legal value for that array. Side note: it would be better to file this sort of thing as an issue. Line comments might get overlooked.

PR #349, opened 2015-07-14
Make hls.stopCheckingBuffer and hls.startCheckingBuffer public

ntadej:
It may be useful to have hls.startCheckingBuffer and hls.stopCheckingBuffer public.

Use case:
We do not support our live streams to be paused over longer period of time, but it is still nicer to keep last frame and not load the playlists or segments. The following combination of commands does the desired for us:
```
hls.playlists.dispose();
hls.cancelSegmentXhr();
hls.stopCheckingBuffer();
```

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -44,7 +44,7 @@ videojs.Hls = videojs.Flash.extend({
     this.segmentBuffer_ = [];
     // periodically check if new data needs to be downloaded or
     // buffered data should be appended to the source buffer
-    this.startCheckingBuffer_();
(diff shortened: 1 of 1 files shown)


dmlap:
Once you dispose the playlist loader, won't it be impossible to rejoin the stream?

ntadej:
The stream will be invalid anyway so a new source will be used. There playlist loader will be initiated again.

forbesjo:
Could you reopen this PR against the development branch, we've done major changes that may make this enhancement incompatible. Also, this request isn't totally clear for us, could you expand upon why this is important to you?

Issue #351, opened 2015-07-21
autoplay attribute causes start playing at very beginning segment of live chunklist


eminden:
when playing live video and autoplay attribute exists on video tag, videojs-hls-contrib sometimes start downloading the very first segment listed in downloaded chunklist instead of the latest segments. 

tested with latest videojs-hls-contrib # v0.17.5

dmlap:
You are correct and I happen to be working on this right now :) Thanks for the report!

dmlap:
Should be fixed in 0.17.6.

PR #353, opened 2015-07-23
Live seeking with autoplay

dmlap:
Depending on the order the media source opened, the media playlist was recieved, and play was invoked, live streams may or may not have been seeked. Consolidate the logic that checks for this condition to ensure we always seek to the live point. Add a check so that if someone seeks before the media source opens, the player does not throw an exception. For #351.

I'm seeing issues with quality switches that occur immediately after the first media playlist loads. Still investigating.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -35,6 +35,14 @@ videojs.Hls = videojs.Flash.extend({
     options.source = source;
     this.bytesReceived = 0;
 
+    this.hasPlayed_ = false;
+    this.on(player, 'loadstart', function() {
+      this.hasPlayed_ = false;
+      this.one(this.mediaSource, 'sourceopen', this.setupFirstPlay);
+    });
+    this.on(player, ['play', 'loadedmetadata'], this.setupFirstPlay);
+
+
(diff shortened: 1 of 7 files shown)


src/videojs-hls.js
```diff
@@ -679,7 +718,7 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
   // being buffering so we don't preload data that will never be
   // played
   if (!this.playlists.media().endList &&
-      !this.player().hasClass('vjs-has-started') &&
+      !player.hasClass('vjs-has-started') &&
```
misteroneill: Most places are checking `hasPlayed_` except here. Wondering if that's intentional or, at least, not harmful.

misteroneill:
Pending resolution of my one inline comment, this LGTM.

src/videojs-hls.js
dmlap: I believe we'll be switching all of the `vjs-has-started` to a mechanism like `hasPlayed_` with the video.js 5.x update. I was trying to minimize the surface area of this particular PR.

PR #355, opened 2015-07-24
Handle when there is no audio stream to create the VTTCue

jsanford-bc:
VTTCues fail when there is no audio stream, which causes the player to halt and not play back. This avoids that issue by making sure there is a minAudioPts and minVideoPts.


forbesjo:
Should be fixed in the development branch.

PR #356, opened 2015-07-28
Trigger loadstart

imbcmdth:
The swf will no longer emit a loadstart event when it is running in "data generation mode" so lets do it ourselves!

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "karma-sauce-launcher": "~0.1.8",
     "qunitjs": "^1.15.0",
     "sinon": "1.10.2",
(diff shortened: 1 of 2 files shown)


gkatsev:
LGTM

imbcmdth:
@dmlap LGTY?

dmlap:
@imbcmdth What version of the SWF is this tied to? Are we pointing to the right video.js version in our package.json to pick that up?

dmlap:
Found it. Dependent on https://github.com/videojs/video-js-swf/commit/08fa8f1fd56c0c63570f8b9d25efd1e7e7e72f1e

PR #361, opened 2015-08-10
Point back to grunt-protractor-runner

forbesjo:
The option to automatically update the selenium drivers got merged into grunt-protractor-runner so I'm pointing back to the original repo.


gkatsev:
LGTM

Issue #277, opened 2015-05-22
Overriding first playlist selection


vasklund:
We are having problems with the initial segment almost always being downloaded in the lowest bitrate, even when available bandwidth allows for playback at higher quality. This is a pretty big problem for short clips (a lot of our clips are around 30-120 sec long).

I know we can override `selectPlaylist` to implement our own selection algorithm, or set `player.hls.bandwidth` higher to make `selectPlaylist` return a higher quality playlist. The problem is that the code in `this.playlists.on('loadedmetadata', function { ... })`, which is responsible for selecting the first playlist, will not respect the playlist returned from `selectPlaylist` if the estimated download time for the first segment is more than 10 seconds. This seems reasonable, but the problem is that it will then fallback to the **lowest** bitrate, instead of selecting the highest quality playlist that satisfies both available bandwidth and initial load time (10 seconds).

Essentially, we want to trade high quality (from the first segment) for longer startup time, but there is no good way to do so now. I'm experimenting with some alternate approaches and will try to prepare a pull request today, but I'm interested in getting some feedback.

vasklund:
https://github.com/videojs/videojs-contrib-hls/blob/master/src/videojs-hls.js#L193

vasklund:
Ok, It seems I'm able to sidestep the problem by setting the playlist manually with `player.hls.playlists.media`:

```
player.ready(function () {
  // force the fourth playlist for the initial segment
  player.hls.playlists.media(player.hls.playlists.master.playlists[3]);
});
```

This seems to work and I can leave player.hls.selectPlaylist and player.hls.bandwidth alone, but it seems kind of counterintuitive that I need to use `player.hls.playlists.media()` to set the quality for the first segment. I thought I would be able to achieve similar results by overriding `player.hls.selectPlaylist()`, but it seems I can't override it before it gets called (if I override it in player.ready() it has already been called once for the initial segment).

I'll comment again if I find anything else, but for now I might just go with the media() workaround.

dmlap:
Did you notice that you can influence initial rendition selection by [overriding the bandwidth estimate](https://github.com/videojs/videojs-contrib-hls#playerhlsbandwidth) on `loadstart`?

```js
player.on('loadstart', function() {
  // choose a really high quality rendition, even if bandwidth estimates don't justify it
  player.hls.bandwidth = 1e9;
});
```
[...]

PR #362, opened 2015-08-11
Revert "Trigger loadstart"

imbcmdth:
This reverts commit 099f282bdd43441c27a0f106db9cf2e620cc17b3.

The 4.x version of video-js-swf was reverted to previous behavior for video.js 4.x so videojs-contrib-hls should no longer be emitting the loadstart event itself. These changes will be included in 5.x.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -107,16 +107,6 @@ videojs.Hls.prototype.src = function(src) {
     this.playlists.dispose();
   }
 
-  // We need to trigger this asynchronously to give others the chance
-  // to bind to the event when a source is set at player creation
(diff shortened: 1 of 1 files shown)


dmlap:
lgtm

dmlap:
This will reopen #277 and #310 until we finish the upgrade to video.js 5. I'm *very* sorry about the false notifications for those issues but we don't want to make a breaking change without following semantic versioning and we made a decision awhile ago to close 4.x to anything but patch releases. (That said, you can continue to use videojs-contrib-hls v0.17.7 with videojs-swf v4.7.3. We won't be removing them, just reverting this breaking change.)

PR #367, opened 2015-08-24
Add loadstart 5

imbcmdth:
The swf will no longer emit a loadstart event when it is running in "data generation mode" so must do so ourselves.


imbcmdth:
Rebased to pull in test fixes from @dmlap..

PR #368, opened 2015-08-24
Build enhancement

imbcmdth:
Output (`dist/videojs.hls.js`) includes:
* `videojs-contrib-media-sources` with `mux.js` as an inline blob webworker
* `videojs-contrib-hls`


Gruntfile.js
```diff
@@ -24,17 +24,11 @@ module.exports = function(grunt) {
       },
       dist: {
         nonull: true,
-        src: ['src/videojs-hls.js',
+        src: ['node_modules/videojs-contrib-media-sources/dist/videojs-media-sources.js',
```
gkatsev: I still hate this so much.

gkatsev: (The fact that we need to list out all our files, that is).

forbesjo: Would browserify make this simplier?

gkatsev: Yes.

.gitignore
```diff
@@ -1,5 +1,4 @@
 .DS_Store
-dist/*
```
forbesjo: Why don't shouldn't the dist just live on npm?

gkatsev: Yeah, re-add it here and get rid of it in `.npmignore`.

PR #370, opened 2015-08-26
Flush audio on seek to prevent wrong audio PTS

ntadej:
Flush audio on seek to prevent first audio tag in seeked segment having the same PTS as it was the last from previously processed segment. This fixes a lot of seek issues for us.

### DIFF
--- a/src/aac-stream.js
+++ b/src/aac-stream.js
@@ -29,6 +29,10 @@ window.videojs.Hls.AacStream = function() {
 
   this.tags = [];
 
+  this.flush = function() {
+    state = 0;
+  };
+
   // (pts:uint):void
   this.setTimeStampOffset = function(pts) {
 
@@ -45,7 +49,7 @@ window.videojs.Hls.AacStream = function() {
 
     // If data is aligned, flush all internal buffers
     if (dataAligned) {
-      state = 0;
+      this.flush();
     }
   };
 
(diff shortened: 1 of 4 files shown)


dmlap:
Thanks! Is there a good way for us to reproduce the audio issues you were encountering?

As a side note, we've been working for the past couple weeks on extracting the segment parser and related transmuxing tools into [a separate project](https://github.com/dmlap/mux.js) as part of adding HTML-based HLS playback. Once we get that project into a more stable state I'll be moving it out of my own organization somewhere more public. I hope we don't have this problem with the new TS to fragmented MP4s tools but if you have a free minute to check out the Media Source Extensions PR (coming any day now...), I'd love to hear how well HTML playback works for you.

ntadej:
I'm not sure if this is specific to our streams, probably we can provide you with a sample. These are PVR recordings made from same source as our live stream (that I already sent you a while ago).
Steps to reproduce:
 - play video for a while so it buffers a few segments
 - seek to position into future -> it works OK and plays from that position, but min audio PTS on first segment will be wrong (it will be max audio PTS from last buffered segment)
 - seek to another position later than previous seek -> it will seek to same position again, because PTS in previously seeked position is wrong and so duration of segment too

When calculating duration of playlist you use `segment.maxPts - segment.minPts` (simplified) which causes wrong seek position.

I will look at Media Source Extensions PR when ready and report back.

ntadej:
We prepared samples of our PVR and VOD videos. This and #379 can be reproducible with first. I sent them over email to @dmlap.

forbesjo:
If this is still an issue could you open a PR/issue against mux.js?

PR #373, opened 2015-08-31
Adjustable live buffer

bc-bbay:
A customer wants to be able to customize the length of the live delay, contrary to the HLS specification. This adds functions to allow it.

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -269,7 +269,7 @@
     // of content from the end of the playlist
     // https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-6.3.3
     if (!playlist.endList) {
-      liveBuffer = targetDuration * 3;
+      liveBuffer = targetDuration * player.hls.liveNumTargetDurations();
       // walk backward from the last available segment and track how
       // much media time has elapsed until three target durations have
       // been traversed. if a segment is part of the interval being
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -5,7 +5,7 @@
   'use strict';
 
   var DEFAULT_TARGET_DURATION = 10;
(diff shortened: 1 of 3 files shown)


gkatsev:
Why not expose it like we have for [`bandwidth`](https://github.com/videojs/videojs-contrib-hls#playerhlsbandwidth) and use it as a variable that way?

src/playlist.js
```diff
@@ -236,6 +236,21 @@
...
+    if (!playlist.liveTargetDurations_) {
+      playlist.liveTargetDurations_ = 3;
+    }
+    return playlist.liveTargetDurations_;
+  };
+
+  setLiveTargetDurations = function(playlist, numTargetDurations) {
+    if (numTargetDurations > 1) {
+      playlist.liveTargetDurations_ = numTargetDurations;
+    } else {
```
bc-bbay: @gkatsev To answer your question, this was basically the reason for the getter/setter structure. I want to make sure this value is at least 1, since making it anything less doesn't make sense and breaks the code. It seemed easier/smarter to police the value in a setter than in every place it's referenced forever and always.

dmlap:
I'm ok with exposing this as a constant (a la [GOAL_BUFFER_LENGTH](https://github.com/videojs/videojs-contrib-hls/blob/220d55f679c9706a1afb0ccb0234b27bd80ae87b/src/videojs-hls.js#L65)) so someone *could* overwrite it if they really wanted to. I don't think it's a good idea for anyone to touch this value and I think a full-fledged setter/getter with validation implies a level of support we don't actually provide. Would you mind trying that route instead?

bc-bbay:
More like #377? I'll close whichever of these you prefer @dmlap

Issue #303, opened 2015-06-10
Use WebkitDataCue for ID3 tags?


dmlap:
Safari creates a WebkitDataCue for in-band metadata track cues. The constructor is public and usable. Should we do the same when it's available or stick with VTTCue?

dmlap:
@gkatsev any thoughts on this one?

gkatsev:
Sounds like this might be the correct course of action. Safari already has this and Chrome has filed an intent to implement (https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/U06zrT2N-Xk/RFS_q-Hi2wMJ).
The spec is available here: http://www.w3.org/html/wg/drafts/html/master/semantics.html#datacue
Seems right now that it required to be used for metadata tracks but are trying to remove that restriction: https://www.w3.org/Bugs/Public/show_bug.cgi?id=25261

saxena-gaurav:
This is getting addressed in #395

PR #375, opened 2015-09-01
make parsed id3 cue json structure same as safari

saxena-gaurav:


src/videojs-hls.js
```diff
@@ -331,7 +331,7 @@ videojs.Hls.prototype.addCuesForMetadata_ = function(segmentInfo) {
       frame = metadata.frames[i];
       time = segmentOffset + ((metadata.pts - minPts) * 0.001);
       cue = new window.VTTCue(time, time, frame.value || frame.url || '');
```
gkatsev: we probably should also see if we can use a `DataCue` instead of a `VTTCue` for this.

test/metadata-stream_test.js
```diff
@@ -233,7 +233,7 @@
 
     equal(events.length, 1, 'parsed one tag');
     equal(events[0].frames.length, 1, 'parsed one frame');
-    equal(events[0].frames[0].id, 'WXXX', 'parsed the frame id');
```
bclwhitaker: would be nice if all the descriptions were updated to match the new property too

saxena-gaurav: done.

src/videojs-hls.js
saxena-gaurav: WebKitDataCue is not available in Chrome/FireFox. We have to stick with VTTCue for now. There is an issue filed for this: https://github.com/videojs/videojs-contrib-hls/issues/303

imbcmdth:
You probably want to make this patch to the `development` branch as well and submit a PR. That branch is the massive ball of changes for the HLS plugin for VJS 5!

imbcmdth:
LGTM :+1:

src/videojs-hls.js
dmlap: You can detect whether DataCue or VTTCue is available and use the appropriate one where available:
```js
new (DataCue || WebKitDataCue || VTTCue)(0, 1, 'hi this is text');
```

dmlap:
This would be a backwards incompatible change. Can we keep the values at the old property names as well and log a deprecation warning if they're accessed?

saxena-gaurav:
@dmlap : Thanks for the review. I have created new PR #386. Closing this PR.

PR #377, opened 2015-09-02
Make the number of target durations a constant

bc-bbay:
For overridability as needed

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -269,7 +269,7 @@
     // of content from the end of the playlist
     // https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-6.3.3
     if (!playlist.endList) {
-      liveBuffer = targetDuration * 3;
+      liveBuffer = targetDuration * videojs.Hls.NUM_TARGET_DURATIONS;
       // walk backward from the last available segment and track how
       // much media time has elapsed until three target durations have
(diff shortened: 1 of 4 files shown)


dmlap:
Are the seekable tests checking this anywhere? Are they using a hard-coded value still? Otherwise, looks good.

src/playlist.js
```diff
@@ -269,7 +269,7 @@
     // of content from the end of the playlist
     // https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-6.3.3
     if (!playlist.endList) {
-      liveBuffer = targetDuration * 3;
+      liveBuffer = targetDuration * videojs.Hls.NUM_TARGET_DURATIONS;
```
bclwhitaker: Would it make sense to add an `|| 1` here?

src/videojs-hls.js
```diff
@@ -219,7 +222,7 @@ videojs.Hls.getMediaIndexForLive_ = function(selectedPlaylist) {
 
   var tailIterator = selectedPlaylist.segments.length,
       tailDuration = 0,
-      targetTail = (selectedPlaylist.targetDuration || 10) * 3;
+      targetTail = (selectedPlaylist.targetDuration || 10) * videojs.Hls.NUM_TARGET_DURATIONS;
```
bclwhitaker: ditto

PR #379, opened 2015-09-03
Properly calculate PTS difference between segments

ntadej:
Another seek related improvement. Our streams do not have audio and video aligned so PTS differ by around 1-1.5 s. Video is played OK, but segment duration is calculated incorrectly as once min was used and another time max. From the tests I read that you want the largest possible duration. I propose using maximum difference between video PTS difference and audio PTS difference. This fixes our problems (still have to thoroughly test for negative side effects).

Tests are currently broken as I'm not sure what was your case when writing them.

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -141,10 +141,8 @@
     // add in the PTS interval in seconds between them
     if (right >= left) {
       result += 0.001 *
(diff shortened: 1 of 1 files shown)


forbesjo:
This calculation is totally redone for contrib-hls 1.0, could you see if this issue still exists in the development branch?

PR #381, opened 2015-09-07
fixes bug (ZEN-TOTAL-DURATION)

KorsaR-ZN:
Fixed a bug that the player does not parse tag ZEN-TOTAL-DURATION

### DIFF
--- a/src/m3u8/m3u8-parser.js
+++ b/src/m3u8/m3u8-parser.js
@@ -126,7 +126,7 @@
     }
 
     // Comments
-    if (line.indexOf('#EXT') !== 0) {
+    if (line.indexOf('#EXT') !== 0 && line.indexOf('#ZEN-') !== 0) {
       this.trigger('data', {
         type: 'comment',
         text: line.slice(1)
--- a/test/manifest/absoluteUris.js
+++ b/test/manifest/absoluteUris.js
@@ -21,6 +21,7 @@
     }
   ],
   "targetDuration": 10,
+  "totalDuration": 57.9911,
   "endList": true,
   "discontinuitySequence": 0,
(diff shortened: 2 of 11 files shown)


dmlap:
Thanks! It looks like this change would cause some of our existing test cases to need an update. In [test/manifest](https://github.com/videojs/videojs-contrib-hls/tree/master/test/manifest) there are a bunch of example m3u8 files and a corresponding parsed version that ends in .js instead of .m3u8. Would you be up for giving that a shot?

KorsaR-ZN:
@dmlap Yes. tomorrow I will fix

src/m3u8/m3u8-parser.js
```diff
@@ -126,7 +126,7 @@
     }
 
     // Comments
-    if (line.indexOf('#EXT') !== 0) {
+    if (line.indexOf('#EXT') !== 0 && line.indexOf('#ZEN-') !== 0) {
```
ntadej: Shouldn't there be `||` instead of `&&`? Your fix changes existing behaviour.

KorsaR-ZN: @ntadej Not
No. It's all right

example:
```
var line = "#EXT";
console.log((line.indexOf("#EXT") !== 0 && line.indexOf("#ZEN-") !== 0)) // false
console.log((line.indexOf("#EXT") !== 0 || line.indexOf("#ZEN-") !== 0)) // true
```

KorsaR-ZN:
@dmlap Please help me fix tests

forbesjo:
We won't be making anymore changes in master at this time unless they're critical. Could you please open a new PR against development?

PR #384, opened 2015-09-10
make id3 cue parser json structure similar to safari

saxena-gaurav:


saxena-gaurav:
This PR is against the development branch. Same as #375 which is against master branch.

PR #385, opened 2015-09-10
Pass codecs along into source buffer creation

dmlap:
If a codecs attribute is present on a variant stream, use it when adding a source buffer. Wait to create the source buffer until the variant playlist is downloaded and the media source is open.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
+sudo: false
 node_js:
-- '0.10'
+  - "stable"
 install:
 - npm install -g grunt-cli && npm install
 notifications:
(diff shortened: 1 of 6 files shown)


imbcmdth:
Just one question:

Should we handle the situation where the codecs change between renditions? Is that even allowed?

dmlap:
I struggled with that myself. I think assuming the codecs are constant seems like the right move initially, mainly because it would be a real pain to have to deactivate source buffers and add new ones between renditions. It's something we could add later if it turns out to be popular (though I really hope no one actually does that).

gkatsev:
:+1: to handling that later

PR #387, opened 2015-09-14
Parse media groups from m3u8s

dmlap:
The [EXT-X-MEDIA](https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-4.3.4.1) tag allows alternate "Renditions" to be associated with streams. A rendition could be an alternate camera angle or a dubbed audio track in an another language, for instance. As the first step toward adding support for handling streams with multiple alternative renditions, parse mediaGroup objects out of M3U8 files.

An [informal description of media groups](https://developer.apple.com/library/ios/technotes/tn2288/_index.html#//apple_ref/doc/uid/DTS40012238-CH1-ALTERNATE_MEDIA) is available from Apple.


dmlap:
Merged into development (after much rebasing)

PR #389, opened 2015-09-16
Discontinuity

imbcmdth:
These are the changes required to support discontinuities in playlists using MSE.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -25,7 +25,7 @@ keyFailed = function(key) {
 };
 
 videojs.Hls = videojs.extends(Component, {
-  constructor: function(tech, source) {
+  constructor: function(tech, options) {
     var self = this, _player;
 
     Component.call(this, tech);
@@ -44,7 +44,8 @@ videojs.Hls = videojs.extends(Component, {
       }
     }
     this.tech_ = tech;
-    this.source_ = source;
+    this.source_ = options.source;
+    this.mode_ = options.mode;
     this.bytesReceived = 0;
 
     // loadingState_ tracks how far along the buffering process we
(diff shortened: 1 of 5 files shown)


src/videojs-hls.js
```diff
@@ -785,9 +815,10 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
     return;
   }
 
+  // To determine how much is buffered, we need to find the buffered region we
+  // are currently playing in and measure it's length
   if (buffered && buffered.length) {
-    // assuming a single, contiguous buffer region
-    bufferedTime = tech.buffered().end(0) - tech.currentTime();
+    bufferedTime = Math.max(0, buffered.end(0) - currentTime);
```
imbcmdth: We can't assume a single region any more so we have to be smarter about calculating the amount of buffered data.

dmlap: Nice.

src/videojs-hls.js
```diff
@@ -739,6 +739,35 @@ videojs.Hls.prototype.stopCheckingBuffer_ = function() {
...
+    for (i = 0;i < buffered.length; i++) {
+      if (buffered.start(i) <= currentTime &&
+          buffered.end(i) >= currentTime) {
+        return videojs.createTimeRange(buffered.start(i), buffered.end(i));
+      }
+    }
+
+    // Just return the first range if one was not found that contain
+    // the play-head
+    return videojs.createTimeRange(buffered.start(0), buffered.end(0));
```
dmlap: This default seems a bit weird given the name of the method.

test/videojs-hls_test.js
```diff
@@ -87,24 +87,24 @@ var
...
     };
-    mockTech(player.tech);
+    mockTech(player.tech_);
 
     return player;
   },
   openMediaSource = function(player) {
     // ensure the Flash tech is ready
-    player.tech.triggerReady();
+    player.tech_.triggerReady();
```
dmlap: Not for this PR but we should just focus on the tech for these tests in the future (i.e. don't create a player at all).

dmlap:
Other than my comment about the negative-case for `findCurrentBuffered_`, LGTM.

PR #395, opened 2015-09-24
make id3 cue structure similar to safari

saxena-gaurav:


src/videojs-hls.js
```diff
@@ -392,6 +392,37 @@ videojs.Hls.prototype.addCuesForMetadata_ = function(segmentInfo) {
   minPts = Math.min(isFinite(segment.minVideoPts) ? segment.minVideoPts : Infinity,
                     isFinite(segment.minAudioPts) ? segment.minAudioPts : Infinity);
 
+  Cue = window.WebKitDataCue || window.VTTCue;
+
+  deprecateOldCue = function(cue) {
+    if (cue.frame.id) {
+      Object.defineProperty(cue.frame, 'id', {
```
dmlap: It looks like these statements are overwriting the values at `cue.frame.whatever` with getters that log a warning and return `undefined`. Am I reading that correctly?

I thinking make the old properties still usable (for now), but log a warning. More like:
```js
Object.defineProperty(cue.frame, 'id', {
  get: function() {
    videojs.log.warn('cue.frame.id is deprecated. Use cue.value.key instead.');
    return cue.value.key;
  }
});
```

gkatsev: Also, we could use `Object.defineProperties` here, though that shouldn't hold up this PR:

```js
Object.defineProperties(cue.frame, {
  'id': {
    get: () => cue.value.key
  },
  value: {
    get: () => cue.value.data
  },
  privateData: {
    get: () => cue.value.data
  }
});
```

saxena-gaurav:
Not needed. This will be in the videojs-contrib-media-sources project. 
https://github.com/videojs/videojs-contrib-media-sources

PR #397, opened 2015-09-28
Live variant syncing

dmlap:
When switching renditions or dealing with a live stream with unaligned variant playlists, we may discover that the segment we buffered isn't associated with the time range we expected it to be. In that case, adjust our information about timeline positioning and try buffering again.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -46,6 +46,8 @@ videojs.Hls = videojs.extend(Component, {
     this.tech_ = tech;
     this.source_ = options.source;
     this.mode_ = options.mode;
+    this.pendingSegment_ = null;
+
     this.bytesReceived = 0;
 
     // loadingState_ tracks how far along the buffering process we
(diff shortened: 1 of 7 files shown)


src/videojs-hls.js
```diff
@@ -320,10 +324,35 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
+      i = this.tech_.buffered().length;
+      while (i--) {
+        if (this.tech_.currentTime() < this.tech_.buffered().start(i)) {
+          // found the misidentified segment's buffered time range
+          // adjust the media index to fill the gap
+          var mi = this.mediaIndex;
+          currentBuffered = this.findCurrentBuffered_();
+          this.playlists.updateTimelineOffset(segmentInfo.mediaIndex, this.tech_.buffered().start(i));
+          this.mediaIndex = this.playlists.getMediaIndexForTime_(currentBuffered.end(0) + 1);
+          console.log(mi, '->', this.mediaIndex, 'expired:', this.tech_.buffered().start(i));
```
imbcmdth: Left behind debugging code?

test/stats/stats.css
```diff
@@ -4,10 +4,15 @@
...
-.intersect {
+.axis path {
+  fill: none;
+  stroke: #111;
+}
+
+.bitrates {
   fill: none;
-  stroke: #000;
+  stroke: steelblue;
```
imbcmdth: I think this should be red. :laughing:

src/videojs-hls.js
```diff
@@ -320,10 +324,35 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
+
+    // When switching renditions or seeking, we may misjudge the media
+    // index to request to continue playback. Check after each append
+    // that a gap hasn't appeared in the buffered region and adjust
+    // the media index to fill it if necessary
+    if (this.tech_.buffered().length === 2 &&
+        segmentInfo.playlist === this.playlists.media()) {
+      i = this.tech_.buffered().length;
+      while (i--) {
+        if (this.tech_.currentTime() < this.tech_.buffered().start(i)) {
```
imbcmdth: This doesn't handle the case where the mediaIndex is for a segment that is too early in the timeline, correct. I know that case will be handled "automatically" but it feels suboptimal and I don't think we ever correct the "TimelineOffset" in that case which might, in the long run, be a confusing behavior.

In other words, I think we should always attempt to correct the timeline offset because some other code may depend on that correction one day.

src/videojs-hls.js
dmlap: Yes :/

src/videojs-hls.js
dmlap: I was shooting for a low-impact change while we figure out something more comprehensive. I agree we should have a more robust solution but (for now) this does seem to work very reliably in my testing.

imbcmdth:
Despite my last commnet, LGTM

dmlap:
@imbcmdth filed https://github.com/videojs/videojs-contrib-hls/issues/399 to track the enhancements to this PR we'd like to get around to.

PR #398, opened 2015-09-28
make id3 cue structure similar to safari

saxena-gaurav:


src/videojs-hls.js
```diff
@@ -335,10 +360,12 @@ videojs.Hls.prototype.addCuesForMetadata_ = function(segmentInfo) {
     for (i = 0; i < metadata.frames.length; i++) {
       frame = metadata.frames[i];
       time = segmentOffset + ((metadata.pts - minPts) * 0.001);
-      cue = new window.VTTCue(time, time, frame.value || frame.url || '');
+      cue = new Cue(time, time, frame.value || frame.url || '');
```
imbcmdth: What about `frame.data`?

saxena-gaurav: In discussion with @imbcmdth , we do not need `frame.data`.

saxena-gaurav:
This is not needed any more. This work has been merged to development branch.

PR #400, opened 2015-09-28
Moved `loadstart` to the flash side of contrib media-sources

imbcmdth:
As a sourcehandler, we should no longer trigger loadstart in HLS and instead trigger it only if flash is being used.

Related to https://github.com/videojs/videojs-contrib-media-sources/pull/44

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -137,12 +137,6 @@ videojs.Hls.prototype.src = function(src) {
   // load the MediaSource into the player
   this.mediaSource.addEventListener('sourceopen', this.handleSourceOpen.bind(this));
 
-  // We need to trigger this asynchronously to give others the chance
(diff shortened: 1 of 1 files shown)


saxena-gaurav:
LGTM

dmlap:
I think the proposed migration to media-sources will cause `loadstart` to be emitted a bit too late in the loading lifecycle. `loadstart` should fire before the resource selection algorithm begins, which would roughly translate to tech/source selection time for video.js. Currently, we wait until the underlying tech fires the event and the SWF doesn't fire it at all in data generation mode. I think the correct solution would be to fire it when [src() is called in the player](https://github.com/videojs/video.js/blob/1d79c4a9e2b292e8dcfb983ccdea7cb648398692/src/js/player.js#L1710). As a temporary measure, we could fire it the moment HLS is selected as a source handler in Flash mode.

PR #404, opened 2015-09-29
Ensure Flash mode fires "loadstart"

dmlap:
For #400. Flash doesn't currently fire "loadstart" in data generation mode. Triggering that event should probably happen in video.js itself but fire it once the source handler is selected to make things behave somewhat sanely for now.


saxena-gaurav:
I think we don't need the below lines now 
https://github.com/dmlap/videojs-contrib-hls/blob/flash-mode-loadstart/src/videojs-hls.js#L147-L149

saxena-gaurav:
LGTM

PR #405, opened 2015-09-29
Remove old transmuxing infrastructure

dmlap:
All the transmuxing code has moved into mux.js so we can get rid of the remnants in this project.


imbcmdth:
LGTM

PR #411, opened 2015-10-08
Determine the segment to load by looking at buffered

dmlap:
When playlists are not segment-aligned or began at different times, we could make bad decisions about which segment to load by just incrementing the media index. Instead, annotate segments in the playlist with timeline information as they are downloaded. When a decision about what segment to fetch is required, simply try to fetch the segment that lines up with the latest edge of the buffered time range that contains the current time. Add a utility to stringify TextRanges for debugging. This is a checkpoint commit; 35 tests are currently failing in Chrome.

Opening this for early feedback &mdash; __NOT READY TO MERGE__

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -187,20 +187,20 @@
        * active media playlist. When called with a single argument,
        * triggers the playlist loader to asynchronously switch to the
        * specified media playlist. Calling this method while the
-       * loader is in the HAVE_NOTHING or HAVE_MASTER states causes an
-       * error to be emitted but otherwise has no effect.
+       * loader is in the HAVE_NOTHING causes an error to be emitted
+       * but otherwise has no effect.
        * @param playlist (optional) {object} the parsed media playlist
        * object to switch to
        */
       loader.media = function(playlist) {
-        var mediaChange = false;
+        var startingState = loader.state, mediaChange;
         // getter
         if (!playlist) {
           return loader.media_;
         }
 
         // setter
-        if (loader.state === 'HAVE_NOTHING' || loader.state === 'HAVE_MASTER') {
+        if (loader.state === 'HAVE_NOTHING') {
           throw new Error('Cannot switch media playlist from ' + loader.state);
         }
 
@@ -213,7 +213,7 @@
           playlist = loader.master.playlists[playlist];
         }
 
-        mediaChange = playlist.uri !== loader.media_.uri;
+        mediaChange = !loader.media_ || playlist.uri !== loader.media_.uri;
 
         // switch to fully loaded playlists immediately
         if (loader.master.playlists[playlist.uri].endList) {
@@ -258,7 +258,17 @@
           withCredentials: withCredentials
         }, function(error, request) {
           haveMetadata(error, request, playlist.uri);
-          loader.trigger('mediachange');
+
+          if (error) {
+            return;
+          }
+
+          // fire loadedmetadata the first time a media playlist is loaded
+          if (startingState === 'HAVE_MASTER') {
+            loader.trigger('loadedmetadata');
+          } else {
+            loader.trigger('mediachange');
+          }
         });
       };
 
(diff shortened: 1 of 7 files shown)


src/playlist-loader.js
```diff
@@ -424,17 +424,46 @@
...
+      segment = this.media_.segments[i];
+      if (segment.end !== undefined && segment.end <= time) {
+        time -= segment.end;
+        break;
+      }
+      if (segment.start !== undefined && segment.start < time) {
+
+        if (segment.end !== undefined && segment.end > time) {
+          // we've found the target segment exactly
+          return i;
```
imbcmdth: Nice!

src/playlist-loader.js
```diff
@@ -424,17 +424,46 @@
...
-                                false);
-
-      // HLS version 3 and lower round segment durations to the
-      // nearest decimal integer. When the correct media index is
-      // ambiguous, prefer the higher one.
-      if (time <= 0) {
-        return i;
+    // 1) Walk backward until we find the latest segment with timeline
+    // information that is earlier than `time`
+    targetDuration = this.media_.targetDuration || 10;
```
imbcmdth: Is the `||` here necessary? Target duration is a required field in HLS, if anything we should throw an error if it was never specified.

dmlap: My philosophy with all things playback is we should not punish the viewer for the sins of the publisher :) If it's possible to work around problems without strong-AI in our error handling, I think contrib-hls should do so.

src/playlist-loader.js
```diff
@@ -424,17 +424,46 @@
...
+        return j;
+      }
+
+      // 2a) If we discover a segment that has timeline information
+      // before finding the result segment, the playlist information
+      // must have been inaccurate. Start a binary search for the
+      // segment which contains `time`. If the guess turns out to be
+      // incorrect, we'll have more info to work with next time.
+      if (segment.start !== undefined || segment.end !== undefined) {
+        return Math.floor((j - i) * 0.5);
```
imbcmdth: I feel like we could be smarter than a binary search here but I am trying to find a way to explain it..

Basically we know the last point we have accurate information on (the `start_point`) Starting with the lemma that "every `segment.end` is the next segment's `start`" - we either know the `start` of `i--` (when we find a known start) or the `start` of `i` (when we find a known end). (It's `i--` or `i` because you have already incremented `i` after the loop.)

Likewise, we just found a segment we know about so now we know: either the `start` of segment j (the case where `segment.start` is not `undefined`) or the `start` of segment j++ (in the case where `segment.end` is not `undefined`).

With that information we know exactly the number of segments between two known points (either `i+j`, `i+j+1` or `i+j+2`) and we know exactly the range - from one known start point to the other known start point.

With that we can calculate `i + Math.floor(numSegments * (range / time))` to get a much better idea of which segment we are likely in. Again, the same convergence is as with the binary search will occur if we happen to be wrong but we are much more likely to be right the first time and, as a consequence, likely to converge on the correct answer more quickly.

src/playlist-loader.js
imbcmdth: Understand. I don't agree but I get the reasoning.. :hear_no_evil:

src/playlist-loader.js
dmlap: Agreed this is a better algorithm. Defer to another PR?

src/playlist.js
```diff
@@ -259,33 +136,14 @@
...
-                           liveBuffer);
-        liveBuffer -= pending;
-        end -= pending;
-      }
-    }
-
+    start = intervalDuration(playlist, playlist.mediaSequence);
+    end = intervalDuration(playlist,
+                           playlist.mediaSequence + playlist.segments.length);
+    end -= (playlist.targetDuration || DEFAULT_TARGET_DURATION) * 3;
```
imbcmdth: Safari handles live streams differently than the "spec" we should probably not use TARGET_DURATION and just base the calculation on "3 segments from the end" like Safari.

imbcmdth:
LGETM

PR #423, opened 2015-10-30
Some minor improvements to how we fetch segments given limited information

imbcmdth:
And some fixes for live stream handling (especially regarding the faking of the duration we must do!)

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -378,7 +378,6 @@
       knownStart,
       knownEnd;
 
-
     if (!this.media_) {
       return 0;
     }
@@ -461,7 +460,7 @@
       // We have a known-end point that is after our desired time so
       // walk from that point backwards
       time = knownEnd - time;
-      for (i = endIndex; i > 0; i--) {
+      for (i = endIndex; i >= 0; i--) {
         segment = this.media_.segments[i];
         time -= segment.duration || targetDuration;
         if (time < 0) {
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -370,6 +370,7 @@
     var
       i,
       segment,
+      originalTime = time,
       targetDuration = this.media_.targetDuration || 10,
       numSegments = this.media_.segments.length,
       lastSegment = numSegments - 1,
@@ -377,7 +378,7 @@
       endIndex,
       knownStart,
       knownEnd;
-
+console.log('>>', time);
     if (!this.media_) {
       return 0;
     }
@@ -396,9 +397,9 @@
         startIndex = i + 1;
         knownStart = segment.end;
         if (startIndex >= numSegments) {
-          // Somehow the last segment claims to end *before* the time we are
+          // The last segment claims to end *before* the time we are
           // searching for so just return it
-          return lastSegment;
+          return numSegments;
         }
         break;
       }
@@ -415,49 +416,51 @@
 
     // 2) Walk forward until we find the first segment with timeline
     // information that is greater than `time`
-    for (i = 0; i < this.media_.segments.length; i++) {
+    for (i = 0; i < numSegments; i++) {
       segment = this.media_.segments[i];
       if (segment.start !== undefined && segment.start > time) {
         endIndex = i - 1;
         knownEnd = segment.start;
         if (endIndex < 0) {
-          // Somehow the first segment claims to start *after* the time we are
+          // The first segment claims to start *after* the time we are
           // searching for so just return it
-          return 0;
+          return -1;
         }
         break;
       }
       if (segment.end !== undefined && segment.end > time) {
-        if (segment.start !== undefined && segment.start < time) {
-          // we've found the target segment exactly
-          return i;
-        }
         endIndex = i;
         knownEnd = segment.end;
         break;
       }
     }
 
-    if (startIndex !== undefined && endIndex !== undefined) {
-      // Do the fancy maths of "Algorithm Jon"
-      return startIndex + Math.floor(
-        ((time - knownStart) / (knownEnd - knownStart)) *
-        (endIndex - startIndex));
-    } else if (startIndex !== undefined) {
+    if (startIndex !== undefined) {
       // We have a known-start point that is before our desired time so
       // walk from that point forwards
       time = time - knownStart;
-      for (i = startIndex; i < numSegments; i++) {
+      for (i = startIndex; i < (endIndex || numSegments); i++) {
         segment = this.media_.segments[i];
         time -= segment.duration || targetDuration;
         if (time < 0) {
           return i;
         }
       }
-      // We haven't found a segment so load the last one
+
+      if (i === endIndex) {
+        // We haven't found a segment but we did hit a known end point
+        // so fallback to "Algorithm Jon" - try to interpolate the segment
+        // index based on the known span of the timeline we are dealing with
+        // and the number of segments inside that span
+        return startIndex + Math.floor(
+          ((originalTime - knownStart) / (knownEnd - knownStart)) *
+          (endIndex - startIndex));
+      }
+
+      // We _still_ haven't found a segment so load the last one
       return lastSegment;
     } else if (endIndex !== undefined) {
-      // We have a known-end point that is after our desired time so
+      // We _only_ have a known-end point that is after our desired time so
       // walk from that point backwards
       time = knownEnd - time;
       for (i = endIndex; i >= 0; i--) {
@@ -470,8 +473,9 @@
       // We haven't found a segment so load the first one
       return 0;
     } else {
-      // We known nothing so use "Algorithm A"
-      i = 0;
+      // We known nothing so use "Algorithm A" - walk from the front
+      // of the playlist naively subtracking durations until we find
+      // a segment that contains time and return it
       for (i = 0; i < numSegments; i++) {
         segment = this.media_.segments[i];
         time -= segment.duration || targetDuration;
@@ -479,7 +483,9 @@
           return i;
         }
       }
-      // We are out of possible candidates so load the last one
+      // We are out of possible candidates so load the last one...
+      // The last one is the least likely to overlap a buffer and therefore
+      // the one most likely to tell us something about the timeline
       return lastSegment;
     }
   };
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -378,7 +378,7 @@
       endIndex,
       knownStart,
       knownEnd;
-console.log('>>', time);
+
     if (!this.media_) {
       return 0;
     }
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -378,7 +378,6 @@
       knownStart,
       knownEnd;
 
-    time = Math.ceil(time * 100) / 100;
 
     if (!this.media_) {
       return 0;
@@ -392,11 +391,9 @@
 
     // 1) Walk backward until we find the first segment with timeline
     // information that is earlier than `time`
-    targetDuration
-
     for (i = lastSegment; i > 0; i--) {
       segment = this.media_.segments[i];
-      if (segment.end !== undefined && segment.end < time) {
+      if (segment.end !== undefined && segment.end <= time) {
         startIndex = i + 1;
         knownStart = segment.end;
         if (startIndex >= numSegments) {
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -391,7 +391,7 @@
 
     // 1) Walk backward until we find the first segment with timeline
     // information that is earlier than `time`
-    for (i = lastSegment; i > 0; i--) {
+    for (i = lastSegment; i >= 0; i--) {
       segment = this.media_.segments[i];
       if (segment.end !== undefined && segment.end <= time) {
         startIndex = i + 1;
@@ -403,7 +403,7 @@
         }
         break;
       }
-      if (segment.start !== undefined && segment.start < time) {
+      if (segment.start !== undefined && segment.start <= time) {
         if (segment.end !== undefined && segment.end > time) {
           // we've found the target segment exactly
           return i;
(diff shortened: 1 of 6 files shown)


src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
-    while (i--) {
+    for (i = lastSegment; i > 0; i--) {
       segment = this.media_.segments[i];
       if (segment.end !== undefined && segment.end <= time) {
-        time -= segment.end;
+        startIndex = i + 1;
+        knownStart = segment.end;
+        if (startIndex >= numSegments) {
+          // Somehow the last segment claims to end *before* the time we are
+          // searching for so just return it
```
dmlap: Statements like "somehow" make me nervous. Why would this condition occur?

src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
+    }
 
-        time -= segment.start;
-        time -= segment.duration || targetDuration;
-        if (time < 0) {
-          // the segment with start information is also our best guess
-          // for the momment
+    // 2) Walk forward until we find the first segment with timeline
+    // information that is greater than `time`
+    for (i = 0; i < this.media_.segments.length; i++) {
```
dmlap: I think you should use `numSegments` as your end condition here for consistency or get rid of the alias entirely. I kind of lean towards dropping `numSegments` since we already have a ton of function variables in scope already but it's your call.

src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
+        knownEnd = segment.start;
+        if (endIndex < 0) {
+          // Somehow the first segment claims to start *after* the time we are
+          // searching for so just return it
+          return 0;
+        }
+        break;
+      }
+      if (segment.end !== undefined && segment.end > time) {
+        if (segment.start !== undefined && segment.start < time) {
```
dmlap: Aren't we guaranteed this condition would be handled in our backward-walking loop?

src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
+        if (time < 0) {
+          return i;
+        }
       }
+      // We haven't found a segment so load the first one
+      return 0;
+    } else {
+      // We known nothing so use "Algorithm A"
+      i = 0;
+      for (i = 0; i < numSegments; i++) {
```
dmlap:
```js
i = 0;
i = 0;
```

src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
+        if (time < 0) {
+          return i;
+        }
+      }
 
-      if (time < 0) {
-        return j;
+      if (i === endIndex) {
+        // We haven't found a segment but we did hit a known end point
+        // so fallback to "Algorithm Jon" (ie. interpolation)
```
dmlap: "Algorithm Jon" is a funny name but I think we should be less entertaining, more informative in the comments.

src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
+        segment = this.media_.segments[i];
+        time -= segment.duration || targetDuration;
+        if (time < 0) {
+          return i;
+        }
       }
+      // We haven't found a segment so load the first one
+      return 0;
+    } else {
+      // We known nothing so use "Algorithm A"
```
dmlap: I'd put more detail on what "Algorithm A" is here.

src/playlist-loader.js
gkatsev: We should still be entertaining but also informative.

src/playlist-loader.js
```diff
@@ -379,57 +389,104 @@
...
+      i = 0;
+      for (i = 0; i < numSegments; i++) {
+        segment = this.media_.segments[i];
+        time -= segment.duration || targetDuration;
+        if (time < 0) {
+          return i;
+        }
+      }
+      // We are out of possible candidates so load the last one
+      return lastSegment;
```
dmlap: You've switched back to clamping return values to valid indices in the segment array. At the time I tried this, I felt like failure to find an index was indicative of a problem elsewhere in the code and clamping would obscure that. You think that worry is unfounded?

src/playlist-loader.js
imbcmdth: This is actually a common condition: We are looking for the next segment in a live playlist that hasn't updated yet so we are effectively searching for a time after the known end of the last segment.

src/playlist-loader.js
imbcmdth: I think `numSegments` is the way to go.. it's easier to read and I hate seeing repeated object traversal. That's a code smell to me.

src/playlist-loader.js
imbcmdth: Good point.

src/playlist.js
```diff
@@ -105,11 +105,12 @@
       if (!playlist.endList) {
         return window.Infinity;
       }
+      endSequence = playlist.segments.length;
```
dmlap: I have tried to consistently use the suffix "sequence" only for variables that store numbers in the media sequence number line. Looks like you're using it for an available segment index here.

src/videojs-hls.js
```diff
@@ -319,26 +332,58 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
-        segment.end = update.end;
+    if (segmentInfo.playlist.uri === this.playlists.media().uri) {
+      playlist = this.playlists.media();
+      currentMediaIndex = segmentInfo.mediaIndex + (segmentInfo.mediaSequence - playlist.mediaSequence);
+
+      // annotate the segment with any start and end time information
+      // added by the media processing
+      segment = playlist.segments[currentMediaIndex];
+      // spread the annotation to surrounding segments so that we are less likely
+      // to loose the information due to a live playlist update
```
dmlap: s/loose/lose/

src/playlist-loader.js
imbcmdth: Thanksgiving leftovers!

src/videojs-hls.js
```diff
@@ -319,26 +332,58 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
+        this.updateDuration(playlist);
+        // check if it's time to download the next segment
+        this.fillBuffer();
+        return;
       }
-    });
 
-    if (timelineUpdates.length) {
-      this.updateDuration(segmentInfo.playlist);
+    if (currentBuffered.length && currentBuffered.end(0) !== segmentInfo.currentBufferedEnd) {
```
dmlap: It's unclear to me why we would look for this condition instead of just checking to see that a timeline update was made. Maybe a comment would help?

src/videojs-hls.js
```diff
@@ -319,26 +332,58 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
-    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex];
-    timelineUpdates = videojs.Hls.bufferedAdditions_(segmentInfo.buffered,
-                                                     this.tech_.buffered());
-    timelineUpdates.forEach(function(update) {
-      if (update.start !== undefined) {
-        segment.start = update.start;
-      }
-      if (update.end !== undefined) {
-        segment.end = update.end;
+    if (segmentInfo.playlist.uri === this.playlists.media().uri) {
```
dmlap: This conditional is pretty hefty. What about inverting it and short-circuiting?

src/playlist-loader.js
imbcmdth: In the live case, we can't rely on knowing anything about the playlist. If we have failed to find an index, we need more information. Clamping tries to pin down the start/end of the playlist (what I think of as 'anchoring'.)

Even if anchoring fails to give us new information we will fall back to walking forward until we have all the segments or finally learned something to help us pick the right segment.

src/videojs-hls.js
imbcmdth: Fancy book lernin'!

src/videojs-hls.js
```diff
@@ -465,15 +507,29 @@ videojs.Hls.prototype.updateDuration = function(playlist) {
...
+      oldDuration, seekable.end(0) + playlist.targetDuration * 3);
+  }
 
   // if the duration has changed, invalidate the cached value
   if (oldDuration !== newDuration) {
-    if (this.mediaSource.readyState === 'open') {
+    if (this.mediaSource.readyState !== 'open') {
+      this.mediaSource.addEventListener('sourceopen', setDuration);
+    } else if (this.sourceBuffer && this.sourceBuffer.updating) {
+      this.sourceBuffer.addEventListener('updateend', setDuration);
```
dmlap: There may be multiple `updateend` listeners on a source buffer and one of the previous ones could perform an operation that causes it to enter the updating state.

src/videojs-hls.js
```diff
@@ -465,15 +507,29 @@ videojs.Hls.prototype.updateDuration = function(playlist) {
...
+        }
+      }.bind(this),
+      seekable = this.seekable();
+
+  if (seekable.length && newDuration === Infinity) {
+    if (isNaN(oldDuration)) {
+      oldDuration = 0;
+    }
+    newDuration = Math.max(
+      oldDuration, seekable.end(0) + playlist.targetDuration * 3);
```
dmlap: Huh?

src/videojs-hls.js
```diff
@@ -676,8 +734,8 @@ videojs.Hls.prototype.findCurrentBuffered_ = function() {
   if (buffered && buffered.length) {
     // Search for a range containing the play-head
     for (i = 0; i < buffered.length; i++) {
-      if (buffered.start(i) <= currentTime &&
-          buffered.end(i) >= currentTime) {
+      if (buffered.start(i) <= (currentTime + 0.05) &&
+          buffered.end(i) >= (currentTime - 0.05)) {
```
dmlap: We discussed normalizing this elsewhere. I still think that's the right move.

src/videojs-hls.js
```diff
@@ -739,28 +797,37 @@ videojs.Hls.prototype.fillBuffer = function(seekToTime) {
...
+      }
+    } else {
+      mediaIndex = this.playlists.getMediaIndexForTime_(this.tech_.currentTime());
+    }
   }
   segment = this.playlists.media().segments[mediaIndex];
 
+  if (segment &&
+      this.lastSegmentUri_ &&
+      this.lastSegmentUri_.indexOf(segment.uri) !== -1) {
```
dmlap: If getMediaIndexForTime() didn't clamp, the instance variable wouldn't be necessary.

src/videojs-hls.js
```diff
@@ -739,28 +797,37 @@ videojs.Hls.prototype.fillBuffer = function(seekToTime) {
...
     return;
   }
 
-  // if there is plenty of content in the buffer and we're not
-  // seeking, relax for awhile
-  if (typeof seekToTime !== 'number' &&
-      bufferedTime >= videojs.Hls.GOAL_BUFFER_LENGTH) {
-    return;
-  }
+  this.lastMediaIndex_ = mediaIndex;
```
dmlap: Unused, remove.

src/videojs-hls.js
```diff
@@ -944,41 +1013,32 @@ videojs.Hls.prototype.drainBuffer = function(event) {
...
       // the start of the buffered region
       this.sourceBuffer.timestampOffset = segmentTimestampOffset;
     }
-  } else if (segment.discontinuity) {
+  } else if (segment.discontinuity && currentBuffered.length) {
     // If we aren't seeking and are crossing a discontinuity, we should set
     // timestampOffset for new segments to be appended the end of the current
     // buffered time-range
     this.sourceBuffer.timestampOffset = currentBuffered.end(0);
+  } else if (!hasBufferedContent && this.tech_.currentTime() > 0.05) {
```
dmlap: This is to account for videos that don't appear to start from zero? I think we need to get to the bottom of that issue in the transmuxer, not here.

src/playlist-loader.js
dmlap: Ok, that makes more sense. Can you replace "somehow" in the comment with that example?

src/playlist-loader.js
dmlap: I'm not sure I totally follow your explanation. Assuming there is a need to signal something about why getMediaIndexForTime() did not find a segment, wouldn't -1 and segments.length indicate in what direction the failure occurred as well as let the caller know that an actual match wasn't found?

src/videojs-hls.js
imbcmdth: Removed

src/videojs-hls.js
imbcmdth: No this is to cover the strange condition mentioned in the comment below. We are not seeking, current time is not zero and we don't have any buffered data. In that case we need to set `timestampOffset` before placing the segment. I saw this several times while trying to get live streams to work.

src/videojs-hls.js
imbcmdth: I'm not sure we ever came up with a good solution for that one yet. This also handles the case where our start isn't quite zero. This is an interim solution.

src/videojs-hls.js
imbcmdth: Wha?

src/videojs-hls.js
imbcmdth: Done.

src/videojs-hls.js
imbcmdth: Done.

src/playlist.js
imbcmdth: In this case, it is being used as a "sequence" for the purposes of the calculation later. It seems silly to introduce another variable and/or a condition when a simple assignment will suffice.

dmlap: I only see it being assigned to here and the comment at the top of this function indicates `playlist.mediaSequence` is already a component. Seems like a bug waiting to happen.

src/videojs-hls.js
dmlap: Can we use `currentTime !== 0` to capture that case? The 0.05 is the part of this condition that has me hesitating.

imbcmdth: Done.

dmlap:
Merged via https://github.com/videojs/videojs-contrib-hls/commit/bce88a1883f7af4573b553c11538afb180f42efd

PR #428, opened 2015-11-05
Resolved endOfStream not being called reliably

imbcmdth:


src/videojs-hls.js
```diff
@@ -307,40 +307,30 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
       segmentInfo = this.pendingSegment_,
       segment,
+      segments,
       playlist,
       currentMediaIndex,
       currentBuffered,
       timelineUpdates;
 
-    // stop here if the update errored or was aborted
-    if (!segmentInfo) {
```
imbcmdth: This was being called twice. The correct place to call it is after `pendingSegment_` is reset.

src/videojs-hls.js
```diff
@@ -307,40 +307,30 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
-      return;
-    }
-
     this.pendingSegment_ = null;
 
-    // if we've buffered to the end of the video, let the MediaSource know
-    currentBuffered = this.findCurrentBuffered_();
-    if (currentBuffered.length &&
```
imbcmdth: Changed and moved to below segment annotation.

src/videojs-hls.js
```diff
@@ -350,15 +340,21 @@ videojs.Hls.prototype.setupSourceBuffer_ = function() {
...
         }
       }
     });
 
+    // if we've buffered to the end of the video, let the MediaSource know
+    if (this.playlists.media().endList &&
+        currentBuffered.length &&
+        segments[segments.length - 1].end <= currentBuffered.end(0) &&
+        this.mediaSource.readyState === 'open') {
+      this.mediaSource.endOfStream();
```
imbcmdth: Basically we say: If we aren't in a live playlist and the last segment has a duration equal to the end point of the currently buffered region, we consider the video to have ended.

src/videojs-hls.js
```diff
@@ -450,6 +446,8 @@ videojs.Hls.prototype.setCurrentTime = function(currentTime) {
     return currentTime;
   }
 
+  this.lastSegmentLoaded_ = null;
```
dmlap: What is this?

imbcmdth: When we are seeking, we shouldn't care if we load the same segment as last time - this was mostly a problem in flash because we don't have a buffer on seek so we potentially need to load the exact same segment again in this one case.

dmlap: I still think we can get rid of this additional instance variable but that explanation makes sense if you allow that it should exist in the first place :)

imbcmdth: I agree that I'd like to have less "global" state but for now it's required..

PR #429, opened 2015-11-06
fix the path to media sources files

saxena-gaurav:


dmlap:
This has been tackled through a different PR.

PR #432, opened 2015-11-10
Don't append segments that have already been appended

### LINKED ISSUE #430: drainBuffer can be called after updating is set to false
`updateend` is fired asynchronously, so it's possible drainBuffer() will be called after `sourceBuffer.updating` is false but before `this.pendingSegment_` is properly cleared out. This causes the ArrayBuffer to be transferred to the transmuxer twice, which triggers an exception because it is neutered.

dmlap:
Fixes #430. If drainBuffer is called after updating is reset on the sourceBuffer but before updateend fires, a segment could be appended twice. Check to make sure a segment hasn't already been appended before proceeding with draining the buffer.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -23,7 +23,8 @@ keyFailed = function(key) {
   return key.retries && key.retries >= 2;
 };
 
-videojs.Hls = videojs.extend(Component, {
+videojs.Hls = {};
+videojs.HlsHandler = videojs.extend(Component, {
   constructor: function(tech, options) {
     var self = this, _player;
 
@@ -110,7 +111,7 @@ videojs.HlsSourceHandler = function(mode) {
           tech.trigger('loadstart');
         }, 1);
       }
-      tech.hls = new videojs.Hls(tech, {
+      tech.hls = new videojs.HlsHandler(tech, {
         source: source,
         mode: mode
       });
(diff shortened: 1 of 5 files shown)


src/videojs-hls.js
```diff
@@ -23,7 +23,7 @@ keyFailed = function(key) {
   return key.retries && key.retries >= 2;
 };
 
-videojs.Hls = videojs.extend(Component, {
+videojs.HlsHandler = videojs.extend(Component, {
```
dmlap: This is the actual rename.

saxena-gaurav:
@dmlap : Changes looks good.

src/videojs-hls.js
```diff
@@ -486,31 +487,32 @@ videojs.Hls.prototype.seekable = function() {
...
-videojs.Hls.prototype.updateDuration = function(playlist) {
+videojs.HlsHandler.prototype.updateDuration = function(playlist) {
   var oldDuration = this.mediaSource.duration,
       newDuration = videojs.Hls.Playlist.duration(playlist),
       setDuration = function() {
         this.mediaSource.duration = newDuration;
         this.tech_.trigger('durationchange');
+
+        // update seekable
+        if (seekable.length !== 0 && newDuration === Infinity) {
```
imbcmdth: Below we trigger `durationchange` after setting seekable; here we do it before. Since there is no `seekablechange` event, we probably want to trigger `durationchange` afterwards here as well for now.

PR #434, opened 2015-11-12
fixes #433

### LINKED ISSUE #433: videojs.Hls.isSupported not returning boolean
The function "videojs.Hls.isSupported" appears to be buggy:

```javascript
videojs.Hls.isSupported = function() {

  return !videojs.Hls.supportsNativeHls &&
    videojs.Flash.isSupported() &&
    videojs.MediaSource &&
    window.Uint8Array;
};
```

would not return a boolean. - it returns the function "window.Uint8Array" instead.

IndyIndyIndy:


dmlap:
@IndyIndyIndy looks like the Travis failure was using `==` instead of `===`. Mind updating the PR so we can pull it in?

forbesjo:
We won't be making anymore changes in master at this time unless they're critical. Could you please open a new PR against development if the same issue exists there?

PR #438, opened 2015-11-17
Blacklist incompatible codecs

dmlap:


src/videojs-hls.js
```diff
@@ -585,7 +621,8 @@ videojs.HlsHandler.prototype.selectPlaylist = function () {
     effectiveBitrate,
     sortedPlaylists = this.playlists.master.playlists.slice(),
     bandwidthPlaylists = [],
-    i = sortedPlaylists.length,
+    now = +new Date(),
```
gkatsev: `Date.now()` is supported in all browsers we care to support. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now#Browser_compatibility
Not a blocker.

dmlap: Oops, forgot to address this one in my update. We should probably do a sweep to fix this all over the place.

gkatsev: Not a big issue.

imbcmdth:
One minor comment. Otherwise LGTM

PR #439, opened 2015-11-18
Minor fixes to fetcher

imbcmdth:
* Increased the `bandwidthVariance` based on similar constraints in other projects
* Added a fudge factor of half a frame to account for TimeRanges rounding
* Don't try to cancel a fetch in the middle of an append operation

* Live point is now 3 segments from the end instead of 3 target durations

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -141,8 +141,7 @@
     // https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-6.3.3
     start = intervalDuration(playlist, playlist.mediaSequence);
     end = intervalDuration(playlist,
-                           playlist.mediaSequence + playlist.segments.length);
-    end -= (playlist.targetDuration || DEFAULT_TARGET_DURATION) * 3;
+                           playlist.mediaSequence + Math.max(0, playlist.segments.length - 3));
     end = Math.max(0, end);
     return videojs.createTimeRange(start, end);
   };
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -142,7 +142,6 @@
     start = intervalDuration(playlist, playlist.mediaSequence);
     end = intervalDuration(playlist,
                            playlist.mediaSequence + Math.max(0, playlist.segments.length - 3));
-    end = Math.max(0, end);
     return videojs.createTimeRange(start, end);
   };
 
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -397,73 +397,7 @@ videojs.HlsHandler.prototype.setupSourceBuffer_ = function() {
 
   // transition the sourcebuffer to the ended state if we've hit the end of
   // the playlist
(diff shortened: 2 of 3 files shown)


src/playlist.js
```diff
@@ -141,8 +141,7 @@
     // https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-6.3.3
     start = intervalDuration(playlist, playlist.mediaSequence);
     end = intervalDuration(playlist,
-                           playlist.mediaSequence + playlist.segments.length);
-    end -= (playlist.targetDuration || DEFAULT_TARGET_DURATION) * 3;
+                           playlist.mediaSequence + Math.max(0, playlist.segments.length - 3));
     end = Math.max(0, end);
```
dmlap: Can `end` be less than zero after the previous line?

src/videojs-hls.js
```diff
@@ -396,7 +397,7 @@ videojs.HlsHandler.prototype.setupSourceBuffer_ = function() {
 
   // transition the sourcebuffer to the ended state if we've hit the end of
   // the playlist
-  this.sourceBuffer.addEventListener('updateend', function() {
+  this.sourceBuffer.addEventListener('updateend', function updateEndHandler() {
     var
```
dmlap: Should we make this a private function instead of keeping it inline?

src/videojs-hls.js
```diff
@@ -540,6 +542,11 @@ videojs.HlsHandler.prototype.setCurrentTime = function(currentTime) {
     return currentTime;
   }
 
+  // if we are in the middle of appending a segment, let it finish up
+  if (this.pendingSegment_ && this.pendingSegment_.buffered) {
+    return currentTime;
```
dmlap: Wouldn't this become a no-op in `fillBuffer()` anyways?

imbcmdth: The issue is that the following code cancels the segment XHR and  clears the `this.pendingSegment_` property. When 'updateend' is called we have no way of annotating segments. This was resulting in us getting the same segment (the first) twice during a live playlist start.

The steps went something like:
* We would start the request for the right segment via checkBuffer_ (based on currentTime)
* The seek handler would try to cancel it but it was already in flight
* We would finish appending the segment but store no information about it in `updateend`
* Then we would fetch the same segment (based on currentTime)
* In that second segment's `updateend`, we would record *no* new information because the buffer already has the same segment!

It was a strange and difficult to track down situation.

src/videojs-hls.js
imbcmdth: I would support this.. It's messy as-is.

src/playlist.js
imbcmdth: No. Fixing.

src/videojs-hls.js
dmlap: Yuck. We need to clean up this workflow but I'm ok with this as a temporary solution.

PR #440, opened 2015-11-19
Blacklist

imbcmdth:
First shot at temporarily blacklisting problematic playlists. Will flag a playlist as bad and then select the next appropriate playlist. If no playlists remain, it will signal the error that caused the problem to begin with.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -180,7 +180,7 @@ videojs.HlsHandler.prototype.src = function(src) {
   }.bind(this));
 
   this.playlists.on('error', function() {
-    this.blacklistCurrentPlaylist_(this.playlists.error.code);
+    this.blacklistCurrentPlaylist_(this.playlists.error);
   }.bind(this));
 
   this.playlists.on('loadedplaylist', function() {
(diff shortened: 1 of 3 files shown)


src/videojs-hls.js
```diff
@@ -179,15 +180,7 @@ videojs.HlsHandler.prototype.src = function(src) {
...
-    // close the media source with the appropriate error type
-    if (this.playlists.error.code === 2) {
-      this.mediaSource.endOfStream('network');
-    } else if (this.playlists.error.code === 4) {
-      this.mediaSource.endOfStream('decode');
-    }
-
-    // if this error is unrecognized, pass it along to the tech
-    this.tech_.error(this.playlists.error);
+    this.blacklistCurrentPlaylist_(this.playlists.error.code);
```
dmlap: Should we be passing the entire error object here?

src/videojs-hls.js
```diff
@@ -916,6 +909,34 @@ videojs.HlsHandler.prototype.setBandwidth = function(xhr) {
   this.tech_.trigger('bandwidthupdate');
 };
 
+videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
+  var currentPlaylist, nextPlaylist;
+
+  currentPlaylist = this.playlists.media();
+
+  if (!currentPlaylist) {
```
dmlap: When would this occur? (Maybe a comment would help here)

src/videojs-hls.js
```diff
@@ -930,7 +951,11 @@ videojs.HlsHandler.prototype.loadSegment = function(segmentInfo) {
   this.segmentXhr_ = videojs.Hls.xhr({
     uri: segmentInfo.uri,
     responseType: 'arraybuffer',
-    withCredentials: this.source_.withCredentials
+    withCredentials: this.source_.withCredentials,
+    // Set xhr timeout to 150% of the segment duration to allow us
+    // some time to switch renditions in the event of a catastrophic
+    // decrease in network performance or a server issue.
+    timeout: (segment.duration * 1.5) * 1000
```
dmlap: Nice. We shouldn't assume `segment.duration` is always defined, though-- how about a fallback to `targetDuration`?

test/videojs-hls_test.js
```diff
@@ -1510,32 +1495,23 @@ test('playlist 404 should end stream with a network error', function() {
...
-
   openMediaSource(player);
 
-  standardXHRResponse(requests[0]);
-  requests[1].respond(404);
-  ok(player.tech_.hls.error.message, 'an error message is available');
-  equal(2, player.tech_.hls.error.code, 'Player error code should be set to MediaError.MEDIA_ERR_NETWORK');
-});
-
-test('segment 500 should trigger MEDIA_ERR_ABORTED', function () {
```
dmlap: You kind of killed a test for 500 response codes here. Intentional?

dmlap:
Looks good other than the couple questions/comments I put in there. It's awesome that adding support for redundant streams turned out to be a relatively tiny change!

test/videojs-hls_test.js
imbcmdth: All non-200 responses are handled by exactly the same code. I didn't see the point of testing each individual code! :D

imbcmdth:
> It's awesome that adding support for redundant streams turned out to be a relatively tiny change! 

Even WITH blacklisting timeouts!  :wink:

PR #444, opened 2015-11-20
Fix two bugs in the blacklisting facility

imbcmdth:
* Moved blacklisting to above the rendition selection line so that we don't try to select the rendition that was just blacklisted
* Playlist Loader now emits the playlist it was trying to load so that it can be blacklisted when initial load was unsuccessful


src/videojs-hls.js
```diff
@@ -909,10 +909,18 @@ videojs.HlsHandler.prototype.setBandwidth = function(xhr) {
...
+ * Blacklists a playlist when an error occurs for a set amount of time
+ * making it unavailable for selection by the rendition selection algorithm
+ * and then forces a new playlist (rendition) selection.
+ */
 videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
   var currentPlaylist, nextPlaylist;
 
-  currentPlaylist = this.playlists.media();
+  // If the `error` was generated by the playlist loader, it may contain a
+  // the playlist we were trying to load (but failed) and we will blacklist
```
dmlap: > a ... the

dmlap:
Seems like we must have a gap in our tests if this snuck by. Otherwise LGTM

PR #448, opened 2015-11-23
fix skipping over discontinuity

saxena-gaurav:
If the timestamp offset is greater than the end of the buffered region or if the timestamp offset is less than the start of the buffered region, then set the timestamp offset to the end of the current buffered time range.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -1107,6 +1107,13 @@ videojs.HlsHandler.prototype.drainBuffer = function(event) {
       // the start of the buffered region
       this.sourceBuffer.timestampOffset = segmentTimestampOffset;
     }
+  } else if (currentBuffered && currentBuffered.length &&
+    (currentBuffered.end(0) < this.sourceBuffer.timestampOffset ||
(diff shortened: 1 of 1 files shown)


src/videojs-hls.js
```diff
@@ -1107,6 +1107,13 @@ videojs.HlsHandler.prototype.drainBuffer = function(event) {
       // the start of the buffered region
       this.sourceBuffer.timestampOffset = segmentTimestampOffset;
     }
+  } else if (currentBuffered && currentBuffered.length &&
+    (currentBuffered.end(0) < this.sourceBuffer.timestampOffset ||
+    currentBuffered.start(0) > this.sourceBuffer.timestampOffset)) {
+    // If the timestamp offset is greater than the end of the buffered region
+    // or if the timestamp offset is less than the start of the buffered region
+    // set the timestamp offset to the end of the current buffered time range.
```
dmlap: Yep, this is definitely confusing :)

dmlap:
I believe we've found (and fixed) other underlying issues that were causing this problem. Please re-open if I'm mistaken.

Issue #342, opened 2015-07-08
Support EXT-X-BYTERANGE


jpfranco:
Hello, I'm trying figure out whether there is something wrong with our streams or with videojs-contrib-hls itself. I'm testing a few videos that were processed with Amazon's Elastic Transcoder (HLS protocol version 4) and I'm experiencing some strange behavior. 

On Chrome the videos take considerably longer to start playing than on an HLS capable browser, and when playback begins, the duration displayed on the player changes to 2x the actual duration of the video. Some times when you seek forward, the duration will increase again, and then it will keep on increasing over time.

This doesn't seem to happen with other HLS streams that I've found on the web (including the one on your test page), but the streams I'm testing play without issues on Safari and Android Chrome using vanilla VideoJS. CORS is also properly set up.

I don't have the full setup online, but I do have two sample streams with this issue:
[stream 1](https://authproxy-1.ocdevel.net/s3-m3u8/organization-jpenterprises/video-0gRO9PQqL/video-hls-low.m3u8?Signature=h8acTufNizcaa4LwTQftKJfYsf4lZwNINHKUz~dcPirKyCr9kEMg33eG0dU5AMntoQP5O5k3Rs4KZrWvcl9iFEhKrnO80D4FD98Z6qBChmt3fNEB1-xff5gRu9ZE6QKA6unq1Pszl~m-~xLSlG4-7EsGgrSRrtp2swD93b5SOG14VeWSWUM9wK5mgkml4tE51mFZhoQ8f9OqTC2bLnJHIC1UjcZ9t0NyBaZRy7vesk03bHBopG1LAqErsHFJqfRa2yTAOX1LF7NWpbFPSrKoQqnzhS~C~BlsOQWCPnO6dUlG815~DlsVUBEUzx3LnZ0chKbUzI9mON3MxAZAVyiptg__&Key-Pair-Id=APKAJ5JH3L2ZNI5QDHEQ&Expires=2147483647)
[...]

PR #451, opened 2015-11-24
Feature/byterange support

KatrinaE:
* This pull request adds support for the #EXT-X-BYTERANGE tag to support HLS playlists with byteranges, like the one in Listing 15 of Apple's [Example Playlist Files](https://developer.apple.com/library/ios/technotes/tn2288/_index.html#//apple_ref/doc/uid/DTS40012238-CH1-BYTE_RANGE_SUPPORT_FOR_SEGMENTS). It does not break compatibility with old-style playlists, like the one in Listing 14. This would fix [#342](https://github.com/videojs/videojs-contrib-hls/issues/342).

* The byterange is added to the request for each segment as an HTTP Range header. Servers serving videos for playlists with #EXT-X-BYTERANGE tags need to parse request headers like this one:

        Range:bytes=0-288955
   
    The server should respond with status code 206 (partial content) and the appropriate segment of the video. The response should also contain the `Content-Range` header, e.g:
    
        Content-Range:bytes 0-288955/2521832

* If the video is on a different domain from the player, the server must give the player explicit permission to use the `Range` header by including the `Access-Control-Allow-Headers` header in its response to the CORS preflight request:

        Access-Control-Allow-Headers:Range

    The preflight request uses the HTTP 'OPTIONS' verb and is sent automatically by the browser. A CORS error will be thrown if the server does not respond with the proper header.

* Tested in Chrome, Firefox, Safari, Edge 12, and IE 11 using a player based on the [example player](http://videojs.github.io/videojs-contrib-hls/). Does not work with IE 10, but neither did the existing videojs.hls.js.

* Tested against version 4.12.6 of video.js. Does not work with the latest version (5.2), but neither did the existing videojs.hls.js.

* No tests were added to the test suite. The fix depends on a change I made to `videojs.Hls.xhr` to support sending arbitrary headers, and it didn't look to me like `videojs.Hls.xhr` was tested anywhere. I would be happy to add test(s) if someone can point me in the right direction.

* Known issue: Chrome issues an OPTIONS request for every video segment if CORS is required. As far as I know, it should cache the OPTIONS response for a while, so this shouldn't be the case. Firefox does not have this problem - it only issues an OPTIONS request for the first segment.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -693,8 +693,7 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
     player = this.player(),
     buffered = player.buffered(),
     bufferedTime = 0,
-    segment,
-    segmentUri;
+    segment;
 
   // if preload is set to "none", do not download segments until playback is requested
   if (!player.hasClass('vjs-has-started') &&
@@ -750,10 +749,7 @@ videojs.Hls.prototype.fillBuffer = function(offset) {
     return;
   }
 
-  // resolve the segment URL relative to the playlist
-  segmentUri = this.playlistUriToUrl(segment.uri);
-
-  this.loadSegment(segmentUri, offset);
+  this.loadSegment(segment, offset);
 };
 
 videojs.Hls.prototype.playlistUriToUrl = function(segmentRelativeUrl) {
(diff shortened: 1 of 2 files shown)


gkatsev:
That's awesome. Thanks.
We're currently trying to get MSE support with HLS happen in the development branch of this repo. That branch also supports videojs 5.x. Would be awesome if you could update this PR against that branch instead.

KatrinaE:
Sure, I can take a look.

KatrinaE:
Is there a build of videojs-contrib-hls that works with 5.x?

gkatsev:
The 1.x tags (or @next on npm) are vjs 5.x compatible. It's what is in the development branch.

KatrinaE:
Ahh, sorry. Just reread your comment. My apologies.

gkatsev:
No worries.

dmlap:
Closed via 7080865c5dbba6c01b46d649c7107d41326882b6

PR #454, opened 2015-11-26
Support #EXT-X-BYTERANGE in video-contrib-hls v1.x

### LINKED ISSUE #342: Support EXT-X-BYTERANGE

KatrinaE:
This patch fixes #342. It offers the same functionality as #451, but is intended for the video-contrib-hls development branch (v1.x) instead of the master branch (v0.x).

Notes in addition to those in #451:

* I had to change `lastSegmentLoaded_` from the segment URI to the entire segment. Otherwise, the player sees all segments after the first one as duplicates (because they all have the same URI).

* I added support for status code 206 (partial content) to videojs.Hls.xhr, even though the current code explicitly limits response codes to 200 and 0. Responses that satisfy part of a `Range` request may respond with 206 instead of 200.

Also, I ran into a couple of possible bugs in video-contrib-hls. Both of these were present before I made my changes.

* The segments for small videos load in an infinite loop until you click the play button. I noticed this problem in several videos < 3 MB, but not in any larger videos. I didn't spend a great deal of time looking into it; I can do more research if the problem doesn't sound familiar.

* IE 11 is behaving strangely: the time remaining counts down very quickly at first, so only a portion of the video is played before time runs out. You can see this in the sample video included in example.html.


imbcmdth:
Amazing - thanks for doing the extra legwork to add this feature to the development branch!

RE:

1) Segment fetching in infinite loop - hopefully this behavior has been corrected or at least improved by #463
2) IE11 this is definitely strange! @forbesjo Can you reproduce this?

forbesjo:
Looks like IE11 Win 8.1 and IE11 Win 10 have different behaviors? IE11 on Win 10 skips the first 10 seconds. @KatrinaE did you see the issue on Win 10?

IE11/Win8.1
ie11-win8_1

IE11/Win10
ie11-win10

KatrinaE:
@forbesjo yes, this was just on Win 10 - I did not test on Win 8.

src/videojs-hls.js
```diff
@@ -854,7 +854,7 @@ videojs.HlsHandler.prototype.fillBuffer = function(mediaIndex) {
   // we have entered a state where we are fetching the same segment,
   // try to walk forward
   if (this.lastSegmentLoaded_ &&
-      this.lastSegmentLoaded_ === this.playlistUriToUrl(segment.uri)) {
+      this.lastSegmentLoaded_ === segment) {
```
dmlap: When we have to reload the playlist because the video is live, I believe we recreate the segment objects and merge the properties together, which would cause this check to fail.

src/videojs-hls.js
```diff
@@ -895,6 +895,29 @@ videojs.HlsHandler.prototype.playlistUriToUrl = function(segmentRelativeUrl) {
   return playListUrl;
 };
 
+/*  Turns segment byterange into a string suitable for use in
+ *  HTTP Range requests
+ */
+videojs.HlsHandler.prototype.byterangeStr = function(byterange) {
```
dmlap: If you like this function on the HlsHandler prototype, I think it should be marked "private". We normally indicate a function is internal-only by appending an underscore.

src/videojs-hls.js
```diff
@@ -895,6 +895,29 @@ videojs.HlsHandler.prototype.playlistUriToUrl = function(segmentRelativeUrl) {
...
+    // `byterangeEnd` is one less than `offset + length` because the HTTP range
+    // header uses inclusive ranges
+    byterangeEnd = byterange.offset + byterange.length - 1;
+    byterangeStart = byterange.offset;
+    return "bytes=" + byterangeStart + "-" + byterangeEnd;
+};
+
+/*  Defines headers for use in the xhr request for a particular segment.
+ */
+videojs.HlsHandler.prototype.segmentXhrHeaders = function(segment) {
```
dmlap: Same comment about this function being marked internal.

imbcmdth:
Thanks again for the feature! We are prepping this for inclusion and have the PR in #515.

PR #459, opened 2015-11-27
Long term seekable start

dmlap:
When we switch playlists in a live video, we have to find the right place in the new playlist to continue buffering. This is complicated because we can't guarantee the two variants are segmented at the same time positions or that the windows of time they represent are exactly in sync. Most of the time, they're pretty close to one another and we can use that fact to make better guesses at which segment to download when switching.

This PR adds back tracking of expired content in the playlist loader, which can then be used to estimate the seekable window for live playlists even before we've buffered any segments from them. This also allows seekable to be accurate even when the player has paused for a long time and all the segment timing information we gathered has gone out of date. To make rejoining or seeking in a live stream even more robust, we detect when a seek "misses" the live window and seek again to a safe position.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -1206,7 +1206,7 @@ videojs.HlsHandler.prototype.updateEndHandler_ = function () {
       var next = this.findNextBufferedRange_();
       if (next.length) {
         videojs.log('tried seeking to', this.tech_.currentTime(), 'but that was too early, retrying at', next.start(0));
-        this.tech_.setCurrentTime(next.start(0));
+        this.tech_.setCurrentTime(next.start(0) + TIME_FUDGE_FACTOR);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


src/videojs-hls.js
```diff
@@ -1156,9 +1194,25 @@ videojs.HlsHandler.prototype.updateEndHandler_ = function () {
...
+  // in that case, seek again
+  seekable = this.seekable();
+  if (this.tech_.seeking() &&
+      currentBuffered.length === 0) {
+    if (seekable.length &&
+        this.tech_.currentTime() < seekable.start(0)) {
+      var next = this.findNextBufferedRange_();
+      if (next.length) {
+        videojs.log('tried seeking to', this.tech_.currentTime(), 'but that was too early, retrying at', next.start(0));
+        this.tech_.setCurrentTime(next.start(0) + TIME_FUDGE_FACTOR);
```
dmlap: This fudge factor shouldn't be necessary but it looks like getMediaIndexForTime is rounding improperly right now. I'm looking into the fix but this makes rejoining live streams reliable for the moment.

PR #463, opened 2015-12-01
Simplify stuff

### LINKED ISSUE #458: Infinite looping after multiple seek
Plugin version 1.2.1
After doing multiple seeks player goes in infinite loop and tries to download same chunks again and again.

imbcmdth:
This fixes several bad behaviors we can fall into when seeking multiple times by simplifying things and taking a YAGNI approach to debugging.

Playlist-loader's `getmediaIndexForTime_` no longer uses segment start times and has been drastically simplified.

Hls's `bufferedAdditions_` has been changed completely and renamed `findSoleUncommonTimeRangesEnd_` - the only thing that wasn't simplified was the name! It now attempts to be sure 100% that it is returning a proper end-point for the segment or it returns nothing. Should fix many issues that were caused by bad information being saved to the playlist that was throwing off other calculations.

Aborted XHRs were still eventually triggering callbacks due to timeouts. The callback didn't check to make sure it was being executed for the current in-flight request before blowing away the state used to track requests.

Fixes #458

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -450,7 +450,7 @@
           break;
         } else {
           knownStart = segment.end;
-          startIndex = i;
+          startIndex = i + 1;
         }
       }
     }
@@ -494,8 +494,13 @@
           return i;
         }
       }
-      // We haven't found a segment so load the first one
-      return 0;
+
+      // We haven't found a segment so load the first one if time is zero
+      if (time === 0) {
+        return 0;
+      } else {
+        return -1;
+      }
     } else {
       // We known nothing so walk from the front of the playlist,
       // subtracting durations until we find a segment that contains
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -422,7 +422,6 @@
       i,
       segment,
       originalTime = time,
-      targetDuration = this.media_.targetDuration || 10,
       numSegments = this.media_.segments.length,
       lastSegment = numSegments - 1,
       startIndex,
@@ -442,70 +441,36 @@
 
     // find segments with known timing information that bound the
     // target time
-
-    // Walk backward until we find the first segment with timeline
-    // information that is earlier than `time`
-    for (i = lastSegment; i >= 0; i--) {
-      segment = this.media_.segments[i];
-      if (segment.end !== undefined && segment.end <= time) {
-        startIndex = i + 1;
-        knownStart = segment.end;
-        if (startIndex >= numSegments) {
-          // The last segment claims to end *before* the time we are
-          // searching for so just return it
-          return numSegments;
-        }
-        break;
-      }
-      if (segment.start !== undefined && segment.start <= time) {
-        if (segment.end !== undefined && segment.end > time) {
-          // we've found the target segment exactly
-          return i;
-        }
-        startIndex = i;
-        knownStart = segment.start;
-        break;
-      }
-    }
-
-    // Walk forward until we find the first segment with timeline
-    // information that is greater than `time`
     for (i = 0; i < numSegments; i++) {
       segment = this.media_.segments[i];
-      if (segment.start !== undefined && segment.start > time) {
-        endIndex = i - 1;
-        knownEnd = segment.start;
-        if (endIndex < 0) {
-          // The first segment claims to start *after* the time we are
-          // searching for so the target segment must no longer be
-          // available
-          return -1;
+      if (segment.end) {
+        if (segment.end > time) {
+          knownEnd = segment.end;
+          endIndex = i;
+          break;
+        } else {
+          knownStart = segment.end;
+          startIndex = i;
         }
-        break;
-      }
-      if (segment.end !== undefined && segment.end > time) {
-        endIndex = i;
-        knownEnd = segment.end;
-        break;
       }
     }
 
     // use the bounds we just found and playlist information to
     // estimate the segment that contains the time we are looking for
-
     if (startIndex !== undefined) {
       // We have a known-start point that is before our desired time so
       // walk from that point forwards
       time = time - knownStart;
       for (i = startIndex; i < (endIndex || numSegments); i++) {
         segment = this.media_.segments[i];
-        time -= segment.duration || targetDuration;
+        time -= segment.duration;
+
         if (time < 0) {
           return i;
         }
       }
 
-      if (i === endIndex) {
+      if (i >= endIndex) {
         // We haven't found a segment but we did hit a known end point
         // so fallback to interpolating between the segment index
         // based on the known span of the timeline we are dealing with
@@ -523,7 +488,8 @@
       time = knownEnd - time;
       for (i = endIndex; i >= 0; i--) {
         segment = this.media_.segments[i];
-        time -= segment.duration || targetDuration;
+        time -= segment.duration;
+
         if (time < 0) {
           return i;
         }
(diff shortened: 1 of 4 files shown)


src/playlist-loader.js
```diff
@@ -523,24 +488,32 @@
...
+
         if (time < 0) {
           return i;
         }
       }
-      // We haven't found a segment so load the first one
-      return 0;
+
+      // We haven't found a segment so load the first one if time is zero
+      if (time === 0) {
```
dmlap: Could we eliminate this extra condition by testing if `time <= 0` on line 493 above?

src/playlist-loader.js
```diff
@@ -523,24 +488,32 @@
...
+
       if (time < 0) {
         return -1;
       }
+
       for (i = 0; i < numSegments; i++) {
         segment = this.media_.segments[i];
-        time -= segment.duration || targetDuration;
+        time -= segment.duration;
         if (time < 0) {
```
dmlap: Do we need to handle the `time === 0` case?

src/videojs-hls.js
```diff
@@ -10,8 +10,8 @@ var
   // A fudge factor to apply to advertised playlist bitrates to account for
   // temporary flucations in client bandwidth
   bandwidthVariance = 1.2,
-  blacklistDuration = 5 * 60 * 1000, // 2 minute blacklist
-  TIME_FUDGE_FACTOR = 1 / 60, // Fudge factor to account for TimeRanges rounding
+  blacklistDuration = 5 * 60 * 1000, // 5 minute blacklist
```
dmlap: Why increase the blacklist time?

src/videojs-hls.js
```diff
@@ -10,8 +10,8 @@ var
...
+  TIME_FUDGE_FACTOR = 1 / 30, // Fudge factor to account for TimeRanges rounding
```
dmlap: Why increase the fudge factor?

src/videojs-hls.js
```diff
@@ -242,61 +242,57 @@ videojs.HlsHandler.prototype.handleSourceOpen = function() {
...
+  var
+    i, start, end,
+    result = [],
+    edges = [],
+    // In order to qualify as a possible candidate, the end point must:
+    //  1) Not have already existed in the `original` ranges
+    //  2) Not result from the shrinking of a range that already existed
+    //     in the `original` ranges
+    //  3) Not be contained inside of a range that existed in `original`
+    checkEdges = function(span) {
```
dmlap: I think this function would be easier to understand if the name of checkEdges() laid out what exactly it was checking. What about overlapsCurrentEnd()?

src/videojs-hls.js
```diff
@@ -242,61 +242,57 @@ videojs.HlsHandler.prototype.handleSourceOpen = function() {
...
+      return (span[0] <= end && span[1] > end) || span[1] === end;
```
dmlap: Could this be `span[0] <= end && span[1] >= end`?

src/videojs-hls.js
```diff
@@ -242,61 +242,57 @@ videojs.HlsHandler.prototype.handleSourceOpen = function() {
...
-    // this edge occurred outside the range of `original`
-    result.push(edges[i]);
   }
-  return result;
+
+  // we err on the side of caution and return null if didn't find
+  // exactly *one* differing end edge in the search above
+  if (result.length !== 1) {
+    return null;
+  } else {
```
dmlap: This `else` is unnecessary. I typically drop the nesting in that case but the current `else` clause is so short it's not a big deal. Just noting for consistency elsewhere.

src/videojs-hls.js
```diff
@@ -1008,17 +999,25 @@ videojs.HlsHandler.prototype.loadSegment = function(segmentInfo) {
...
+    // This is a timeout of a previously aborted segment request
+    // so simply ignore it
+    if (!self.segmentXhr_ || request !== self.segmentXhr_) {
+      return;
+    }
 
     // if a segment request times out, we may have better luck with another playlist
     if (request.timedout) {
+      // the segment request is no longer outstanding
+      self.segmentXhr_ = null;
```
dmlap: It looks like we're still nulling `segmentXhr_` for all code paths after this point. Why repeat it within each of the conditionals?

src/videojs-hls.js
```diff
@@ -1170,12 +1176,12 @@ videojs.HlsHandler.prototype.updateEndHandler_ = function () {
     currentMediaIndex,
     currentBuffered,
     seekable,
-    timelineUpdates;
+    timelineUpdate;
 
-  this.pendingSegment_ = null;
```
dmlap: I'm not clear on why this got distributed below either.

test/videojs-hls_test.js
```diff
@@ -2745,47 +2745,72 @@ test('does not download segments if preload option set to none', function() {
...
-    { end: 30}
-  ], 'detected a non-contiguous addition');
-});
-
-test('treats null buffered ranges as no addition', function() {
-  var edges = videojs.Hls.bufferedAdditions_(null,
-                                             videojs.createTimeRange([[0, 11]]));
-
-  equal(edges.length, 0, 'no additions');
+test('detects time range end-point changed by updates', function() {
```
dmlap: Beautiful tests!

src/videojs-hls.js
dmlap: ... and I should look more carefully before commenting.

src/playlist-loader.js
imbcmdth: Unfortunately no. That would select segments which we don't necessarily want because the time can equal 0 right at the end of a segment (it's the old exclusive segment end time issue) when we really want the next segment before that one.

The only time time can be 0 and we *do* want to fetch that segment is on the very first segment. It's a complicated case.

Say you have three segments:
1.  4.5 seconds long
1. 4.6 seconds long
1. 4.5 seconds long - this is the only segment you have information for
 - Ends at 13.6 exactly

The fetcher is going to try and fetch time 9.1. So we find the last segment and do `time = 13.6 - 9.1` so time is `4.5`. It then enters this loop and gets the first segment subtracting `4.5` and `time` is equal to `0` - if it fetches the last segment it enters a loop since it'll never really get the right segment.

src/playlist-loader.js
imbcmdth: I don't think we do when walking forward.

src/videojs-hls.js
imbcmdth: Because of Sintel! (It's doesn't start at zero and 1/60th is way too small a fudge for a 30fps video...)

src/videojs-hls.js
imbcmdth: This was something left over when I was befuddled by the randomness of failures. 

I knew something was causing stuff to run out of order. I also knew that the fill/drain cycle is locked by the truthyness of `pendingSegment_` and `segmentXhr_` so I was making sure that either `pendingSegment_` or `segmentXhr_` was always set by not clearing `segmentXhr_` until *after* `pendingSegment_` was set which required re-organization.

src/videojs-hls.js
imbcmdth: Same reason as above.

src/playlist-loader.js
dmlap: Ah, ok. The exclusive-end argument makes sense to me.

src/videojs-hls.js
dmlap: So it can be un-redistributed?

PR #464, opened 2015-12-02
fix calling abort on a sourceBuffer whose mediaSource.readyState != open

jvquarck:
#### Scenario
Loading hls source and changing source on end was firing an error for aborting a sourceBuffer whose parent mediaSource's readyState wasn't 'open'.

#### Reasoning
- Not sure if ffmpeg encoding related issue.
- Or browser-webkit related issue: https://w3c-test.org/media-source/SourceBuffer-abort-readyState.html
- Or maybe the fix is acceptable

#### Background
@lardawge claiming this in: https://github.com/videojs/videojs-contrib-hls/issues/382, among other mentions.

#### Conclussion
Not sure if fixing something or obscurifying another problem. Would appreciate some insight in this matter.

PR #466, opened 2015-12-05
(development branch) fix calling abort on a sourceBuffer whose mediaSource.readyState != open

forbesjo:
This is PR #464 by @jvquarck but against the development branch. 90afb6f seems to be the only applicable commit, bb12d03 cannot be applied due to conflicts (it doesn't exist).


dmlap:
Thanks @jvquarck and @forbesjo!

PR #470, opened 2015-12-10
Remove from buffer

imbcmdth:
Call `remove` on the sourceBuffer when loading a segment to trim the length of the buffer before the `currentTime`. Chrome will often fail to "garbage collect" sufficient space on a long-lived sourceBuffer and fail with a **QuotaExceeded** error. This is a work-around to force removal of old data from the sourceBuffer before we ever try to append.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -982,7 +982,26 @@ videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
 videojs.HlsHandler.prototype.loadSegment = function(segmentInfo) {
   var
     self = this,
-    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex];
+    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex],
+    removeToTime = 0,
+    seekable = this.seekable();
+
+  // Chrome has a hard limit of 150mb of buffer and a very conservative "garbage collector"
+  // We manually clear out the old buffer to ensure we don't trigger the QuotaExceeded error
+  // on the source buffer during subsequent appends
+  if (this.sourceBuffer) {
+    // If we have a seekable range use that as the limit for what can be removed safely
+    // otherwise remove anything older than 1 minute before the current play head
+    if (seekable.length && seekable.start(0) > 0) {
+      removeToTime = seekable.start(0);
+    } else {
+      removeToTime = this.tech_.currentTime() - 60;
+    }
+
+    if (removeToTime > 0) {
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -982,7 +982,26 @@ videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
...
+  // Chrome has a hard limit of 150mb of buffer and a very conservative "garbage collector"
+  // We manually clear out the old buffer to ensure we don't trigger the QuotaExceeded error
+  // on the source buffer during subsequent appends
+  if (this.sourceBuffer) {
+    // If we have a seekable range use that as the limit for what can be removed safely
+    // otherwise remove anything older than 1 minute before the current play head
+    if (seekable.length && seekable.start(0) > 0) {
+      removeToTime = seekable.start(0);
+    } else {
+      removeToTime = this.tech_.currentTime() - 60;
```
dmlap: This might be a controversial move but I'm ok with trying it for awhile.

src/videojs-hls.js
```diff
@@ -982,7 +982,26 @@ videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
...
     self = this,
-    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex];
+    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex],
+    removeToTime = 0,
+    seekable = this.seekable();
+
+  // Chrome has a hard limit of 150mb of buffer and a very conservative "garbage collector"
+  // We manually clear out the old buffer to ensure we don't trigger the QuotaExceeded error
+  // on the source buffer during subsequent appends
+  if (this.sourceBuffer) {
```
dmlap: You should check if the sourceBuffer is currently updating here, too.

test/videojs-hls_test.js
```diff
@@ -373,6 +373,39 @@ test('duration is set when the source opens after the playlist is loaded', funct
...
+  player.currentTime(120);
+  standardXHRResponse(requests[1]);
+  standardXHRResponse(requests[2]);
+
+  strictEqual(requests[0].url, 'manifest/master.m3u8', 'master playlist requested');
+  strictEqual(requests[1].url,
+              absoluteUrl('manifest/media3.m3u8'),
+              'media playlist requested');
+  equal(removes.length, 1, 'remove called');
+  deepEqual(removes[0], [0, 120 - 60], 'remove called with the right range');
```
dmlap: This doesn't test the live case which is the more important one, right?

PR #473, opened 2015-12-11
Firefox live

imbcmdth:
A few enhancements and a big fix:
* Check that the playlist is still the same uri when we try to update expired_ in playlist loader. If we switch playlists, changes in the media-sequence value may no longer be a valid way of determining the time that has expire. 
* Move the timestampOffset calculation from `drainBuffer` to `fillBuffer`. The state of the player can change between those two functions so we should calculate where to place a segment when we decided which segment to fetch.
* Correct seek-to-live behavior in firefox. Firefox doesn't allow you to seek until `canplay` is fired and doesn't fire `canplay` until after playback has begun. As a result, seeking to the right point in a live stream requires setupFirstPlay to wait for `canplay` and checking that the `readyState` signifies that we can *finally* seek.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -819,11 +819,14 @@ videojs.HlsHandler.prototype.fillBuffer = function(mediaIndex) {
   var
     tech = this.tech_,
     currentTime = tech.currentTime(),
+    hasBufferedContent = (this.tech_.buffered().length !== 0),
     currentBuffered = this.findBufferedRange_(),
+    outsideBufferedRanges = !(currentBuffered && currentBuffered.length),
     currentBufferedEnd = 0,
     bufferedTime = 0,
     segment,
-    segmentInfo;
+    segmentInfo,
+    segmentTimestampOffset;
 
   // if preload is set to "none", do not download segments until playback is requested
   if (this.loadingState_ !== 'segments') {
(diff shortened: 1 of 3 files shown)


src/videojs-hls.js
```diff
@@ -906,9 +915,39 @@ videojs.HlsHandler.prototype.fillBuffer = function(mediaIndex) {
...
+    timestampOffset: null
   };
 
+  if (mediaIndex > 0) {
+    segmentTimestampOffset = videojs.Hls.Playlist.duration(segmentInfo.playlist,
+      segmentInfo.playlist.mediaSequence + mediaIndex) + this.playlists.expired_;
+  }
+
+  // If we have seeked into a non-buffered time-range, remove all buffered
+  // time-ranges because they could have been incorrectly placed originally
```
dmlap: This comment seems out of date (specifically, the bits about clearing the buffer aren't at least out of order in this new location)

src/videojs-hls.js
```diff
@@ -906,9 +915,39 @@ videojs.HlsHandler.prototype.fillBuffer = function(mediaIndex) {
...
+    if (segmentInfo.playlist.discontinuityStarts.length) {
+      // Now that the forward buffer is clear, we have to set timestamp offset to
+      // the start of the buffered region
+      segmentInfo.timestampOffset = segmentTimestampOffset;
+    }
+  } else if (segment.discontinuity && currentBuffered.length) {
+    // If we aren't seeking and are crossing a discontinuity, we should set
+    // timestampOffset for new segments to be appended the end of the current
+    // buffered time-range
+    segmentInfo.timestampOffset = currentBuffered.end(0);
```
dmlap: Couldn't this case be handled using `segmentTimestampOffset` as well?

imbcmdth: As we discussed, I feel like this whole section must be revisited at some point in the near future but I am hesitant to make changes until we have the first official release of MSE-HLS out the door, so to speak.

src/videojs-hls.js
imbcmdth: Yes this was vestigial. I will remove/reword.

src/videojs-hls.js
dmlap: Ok, I wrote up #474 to keep track of that.

dmlap:
LGTM

PR #476, opened 2015-12-15
Fix the fixes for Firefox

imbcmdth:
* `readyState` is a function not a getter!
* only call 'canplay' once to avoid multiple the possibility of triggering the live start seek multiple times
* do not track expired time until we have successfully executed `setupFirstPlay`


forbesjo:
Looks good to me.

PR #503, opened 2016-01-11
Segment Loader

dmlap:
Pull out the logic for loading segments into its own class.

** Still a work-in-progress **


src/segment-loader.js
```diff
@@ -0,0 +1,400 @@
...
+        this.xhr_.onreadystatechange = null;
+        this.xhr_.abort();
+        this.xhr_ = null;
+      }
+
+      // clear out the segment being processed
+      this.pendingSegment_ = null;
+    },
+
+    fillBuffer_: function() {
```
imbcmdth: If we are going through the effort to pull these functions into their own file, maybe we should consider giving `fillBuffer`/`drainBuffer` better names?

src/videojs-hls.js
```diff
@@ -829,7 +846,14 @@ videojs.HlsHandler.prototype.fillBuffer = function(mediaIndex) {
...
     hasBufferedContent = (this.tech_.buffered().length !== 0),
-    currentBuffered = this.findBufferedRange_(),
+
+    // !!The order of the next two assignments is important!!
+    // `currentTime` must be equal-to or greater-than the start of the
+    // buffered range. Flash executes out-of-process so, every value can
+    // change behind the scenes from line-to-line. By reading `currentTime`
+    // after `buffered`, we ensure that it is always a current or later
+    // value during playback.
+    currentBuffered = videojs.Hls.Ranges.findRange_(tech.buffered(), tech.currentTime()),
```
imbcmdth: Not sure that the order-of-operations in this comment applies to what the code is doing..

src/videojs-hls.js
```diff
@@ -1208,7 +1232,14 @@ videojs.HlsHandler.prototype.updateEndHandler_ = function () {
...
   currentMediaIndex = segmentInfo.mediaIndex + (segmentInfo.mediaSequence - playlist.mediaSequence);
-  currentBuffered = this.findBufferedRange_();
+
+  // !!The order of the next two assignments is important!!
+  // `currentTime` must be equal-to or greater-than the start of the
+  // buffered range. Flash executes out-of-process so, every value can
+  // change behind the scenes from line-to-line. By reading `currentTime`
+  // after `buffered`, we ensure that it is always a current or later
+  // value during playback.
+  currentBuffered = videojs.Hls.Ranges.findRange_(this.tech_.buffered(), this.tech_.currentTime());
```
imbcmdth: Nor here either..

test/manifest/alternate-audio.m3u8
```diff
@@ -0,0 +1,9 @@
+#EXTM3U
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio",LANGUAGE="eng",NAME="English",AUTOSELECT=YES, DEFAULT=YES,URI="eng/prog_index.m3u8"
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio",LANGUAGE="fre",NAME="Français",AUTOSELECT=YES, DEFAULT=NO,URI="fre/prog_index.m3u8"
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio",LANGUAGE="sp",NAME="Espanol",AUTOSELECT=YES, DEFAULT=NO,URI="sp/prog_index.m3u8"
+
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=195023,CODECS="avc1.42e00a,mp4a.40.2",AUDIO="audio"
+lo/prog_index.m3u8
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=591680,CODECS="avc1.42e01e,mp4a.40.2",AUDIO="audio"
+hi/prog_index.m3u8
```
imbcmdth: No newline! :smiley:

test/test-helpers.js
```diff
@@ -0,0 +1,88 @@
...
+      restore: videojs.restoreEnvironment
+    };
+  };
+  videojs.restoreEnvironment = function() {
+    clock.restore();
+    videojs.xhr.XMLHttpRequest = window.XMLHttpRequest;
+    xhr.restore();
+  };
+
+})(window, window.videojs);
```
imbcmdth: This whole file :-1:

src/videojs-hls.js
dmlap: I think it still does. In this case, the order the arguments are evaluated as they are passed to findRange_() is actually relevant.

test/test-helpers.js
dmlap: :tophat:

src/source-updater.js
```diff
@@ -0,0 +1,120 @@
...
+        return videojs.createTimeRanges();
+      }
+      return this.sourceBuffer_.buffered;
+    },
+
+    /**
+     * Queue an update to set the duration.
+     * @see http://www.w3.org/TR/media-source/#widl-MediaSource-duration
+     */
+    duration: function(duration) {
```
dmlap: Note for the future: this doesn't work at all. Duration is set in MSE on the MediaSource, not the SourceBuffer, and requires all attached source buffers to be in an updatable state. That might mean we need an abstraction that manages updates all the way up at the MediaSource level but that seems complicated so I've skipped it so far.

dmlap:
The diff doesn't show it but I started moving tests in videojs-hls_test.js into a new module that uses the new test-helpers. I think we should move as much as possible over time but it would be a big effort and this PR is already gigantic.

src/segment-loader.js
```diff
@@ -0,0 +1,400 @@
...
+     * @param buffered {TimeRanges} the state of the buffer
+     * @param playlist {object} the playlist object to fetch segments from
+     * @param currentTime {number} the playback position in seconds
+     * @param timestampOffset (optional} {number} the duration of
+     * content, in seconds, that has been removed from the front of
+     * this playlist. If unspecified, it is assumed to be zero.
+     * @return {object} a segment info object that describes the
+     * request that should be made or null if no request is necessary
+     */
+    checkBuffer_: function(buffered, playlist, currentTime, timestampOffset) {
```
dmlap: I intentionally made this function explicit in the inputs it accepts, instead of relying on state stored on `this` (maybe I should have put it on the module object instead of making it instance level), so that we could easily unit test it. Hopefully this will make tracking down stream-sync issues easier to debug.

dmlap:
Merged into development (after many rebases)

PR #505, opened 2016-01-11
Duration fixes

imbcmdth:
- If the playlist hasn't been loaded yet, use 0.
- If the playlist duration is Infinity, use the playlist duration.
- If the mediaSource doesn't exist, use the playlist duration.
- In all other cases, use the mediaSource's duration.


src/videojs-hls.js
```diff
@@ -506,11 +506,23 @@ videojs.HlsHandler.prototype.setCurrentTime = function(currentTime) {
...
-    return videojs.Hls.Playlist.duration(playlists.media());
+    playlistDuration = videojs.Hls.Playlist.duration(playlists.media());
+  } else {
+    return 0;
+  }
+
+  if (playlistDuration === Infinity) {
+    return playlistDuration;
+  } else if (this.mediaSource) {
+    return this.mediaSource.duration;
```
dmlap: If the media source duration isn't NaN, should we ever be returning `Playlist.duration()`?

dmlap:
Tests!

src/videojs-hls.js
imbcmdth: Because of the way we fake mediaSource duration during live playlists, we need to return the playlist's duration because the mediaSource.duration will be an actual value so we can seek to the live point on play.

src/videojs-hls.js
```diff
@@ -297,6 +297,77 @@ videojs.Hls.findSoleUncommonTimeRangesEnd_ = function(original, update) {
...
+  segment = playlist.segments[segmentIndex];
+  previousSegment = playlist.segments[segmentIndex - 1];
+
+  if (segmentEnd && segment) {
+    segment.end = segmentEnd;
+
+    // fix up segment durations based on segment end data
+    if (!previousSegment) {
+      // first segment is always has a start time of 0 making it's duration
+      // equal to the segment end
```
dmlap: > of 0 making __its__ duration

src/videojs-hls.js
```diff
@@ -297,6 +297,77 @@ videojs.Hls.findSoleUncommonTimeRangesEnd_ = function(original, update) {
...
+
+/**
+ * Determines if we should call endOfStream on the media source based on the state
+ * of the buffer or if appened segment was the final segment in the playlist.
+ * @param playlist {object} a media playlist object
+ * @param segmentIndex {number} the index of segment we last appended
+ * @param currentBuffered {object} the buffered region that currentTime resides in
+ * @return {boolean} whether or not endOfStream was called on the MediaSource
+ */
+videojs.HlsHandler.prototype.maybeEndOfStream_ = function(playlist, segmentIndex, currentBuffered) {
```
dmlap: Subjective: I think it's better to split this kind of thing into a non-side-effecting function that determines whether endOfStream() should be called and perform the side-effect at the call site. It's easier to test if that's necessary and (hopefully) keeps the side effects clear and close together.

src/videojs-hls.js
```diff
@@ -506,11 +577,18 @@ videojs.HlsHandler.prototype.setCurrentTime = function(currentTime) {
...
+  var
+    playlists = this.playlists;
+
+  if (!playlists) {
+    return 0;
   }
-  return 0;
+
+  if (this.mediaSource) {
+    return this.mediaSource.duration;
```
dmlap: I'm kind of shocked we weren't doing this already...

src/videojs-hls.js
```diff
@@ -1205,7 +1288,7 @@ videojs.HlsHandler.prototype.updateEndHandler_ = function () {
     return;
   }
 
-  playlist = this.playlists.media();
+  playlist = segmentInfo.playlist;
```
dmlap: Given the check on line 1213/1298, why is this necessary?

imbcmdth: There are two reasons for this change:
1. It just makes it consistent that all the values we are pulling into local variables come from `segmentInfo`
1. I had planned to move `maybeEndOfStream_` functionality to *before* the early exit if we switched renditions. I wanted to signal the end-of-stream no matter what happens if we need to... 

Obviously, #2 never happened but since I shall be changing the function anyway given your other comment, I might as well do that too!

src/videojs-hls.js
imbcmdth: Probably just something "left on the cutting room floor" during source-handler conversion.

imbcmdth:
Closing to make PR against master

PR #507, opened 2016-01-11
Blacklist stuck manifests

imbcmdth:
Emit an error when a live playlist fails to update in a timely manner. This allows existing blacklisting code to run and attempt to recover by switching playlists.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-hls",
-  "version": "1.3.3",
+  "version": "1.3.4",
   "engines": {
     "node": ">= 0.10.12"
   },
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -96,7 +96,12 @@
 
       // a flag that disables "expired time"-tracking this setting has
       // no effect when not playing a live stream
-      this.trackExpiredTime_ = false;
+      loader.trackExpiredTime_ = false;
+
+      // a flag that signals if we are in the process of retrying the
+      // process of fetching a live playlist because the last fetch
+      // introduced no changes
+      loader.retryUpdate_ = false;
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-hls",
-  "version": "1.3.3",
+  "version": "1.3.4",
```
dmlap: This is a bit confusing. Did you forget to bump the package.json in the last release?

src/playlist-loader.js
```diff
@@ -96,7 +96,12 @@
 
       // a flag that disables "expired time"-tracking this setting has
       // no effect when not playing a live stream
-      this.trackExpiredTime_ = false;
+      loader.trackExpiredTime_ = false;
+
+      // a flag that signals if we are in the process of retrying the
+      // process of fetching a live playlist because the last fetch
+      // introduced no changes
+      loader.retryUpdate_ = false;
```
dmlap: Another flag, eh?

dmlap:
Needs tests.

package.json
imbcmdth: I did but then I pushed to development and these commits didn't go away. They are the same commit as the latest master so I think it's just cruft.

gkatsev: If you rebase against dev, it should fix itself.

src/playlist-loader.js
imbcmdth: Damnit you noticed! I couldn't think of a better way to detect that we have failed (twice) to detect changes.

tenacex:
@dmlap @imbcmdth https://github.com/videojs/videojs-contrib-hls/pull/655

BrandonOCasey:
Did this already get merged with #670

PR #508, opened 2016-01-11
Rendition selection

imbcmdth:
The rendition selection code has a lot of extra conditionals and an extra variable that can be removed making it easier to understand what is happening in there.


dmlap:
Good idea though it seems like something may have broken based on the Travis results.

imbcmdth:
I ended up making this slightly more complicated because there was still a case that didn't seem right to me - we always want to pick the highest bitrate we can play among the renditions that aren't too much larger than the player size.

The algorithm in https://github.com/videojs/videojs-contrib-hls/commit/af7f8e21f5b43340335106373527621b632beb61 was always selecting the lowest bitrate rendition if multiple renditions shared the same resolution. The later commit https://github.com/videojs/videojs-contrib-hls/commit/7671b5a0ae529576d07334cd7719e94b6d11b0cc corrects that.

src/videojs-hls.js
```diff
@@ -711,14 +709,24 @@ videojs.HlsHandler.prototype.selectPlaylist = function () {
       break;
     } else if (variant.attributes.RESOLUTION.width < width &&
```
dmlap: I think we're kind of inconsistent with this `if`. We guard the second clause with an `else if` even though we short-circuit the loop if the first clause is hit. Then in the third clause, we *don't* guard with an `else`, even though the second clause now short-circuits if it's run.

src/videojs-hls.js
```diff
@@ -711,14 +709,24 @@ videojs.HlsHandler.prototype.selectPlaylist = function () {
...
+
+    // If we still haven't found a good match so keep a
+    // reference to the previous variant for the next loop
+    // iteration
+
+    // By only saving variants if they are smaller than the
+    // previously saved variant, we ensure that we also pick
+    // the highest bandwidth variant that is just-larger-than
+    // the video player
+    if(!resolutionPlusOne ||
```
dmlap: Space between `if` and `(`.

dmlap:
@imbcmdth two nitpicky comments, otherwise LGTM

PR #511, opened 2016-01-12
browserifiy, pluginifiy, nodeifiy

BrandonOCasey:


package.json
```diff
@@ -1,54 +1,96 @@
...
-    "url": "git@github.com:videojs/videojs-contrib-hls.git"
-  },
+  "description": "A video.js tech that plays HLS video on platforms that don't support it but have Flash",
+  "author": "Brightcove",
   "license": "Apache-2.0",
-  "scripts": {
-    "test": "grunt test",
-    "prepublish": "if [ -z \"$TRAVIS\" ]; then grunt; fi"
-  },
+  "version": "0.0.0",
```
gkatsev: version got reset. Also, we try to keep the order of properties the same (name, version, description). This may have been fixed in a newer version of the generator.

PR #515, opened 2016-01-13
Support #EXT-X-BYTERANGE in video-contrib-hls v1.x

tenacex:
Adding tests and some minor fix-ups to original PR #454.

Thanks @KatrinaE for the feature!


src/videojs-hls.js
```diff
@@ -995,7 +1019,7 @@ videojs.HlsHandler.prototype.loadSegment = function(segmentInfo) {
       return;
     }
 
-    self.lastSegmentLoaded_ = segmentInfo.uri;
+    self.lastSegmentLoaded_ = segment;
```
dmlap: I don't think this works. Elsewhere in this file we're comparing `lastSegmentLoaded` to a URL which will always fail after this change.

KatrinaE: Are you referring to the `this.lastSegmentLoaded_ === this.playlistUriToUrl(segment.uri)` check on line 857? I had updated that comparison to look at the entire segment, but per your comment [here](https://github.com/videojs/videojs-contrib-hls/pull/454/files#r46886998) it sounds like that won't work. Is there some other way we can uniquely identify segments there? The quick and dirty option would be checking both URI and byterange as part of that if statement, but I'm not sure if that's up to your standards.

I've been away for a bit but have some bandwidth now to fix this up. Would you be okay with a quick fix like the one I suggest? Would you prefer a different type of identifier? Or is this not worth tackling right now?

I think this is the only other place in the file where we use `lastSegemntLoaded_`. If we can take care of this, it should be ready to go. Let me know what I can do to help.

dmlap: @KatrinaE: actually, I do think checking the URI and byterange seems like the right move here (I'd really like to get rid of `lastSegmentLoaded_` entirely but I'm not sure how to yet). If you have bandwidth, we'd love a helping hand to get this merged. If you could switch this to the URI and byterange check instead of the object equality one, I think we're good to go.

KatrinaE: I've added the byterange comparison, merged the development branch and @tenacex's changes (thanks for adding them!), and pushed them to [my fork](https://github.com/photoshelter/videojs-contrib-hls/tree/feature/byterange-support-dev). I can't update this PR since I don't have push access to @tenacex's fork. Should I create (yet another) one?

gkatsev: You could either PR to @tenacex's fork or just make a new PR to this repo.

KatrinaE: Right, didn't know if you guys had a preference.

KatrinaE: If not I will just PR to @tenacex to keep things cleaner.

imbcmdth: I think a PR to @tenacex will be cleaner. Thanks!

dmlap:
Closed via 7080865c5dbba6c01b46d649c7107d41326882b6. Thanks @KatrinaE and @tenacex!

PR #516, opened 2016-01-13
browserifiy, pluginifiy, and nodeifiy hls

BrandonOCasey:
switched to using plugin generator
code updated to use es6
code updated to follow node conventions
moved unused files to a test/files folder


CONTRIBUTING.md
```diff
@@ -1,31 +1 @@
...
-## Submitting pull requests
-
-1. Create a new branch, please don't work in your `master` branch directly.
-1. Add failing tests for the change you want to make. Run `grunt` to see the tests fail.
-1. Fix stuff.
-1. Run `grunt` to see if the tests pass. Repeat steps 2-4 until done.
-1. Open `test/*.html` unit test file(s) in actual browser to ensure tests pass everywhere.
-1. Update the documentation to reflect any changes.
-1. Push to your fork and submit a pull request.
+# CONTRIBUTING
```
dmlap: I know this is a bit out of date but why delete everything?

LICENSE
```diff
@@ -11,39 +11,3 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-
-The AES decryption implementation in this project is derived from the
```
dmlap: This additional license information is important, it cannot be deleted.

README.md
```diff
@@ -1,13 +1,11 @@
-# video.js HLS Source Handler
+# video.js HLS Tech
```
dmlap: HLS is not a tech, it is a source handler.

dmlap: It looks like this whole file is out of date. I think that means we need a rebase on this PR.

package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "videojs-contrib-hls",
   "version": "1.3.4",
+  "description": "A video.js tech that plays HLS video on platforms that don't support it but have Flash",
```
dmlap: Out of date.

src/decrypter/aes.js
```diff
@@ -0,0 +1,206 @@
+/**
+ * Schedule out an AES key for both encryption and decryption. This
+ * is a low-level class. Use a cipher mode to do bulk encryption.
+ *
+ * @constructor
+ * @param key {Array} The key as an array of 4, 6 or 8 words.
+ */
+const AES = function(key) {
```
dmlap: This is the file that actually needs the license notice from SJCL

src/decrypter/index.js
```diff
@@ -0,0 +1,47 @@
...
+ * BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+ * OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN
+ * IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation
+ * are those of the authors and should not be interpreted as representing
+ * official policies, either expressed or implied, of the authors.
+ */
+
```
dmlap: This file doesn't actually contain any of the code referenced in this notice.

src/decrypter/ntoh.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Convert network-order (big-endian) bytes into their little-endian
+ * representation.
+ */
+const ntoh = function(word) {
```
dmlap: Personally, I think breaking out single functions that aren't used outside a particular module is overkill and makes things harder to track down.

src/m3u8/line-stream.js
```diff
@@ -0,0 +1,27 @@
...
+  * line.
+  */
+export default class LineStream extends Stream {
+  constructor() {
+    super();
+    this.buffer = '';
+  }
+
+  /**
+    * Add new data to be parsed.
```
dmlap: Indent is off here

src/m3u8/parser.js
```diff
@@ -0,0 +1,251 @@
...
+                });
+              }
+
+              // setup an encryption key for upcoming segments
+              key = {
+                method: entry.attributes.METHOD || 'AES-128',
+                uri: entry.attributes.URI
+              };
+
+              if (typeof entry.attributes.IV !== 'undefined') {
```
dmlap: Why use `typeof` checks to find `undefined`?

src/plugin.js
```diff
@@ -3,32 +3,58 @@
...
+import utils from './bin-utils';
+import xhr from './xhr';
+import resolveUrl from './resolve-url';
+import videojs from 'video.js';
+
+const Player = videojs.getComponent('Player');
+
+if (typeof Player.prototype.MediaSource === 'undefined') {
+  videojs.plugin('MediaSource', MediaSource);
+  videojs.plugin('URL', URL);
```
dmlap: These objects don't make sense as plugins to me. Plugins are intended for player-level customizations. MediaSource and URL don't affect individual player operation.

dmlap:
This looks like it was a HUGE effort, thank you! That's kind of a double-edged sword though-- it's so much change, it's really hard to review. Is there any way you can break this up into more bite-sized chunks so the changes are easier to verify?

src/m3u8/parser.js
BrandonOCasey: vjsstandard (linter) complains about directly using undefined, not sure why other than that

gkatsev: `typeof` returns a string, so yes. Why not just `if (entry.attributes.IV) {...}`?

src/plugin.js
```diff
@@ -110,60 +137,63 @@ videojs.HlsSourceHandler = function(mode) {
...
+  if (Hls.supportsNativeHls()) {
     return false;
   }
   return mpegurlRE.test(type);
 };
 
 // register source handlers with the appropriate techs
-if (videojs.MediaSource.supportsNativeMediaSources()) {
-  videojs.getComponent('Html5').registerSourceHandler(videojs.HlsSourceHandler('html5'));
+if (MediaSource.prototype.supportsNativeMediaSources()) {
```
imbcmdth: These type functions should be static functions so that they are at the constructor level and not in the prototype.

src/plugin.js
```diff
@@ -406,10 +439,11 @@ videojs.HlsHandler.prototype.setupSourceBuffer_ = function() {
...
  * player and video are loaded and initialized.
  */
-videojs.HlsHandler.prototype.setupFirstPlay = function() {
-  var seekable, media;
-  media = this.playlists.media();
+HlsHandler.prototype.setupFirstPlay = function() {
+  let seekable;
+  let media;
 
+  media = this.playlists.media();
```
imbcmdth: Might as well make this assignment part of the media variable's declaration above.

src/plugin.js
```diff
@@ -548,15 +583,17 @@ videojs.HlsHandler.prototype.seekable = function() {
...
  */
-videojs.HlsHandler.prototype.updateDuration = function(playlist) {
-  var oldDuration = this.mediaSource.duration,
-      newDuration = videojs.Hls.Playlist.duration(playlist),
-      setDuration = function() {
-        this.mediaSource.duration = newDuration;
-        this.tech_.trigger('durationchange');
+HlsHandler.prototype.updateDuration = function(playlist) {
+  let oldDuration = this.mediaSource.duration;
+
```
imbcmdth: Extra line.

src/plugin.js
```diff
@@ -969,11 +1015,11 @@ videojs.HlsHandler.prototype.playlistUriToUrl = function(segmentRelativeUrl) {
  *  * `bytesReceived` - amount of bytes downloaded
  * `bandwidth` is the only required property.
  */
-videojs.HlsHandler.prototype.setBandwidth = function(xhr) {
+HlsHandler.prototype.setBandwidth = function(localxhr) {
```
imbcmdth: We capitalize the first letter of multi-word variable names (and we don't capitalize the whole acronym) so this would be `localXhr`.

src/plugin.js
```diff
@@ -1006,26 +1053,34 @@ videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
   nextPlaylist = this.selectPlaylist();
 
   if (nextPlaylist) {
-    videojs.log.warn('Problem encountered with the current HLS playlist. Switching to another playlist.');
+    videojs.log.warn(
+      'Problem encountered with the current' +
+      'HLS playlist. Switching to another playlist.'
```
imbcmdth: Missing a space after the word 'current'.

src/plugin.js
```diff
@@ -1152,29 +1207,31 @@ videojs.HlsHandler.prototype.drainBuffer = function() {
...
-      decrypter = new videojs.Hls.Decrypter(segmentInfo.encryptedBytes,
-                                            segment.key.bytes,
-                                            segIv,
-                                            function(err, bytes) {
-                                              segmentInfo.bytes = bytes;
-                                            });
-      segmentInfo.decrypter = decrypter;
-      return;
     }
+    // if the media sequence is greater than 2^32, the IV will be incorrect
```
imbcmdth: You removed the else which is possible since all of the other cases explicitly return. We tend to keep the else (even if it is redundant) just to make the control-flow explicit. Also it protects us if, for some reason, one of the previous clauses are changed to no longer return.

src/plugin.js
```diff
@@ -1152,29 +1207,31 @@ videojs.HlsHandler.prototype.drainBuffer = function() {
...
+    segIv = segment.key.iv ||
+      new Uint32Array([0, 0, 0, mediaIndex + playlist.mediaSequence]);
+
+    // create a decrypter to incrementally decrypt the segment
+    decrypter = new Hls.Decrypter(
+      segmentInfo.encryptedBytes,
+      segment.key.bytes,
+      segIv,
+      function(err, localBytes) {
+        if (err) {
```
imbcmdth: Nice, though I am not sure that we actually emit errors from this function. Might want to do something like what we do when a key fails though instead of simply throwing.

PR #521, opened 2016-01-18
fixed bug where initial playlist was blacklisted

tenacex:
Made sure loadedmetadata was triggered when the second (non-failing playlist) is eventually loaded and parsed.


src/videojs-hls.js
```diff
@@ -1007,7 +1007,7 @@ videojs.HlsHandler.prototype.blacklistCurrentPlaylist_ = function(error) {
 
   if (nextPlaylist) {
     videojs.log.warn('Problem encountered with the current HLS playlist. Switching to another playlist.');
-
+    this.player.trigger('loadedmetadata');
```
dmlap: This is too early to fire `loadedmetada`. It's supposed to fire when we've successfully *received* the first media playlist. This fires right before we make the second request.

PR #526, opened 2016-01-20
converted the hls project to generator conventions

BrandonOCasey:
> ~~NOTE: tests will not pass from the command line~~

* moved non-test files to a top level utils folder
* used generator to setup new file structure and build pipeline
* removed the use of grunt


package.json
```diff
@@ -8,47 +10,109 @@
...
+    "build:test": "node scripts/build-test.js",
+    "clean": "node -e \"var s=require('shelljs'),d=['dist','dist-test','es5'];s.rm('-rf',d);s.mkdir('-p',d);\"",
+    "docs": "npm-run-all docs:*",
+    "docs:api": "jsdoc src -r -d docs/api",
+    "docs:toc": "doctoc README.md",
+    "lint": "vjsstandard :",
+    "prestart": "npm-run-all docs build",
+    "start": "npm-run-all -p start:* watch:*",
+    "start:serve": "babel-node scripts/server.js",
+    "pretest": "npm-run-all lint build",
```
gkatsev: If you remove `lint` from this *for now*, it won't fail linting. We can then have a separate PRs for the browserification and es6ification and then we can add it back in then.

BrandonOCasey: right now we just lint ':' which pretty much means lint nothing.

.travis.yml
```diff
@@ -1,23 +1,12 @@
...
-  - "stable"
-install:
-- npm install -g grunt-cli && npm install
-notifications:
-  hipchat:
-    rooms:
-      secure: l5TTd5JuPAW883PtcyaIBcJI9Chr9JpsZPQAEUBKAgIEwzuS6y7t5arlkS1PwH6gi1FADzYDf+OXSIou4GkTSrIetnBcT/SAgF0gBKgIhj+eRkuCfZ4VaC7BPhfZ0hgYRE+5Ejf5BM2MJafRm0pj7OlqG4xKrQZwtuV1te5r3JY=
-  irc:
-    channels:
-      - "chat.freenode.net#videojs"
```
dmlap: hipchat and irc notices should not be removed.

index.html
```diff
@@ -0,0 +1,72 @@
...
+  <script src="/node_modules/videojs-contrib-media-sources/dist/videojs-media-sources.js"></script>
+  <script src="/node_modules/pkcs7/dist/pkcs7.unpad.js"></script>
+  <script src="/src/videojs-hls.js"></script>
+  <script src="/src/xhr.js"></script>
+  <script src="/src/stream.js"></script>
+  <script src="/src/m3u8/m3u8-parser.js"></script>
+  <script src="/src/playlist.js"></script>
+  <script src="/src/playlist-loader.js"></script>
+  <script src="/src/decrypter.js"></script>
+  <script src="/src/bin-utils.js"></script>
```
dmlap: Is this page functional?

src/plugin.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Default options for the plugin.
+const defaults = {};
+
+/**
+ * Function to invoke when the player is ready.
+ *
+ * This is a great place for your plugin to initialize itself. When this
+ * function is called, the player will have its DOM and child components
+ * in place.
```
dmlap: Stub comment.

src/plugin.js
```diff
@@ -0,0 +1,42 @@
...
+ *           An object of options left to the plugin author to define.
+ */
+const contribHls = function(options) {
+  this.ready(() => {
+    onPlayerReady(this, videojs.mergeOptions(defaults, options));
+  });
+};
+
+// Register the plugin with video.js.
+videojs.plugin('contribHls', contribHls);
```
dmlap: contrib-hls isn't a plugin, so this entire file should be removed.

package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "videojs-contrib-hls",
   "version": "1.3.4",
+  "description": "Play back HLS with video.js, even where it's not natively supported",
+  "main": "es5/plugin.js",
```
dmlap: contrib-hls is not a plugin so this shouldn't be the main file.

test/data/manifests.js
```diff
@@ -0,0 +1,1036 @@
+window.manifests = {
```
dmlap: Should this be committed? Previously this was generated as part of the build process.

test/index.html
```diff
@@ -0,0 +1,39 @@
...
+  <script src="/test/data/manifests.js"></script>
+  <script src="/test/data/expected.js"></script>
+  <script src="/test/data/ts-segment-bc.js"></script>
+
+
+  <script src="/test/videojs-hls.test.js"></script>
+  <script src="/test/m3u8.test.js"></script>
+  <script src="/test/playlist.test.js"></script>
+  <script src="/test/playlist-loader.test.js"></script>
+  <script src="/test/decrypter.test.js"></script>
```
dmlap: Do these imports still work?

test/karma/firefox.js
```diff
@@ -0,0 +1,8 @@
+var common = require('./common');
+
+module.exports = function(config) {
+  config.set(common({
+    plugins: ['karma-firefox-launcher'],
+    browsers: ['Firefox']
+  }));
+};
```
dmlap: Can we just combine all of these individual browser modules into a single file? They're only eight lines apiece.

utils/perf.html
```diff
@@ -35,14 +22,28 @@
...
+<!-- HLS plugin -->
+<script src="../src/video-js-hls.js"></script>
+<script src="../src/flv-tag.js"></script>
+<script src="../src/exp-golomb.js"></script>
+<script src="../src/h264-stream.js"></script>
+<script src="../src/aac-stream.js"></script>
+<script src="../src/segment-parser.js"></script>
+
+<!-- MPEG-TS segment -->
+<script src="../test/data/ts-segment-bc.js"></script>
```
dmlap: Just curious, why did this move?

dmlap:
The tests aren't running because a number of dependencies aren't in scope in the test environment. I know we're trying to keep this PR small but I think getting the deps imported into the test harness wouldn't add much line-wise. It would also be a good test to show the new build process is doing sane things.

test/karma/firefox.js
BrandonOCasey: I think we have all these files so that you can test one specific browser at a time example: 
npm run test:chome. 

There is also a "detected" file which will detect which browsers you have available and test on all of them. This is the main thing that will run when you do npm run test.

utils/perf.html
BrandonOCasey: moved it with all the other files that were not specifically unit tests

test/karma/firefox.js
dmlap: I don't think running one specific browser at a time means we need separate files for each of the configs but that sounds like it's a generator problem, not an issue with this patch.

utils/perf.html
dmlap: I didn't even notice that! :) I was actually asking why the `<script>` elements moved from the top of the file to the bottom.

scripts/manifest-data.js
```diff
@@ -0,0 +1,80 @@
...
+var clean = function() {
+  try {
+    fs.unlinkSync(manifestFilepath);
+  } catch(e) {
+    // ignore error, must not exist
+  }
+  try {
+    fs.unlinkSync(expectedFilepath);
+  } catch(e) {
+    // ignore error, must not exist
```
dmlap: It would be wise to at least emit warnings in these cases.

PR #527, opened 2016-01-21
Working initial 404 error

tenacex:
Bug is now addressed.


imbcmdth:
Needs tests.

src/playlist-loader.js
```diff
@@ -104,6 +105,13 @@
 
       // update the playlist loader's state in response to a new or
       // updated playlist.
+      haveMetadataFirst = function(error, xhr, url) {
+        if (error) {
+          loader.state = 'HAVE_MASTER';
```
imbcmdth: We only want to reset the loader.state to 'HAVE_MASTER' if it was in the 'HAVE_MASTER' state before the request that caused the error. Otherwise, this would trigger a loadedmetadata each time we switch playlists in response to an error.

src/playlist-loader.js
```diff
@@ -104,6 +105,13 @@
 
       // update the playlist loader's state in response to a new or
       // updated playlist.
+      haveMetadataFirst = function(error, xhr, url) {
```
imbcmdth: I'm not sure you need a new function to hold this functionality.

src/playlist-loader.js
tenacex: Temporarily I changed the code back to what you had originally. Lets talk about how you want this code to be improved because I am not quite sure I understand exactly what needs to be improved.

src/playlist-loader.js
tenacex: Reverted back to only error checking and reverting loader.state.

test/manifest/4042.m3u8
```diff
@@ -0,0 +1,6 @@
+#EXTM3U
+#EXT-X-VERSION:3
+#EXT-X-STREAM-INF:BANDWIDTH=1799999,AVERAGE-BANDWIDTH=1799999,CODECS="avc1.77.30,mp4a.40.2",RESOLUTION=640x360
+404-1.m3u8
+#EXT-X-STREAM-INF:BANDWIDTH=1799999999999999,AVERAGE-BANDWIDTH=179999999999999,CODECS="avc1.77.30,mp4a.40.2",RESOLUTION=640x360
+https://solutions.brightcove.com/jwhisenant/hls/brightcove/fishtank/16x9-672819-640x360-rendition.m3u8
```
imbcmdth: Make this a local URI to a resource.

test/videojs-hls_test.js
```diff
@@ -1296,6 +1296,26 @@ test('blacklists switching from video-only playlists to video+audio', function()
   equal(videoAudioPlaylist.excludeUntil, Infinity, 'excluded incompatible playlist');
 });
 
+test('Switches to HAVE_METADATA if 404', function() {
```
imbcmdth: Couple of comments:

The goal of the test should be to ensure that `loadedmetadata` event is triggered on a successful load after a 404 for the first playlist. You should bind an event listener to `loadedmetadata` and record the number of times it was called. After the 404 it should have been called `0` times and after the successful playlist load it should be `1`.

test/videojs-hls_test.js
```diff
@@ -1296,6 +1296,28 @@ test('blacklists switching from video-only playlists to video+audio', function()
   equal(videoAudioPlaylist.excludeUntil, Infinity, 'excluded incompatible playlist');
 });
 
+test('loadedMedia event only fire once we have successfully loaded a playlist', function() {
```
imbcmdth: Name should be `loadedmetadata`. How about "After an initial media playlist 404s, we fire loadedmetadata once we successfully load a playlist"? :)

imbcmdth:
Once you fix the name of the test, this looks good. We should squash this so that we can merge a single clean commit for this fix.

dmlap:
LGTM module @imbcmdth's comment as well.

PR #533, opened 2016-01-29
Make a constant for the live edge

dmlap:
Don't use the literal "3" for the number of segments required for stable live playback.

FYI @tomjohnson916


imbcmdth:
I believe the HLS spec specifies the live point as 3 *target durations* from the end of the playlist... LGTM!

PR #534, opened 2016-01-29
browserify-p2: m3u8, stream, and stub

BrandonOCasey:
stub old functionality to get unit tests working for now
stub old test functionality to get unit tests working for now
fixed some issues with the first browserify part
convert m3u8 and stream modules to es6/node/browserify
fixed npm scripts to work for current branch


src/m3u8/index.js
```diff
@@ -5,111 +5,110 @@
...
+ * property, `line`, which captures the entirety of the input without
+ * interpretation. Comments similarly have a single additional property
+ * `text` which is the input without the leading `#`.
+ *
+ * Tags always have a property `tagType` which is the lower-cased version of
+ * the M3U8 directive without the `#EXT` or `#EXT-X-` prefix. For instance,
+ * `#EXT-X-MEDIA-SEQUENCE` becomes `media-sequence` when parsed. Unrecognized
+ * tags are given the tag type `unknown` and a single additional property
+ * `data` with the remainder of the input.
+ */
```
dmlap: This is a class-level comment on the ParseStream object. Seems misplaced.

src/m3u8/index.js
```diff
@@ -320,7 +317,7 @@
         event.attributes = parseAttributes(match[1]);
         // parse the IV string into a Uint32Array
         if (event.attributes.IV) {
-          if (event.attributes.IV.substring(0,2) === '0x') {
+          if (event.attributes.IV.substring(0, 2) === '0x') {
```
dmlap: We should collapse this into the outer `if`. Not vital for this PR.

src/m3u8/index.js
```diff
@@ -341,37 +338,23 @@
...
-   */
-  Parser = function() {
-    var
-      self = this,
-      uris = [],
-      currentUri = {},
-      key;
-    Parser.prototype.init.call(this);
 
+export class Parser extends Stream {
```
dmlap: Class comment is missing.

src/m3u8/index.js
```diff
@@ -509,14 +493,16 @@
               if (!currentUri.attributes) {
                 currentUri.attributes = {};
               }
-              currentUri.attributes = mergeOptions(currentUri.attributes,
-                                                   entry.attributes);
+              currentUri.attributes = mergeOptions(
+                currentUri.attributes,
+                entry.attributes
+              );
```
dmlap: Very nipticky but we don't line break after parens elsewhere in this project.

src/stream.js
```diff
@@ -2,73 +2,84 @@
...
-      };
-    };
-  };
+export default class Stream {
+  constructor() {
+    this.init();
+  }
+
+  init() {
+    this.listeners = {};
```
dmlap: Seems worth consolidating `contructor` and `init` while we're in here.

src/stub.js
```diff
@@ -0,0 +1,10 @@
+import m3u8 from './m3u8';
```
dmlap: I do wish this file name was more descriptive. Are you against videojs-hls.js for the project entry point?

test/index.html
```diff
@@ -16,22 +16,16 @@
   <script src="/node_modules/video.js/dist/video.js"></script>
   <script src="/node_modules/videojs-contrib-media-sources/src/videojs-media-sources.js"></script>
 
-  <script src="/src/videojs-hls.js"></script>
+  <script src="/src/videojs-contrib-hls.js"></script>
   <script src="/src/xhr.js"></script>
-  <script src="/src/stream.js"></script>
-  <script src="/src/m3u8/m3u8-parser.js"></script>
+  <script src="/dist/videojs-contrib-hls.js"></script>
```
dmlap: This mix of src and dist dependencies is surprising to me. Is that because this is only one step of the migration?

test/karma/detected.js
```diff
@@ -29,6 +29,7 @@ module.exports = function(config) {
         postDetection: function(availableBrowsers) {
           var safariIndex = availableBrowsers.indexOf('Safari');
           if(safariIndex !== -1) {
+            console.log("Not running safari it is/was broken");
```
dmlap: We should file an issue for this.

test/m3u8.test.js
```diff
@@ -1,646 +1,650 @@
...
-    setup: function() {
-      lineStream = new LineStream();
+import {ParseStream, LineStream, Parser} from '../src/m3u8';
+import QUnit from 'qunit';
+import testDataExpected from './data/expected.js';
+import testDataManifests from './data/manifests.js';
+
+QUnit.module('LineStream', {
+  beforeEach() {
+    this.lineStream = new LineStream();
```
dmlap: Why not use module-level variables for these items? Technically the scope could leak across QUnit modules but it saves a ton of `this`s.

test/index.html
BrandonOCasey: yeah dist is the only thing that we will need to include in the future. Right now it only contains stub.js (which adds our objects to HLS directly so that the unit tests will still work), m3u8, and stream. As we go through the stages things will be removed from here and into that dist file.

src/stub.js
BrandonOCasey: stub.js will be removed on the final part of these pull requests and videojs-contrib-hls.js will be set as the main file again. stub.js is the main file for now because it allows us to keep all the unit tests passing by stubbing old functionality (adding m3u8 to video.js and Stream to videojs.Hls and more as the PRs go on).

src/stream.js
BrandonOCasey: I thought about this as well, I just was not sure how backwards compatible that was.

test/m3u8.test.js
```diff
@@ -1,646 +1,650 @@
...
+
+  QUnit.ok(element, 'an event was triggered');
+  QUnit.strictEqual(element.type, 'tag', 'the line type is tag');
+  QUnit.strictEqual(element.tagType, 'stream-inf', 'the tag type is stream-inf');
+  QUnit.strictEqual(element.attributes.NUMERIC, '24', 'numeric attributes are parsed');
+  QUnit.strictEqual(
+    element.attributes.ALPHA,
+    'Value',
+    'alphabetic attributes are parsed'
+  );
```
dmlap: Parens-line-breaks

test/m3u8.test.js
BrandonOCasey: i was running into issues in a few of these files related to tests not cleaning up after themselves when they modify the global, or redeclaring the global in their scope (and the linter complaining that the variable was already declared). I have also had problems with tests in the past that use globals which caused future tests to fail/pass making it very confusing to debug. the generator also generated several tests that all used this. so I assumed that was the way we probably wanted to go.

That being said I am not opposed to reverting it to not us `this.`. It just seemed easier to reason about tests if there was no chance that scope would leak.

test/stub.test.js
```diff
@@ -0,0 +1,5 @@
+import manifests from './data/manifests';
```
dmlap: Same naming question with this file. What about test-manifests.js?

test/karma/detected.js
BrandonOCasey: The old unit tests only ran against Chome/Firefox and when I attempted to run them in safari they all failed. Once I updated the code I tried to run them again and they failed, so I turned off safari. Safari does launch and run, but a lot of the unit tests fail.

test/stub.test.js
BrandonOCasey: I will change the name of this file as It makes a lot more sense

src/m3u8/index.js
BrandonOCasey: this has been fixed

src/stream.js
dmlap: Stream is an internal prototype. I think it's fine to consolidate.

test/index.html
dmlap: Ok, makes sense.

test/karma/detected.js
dmlap: Yeah, that makes sense. It's not where we *want* to be though, so it would be good to at least keep track of it.

dmlap: Filed it: https://github.com/videojs/videojs-contrib-hls/issues/538

src/stub.js
dmlap: Ah, ok.

PR #535, opened 2016-01-30
Fix the parsing of the ZEN-TOTAL-DURATION tag

cllu:
The original implementation would always skip the `ZEN-TOTAL-DURATION` tag in [line 129](https://github.com/videojs/videojs-contrib-hls/compare/master...cllu:patch-1#diff-a523235bb2d98ee8619a642cbd1fd307L129).

### DIFF
--- a/src/m3u8/m3u8-parser.js
+++ b/src/m3u8/m3u8-parser.js
@@ -126,7 +126,7 @@
     }
 
     // Comments
-    if (line.indexOf('#EXT') !== 0) {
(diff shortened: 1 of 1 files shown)


forbesjo:
@imbcmdth do you think we should bring this in?

imbcmdth:
There is no reason **not** to bring this in - it seems like a bug in the first place. Rebase and merge at your leisure. 🍊

BrandonOCasey:
it will have to be moved into https://github.com/videojs/m3u8-parser as this code was recently moved there

BrandonOCasey:
closing as this code has been moved out of HLS

PR #540, opened 2016-02-02
browserify-p3: decrypter

BrandonOCasey:
convert decrypter module to es6/node/browserify
stub old functionalify to get all tests working
updated scripts to keep builds and tests working

### DIFF
--- a/src/decrypter.js
+++ b/src/decrypter.js
@@ -353,10 +353,8 @@ export class AsyncStream extends Stream {
   processJob_() {
     this.jobs.shift()();
     if (this.jobs.length) {
-      this.timeout_ = setTimeout(
-        this.processJob_.bind(this),
-        this.delay
-      );
+      this.timeout_ = setTimeout(this.processJob_.bind(this),
+                                 this.delay);
     } else {
       this.timeout_ = null;
     }
@@ -364,10 +362,8 @@ export class AsyncStream extends Stream {
   push(job) {
     this.jobs.push(job);
     if (!this.timeout_) {
-      this.timeout_ = setTimeout(
-        this.processJob_.bind(this),
-        this.delay
-      );
+      this.timeout_ = setTimeout(this.processJob_.bind(this),
+                                 this.delay);
     }
   }
 }
--- a/src/decrypter.js
+++ b/src/decrypter.js
@@ -49,59 +49,6 @@ const ntoh = function(word) {
     (word >>> 24);
 };
 
-/**
- * Expand the S-box tables.
- *
- * @private
- */
-const precompute = function() {
-  let _tables = [[[], [], [], [], []], [[], [], [], [], []]];
-  let encTable = _tables[0];
-  let decTable = _tables[1];
-  let sbox = encTable[4];
-  let sboxInv = decTable[4];
-  let i;
-  let x;
-  let xInv;
-  let d = [];
-  let th = [];
-  let x2;
-  let x4;
-  let x8;
-  let s;
-  let tEnc;
-  let tDec;
-
-  // Compute double and third tables
-  for (i = 0; i < 256; i++) {
-    th[(d[i] = i << 1 ^ (i >> 7) * 283) ^ i] = i;
-  }
-
-  for (x = xInv = 0; !sbox[x]; x ^= x2 || 1, xInv = th[xInv] || 1) {
-    // Compute sbox
-    s = xInv ^ xInv << 1 ^ xInv << 2 ^ xInv << 3 ^ xInv << 4;
-    s = s >> 8 ^ s & 255 ^ 99;
-    sbox[x] = s;
-    sboxInv[s] = x;
-
-    // Compute MixColumns
-    x8 = d[x4 = d[x2 = d[x]]];
-    tDec = x8 * 0x1010101 ^ x4 * 0x10001 ^ x2 * 0x101 ^ x * 0x1010100;
-    tEnc = d[s] * 0x101 ^ s * 0x1010100;
-
-    for (i = 0; i < 4; i++) {
-      encTable[i][x] = tEnc = tEnc << 24 ^ tEnc >>> 8;
-      decTable[i][s] = tDec = tDec << 24 ^ tDec >>> 8;
-    }
-  }
-
-  // Compactify. Considerable speedup on Firefox.
-  for (i = 0; i < 5; i++) {
-    encTable[i] = encTable[i].slice(0);
-    decTable[i] = decTable[i].slice(0);
-  }
-  return _tables;
-};
 
 let aesTables;
 
@@ -126,10 +73,7 @@ class AES {
     *
     * @private
     */
-    if (!aesTables) {
-      aesTables = precompute();
-    }
-    this._tables = JSON.parse(JSON.stringify(aesTables));
+    this._tables = this._precompute();
     let i;
     let j;
     let tmp;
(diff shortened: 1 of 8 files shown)


src/decrypter.js
```diff
@@ -35,21 +49,6 @@
  * are those of the authors and should not be interpreted as representing
  * official policies, either expressed or implied, of the authors.
  */
```
dmlap: This notice should be at the top of the file.

src/decrypter.js
```diff
@@ -58,108 +57,74 @@ ntoh = function(word) {
...
+    * There are two tables, _tables[0] is for encryption and
+    * _tables[1] is for decryption.
+    *
+    * The first 4 sub-tables are the expanded S-box with MixColumns. The
+    * last (_tables[01][4]) is the S-box itself.
+    *
+    * @private
+    */
+    this._tables = [[[], [], [], [], []], [[], [], [], [], []]];
+    this._precompute();
```
dmlap: I think you'll be re-pre-computing for every AES instance by moving this into the constructor.

dmlap: Erm... or we were also re-computing in the old version. Still, I think this step can be done once and then re-used by each of the instances.

dmlap: ...and now I see that we mutate the sub-arrays. precompute() still seems idempotent but maybe not worth the effort. Thoughts?

src/decrypter.js
```diff
@@ -229,24 +266,34 @@ AES.prototype = {
...
-
-    // iteration variable
-    wordIx;
+/* eslint-enable max-len */
+export const decrypt = function(encrypted, key, initVector) {
+  // word-level access to the encrypted bytes
+  let encrypted32 = new Int32Array(
+    encrypted.buffer,
+    encrypted.byteOffset,
+    encrypted.byteLength >> 2
```
dmlap: linebreak on parens

src/decrypter.js
BrandonOCasey: I think changing it makes sense. Only precompute once, and only do it when/if we need to. Any other time AES is called we will use the `tables` that has already been precomputed.

src/decrypter.js
```diff
@@ -266,16 +313,18 @@ decrypt = function(encrypted, key, initVector) {
...
-                     encrypted3,
-                     decrypted32,
-                     wordIx);
+    decipher.decrypt(
+      encrypted0,
+      encrypted1,
+      encrypted2,
+      encrypted3,
+      decrypted32,
+      wordIx
```
dmlap: linebreak on parens

src/decrypter.js
```diff
@@ -290,75 +339,82 @@ decrypt = function(encrypted, key, initVector) {
...
-  this.jobs.push(job);
-  if (!this.timeout_) {
-    this.timeout_ = setTimeout(this.processJob_.bind(this),
-                               this.delay);
+  processJob_() {
+    this.jobs.shift()();
+    if (this.jobs.length) {
+      this.timeout_ = setTimeout(
+        this.processJob_.bind(this),
+        this.delay
```
dmlap: linebreak on parens

src/decrypter.js
```diff
@@ -290,75 +339,82 @@ decrypt = function(encrypted, key, initVector) {
...
+      this.timeout_ = null;
+    }
   }
-};
+  push(job) {
+    this.jobs.push(job);
+    if (!this.timeout_) {
+      this.timeout_ = setTimeout(
+        this.processJob_.bind(this),
+        this.delay
```
dmlap: linebreak on parens (sorry!)

BrandonOCasey: haha its my own fault :)

src/decrypter.js
```diff
@@ -290,75 +339,82 @@ decrypt = function(encrypted, key, initVector) {
...
-  for (i = step; i < encrypted32.length; i += step) {
-    initVector = new Uint32Array([
-      ntoh(encrypted32[i - 4]),
-      ntoh(encrypted32[i - 3]),
-      ntoh(encrypted32[i - 2]),
-      ntoh(encrypted32[i - 1])
-    ]);
+export class Decrypter extends Stream {
+  constructor(encrypted, key, initVector, done) {
+    super(Stream);
```
dmlap: Just noticing the Decrypter never invoked the Stream constructor. Seems like that inheritance is unnecessary.

src/decrypter.js
```diff
@@ -290,75 +339,82 @@ decrypt = function(encrypted, key, initVector) {
...
+      this.timeout_ = setTimeout(
+        this.processJob_.bind(this),
+        this.delay
+      );
+    }
+  }
+}
+
+// the maximum number of bytes to process at one time
+const decrypterStep = 4 * 8000;
```
dmlap: Should this be capitalized to indicate it's a constant?

src/decrypter.js
```diff
@@ -290,75 +339,82 @@ decrypt = function(encrypted, key, initVector) {
...
+      decrypted.set(bytes, encrypted.byteOffset);
+    };
+  }
+}
 
-// exports
-videojs.Hls.decrypt = decrypt;
-videojs.Hls.Decrypter = Decrypter;
-videojs.Hls.AsyncStream = AsyncStream;
+Decrypter.STEP = decrypterStep;
```
dmlap: I would guess it's ok to define the constant value here instead of using the intermediate constant. Did that not work?

src/decrypter.js
BrandonOCasey: yes

test/decrypter.test.js
```diff
@@ -106,62 +87,68 @@ test('executes callback in series', function() {
...
+  // the string "howdy folks" encrypted
+  let encrypted = new Uint8Array([
+    0xce, 0x90, 0x97, 0xd0,
+    0x08, 0x46, 0x4d, 0x18,
+    0x4f, 0xae, 0x01, 0x1c,
+    0x82, 0xa8, 0xf0, 0x67
+  ]);
+  let decrypted;
+  let decrypter = new Decrypter(
+    encrypted,
```
dmlap: linebreak on parens

test/decrypter.test.js
```diff
@@ -106,62 +87,68 @@ test('executes callback in series', function() {
...
-                                            });
-  clock.tick(decrypter.asyncStream_.delay * 2);
-  ok(!done, 'not finished after two ticks');
-
-  clock.tick(decrypter.asyncStream_.delay);
-  ok(done, 'finished after the last chunk is decrypted');
+QUnit.test('breaks up input greater than the step value', function() {
+  let encrypted = new Int32Array(Decrypter.STEP + 4);
+  let done = false;
+  let decrypter = new Decrypter(
```
dmlap: linebreak parens

src/decrypter.js
dmlap: The deep copy might be kind of a pain but if you can find an easy way to do it, sweet.

src/decrypter.js
```diff
@@ -52,114 +50,138 @@ ntoh = function(word) {
...
+ * Expand the S-box tables.
  *
- * @constructor
- * @param key {Array} The key as an array of 4, 6 or 8 words.
+ * @private
  */
-AES = function (key) {
-  this._precompute();
+const precompute = function() {
+  let _tables = [[[], [], [], [], []], [[], [], [], [], []]];
```
dmlap: We normally use `_` to indicate a method or field is private. Is there a reason for the prefix here?

src/decrypter.js
```diff
@@ -52,114 +50,138 @@ ntoh = function(word) {
...
+    *
+    * The first 4 sub-tables are the expanded S-box with MixColumns. The
+    * last (_tables[01][4]) is the S-box itself.
+    *
+    * @private
+    */
+    if (!aesTables) {
+      aesTables = precompute();
+    }
+    this._tables = JSON.parse(JSON.stringify(aesTables));
```
dmlap: Fancy :) Does this actually save us anything over recomputing on each construction?

src/decrypter.js
```diff
@@ -290,75 +341,77 @@ decrypt = function(encrypted, key, initVector) {
...
+    this.jobs.push(job);
+    if (!this.timeout_) {
+      this.timeout_ = setTimeout(this.processJob_.bind(this),
+                                 this.delay);
+    }
+  }
+}
+
+// the maximum number of bytes to process at one time
+const DecrypterStep = 4 * 8000;
```
dmlap: `DECRYPTER_STEP`. Also, can this just be declared and defined as `Decrypter.STEP`?

src/decrypter.js
BrandonOCasey: yeah it should be much faster

src/decrypter.js
BrandonOCasey: well we made the function local to the module in this case, and we don't export it so essentially it is private now. My gut feeling on the prefix with _ rather than postfix _ was probably from a copy paste.

gkatsev: if it's just a local variable and doesn't get exported, it doesn't need any prefix or postfix indicating that it is private. Variables are inherently private by default.

src/decrypter.js
```diff
@@ -58,108 +59,128 @@ ntoh = function(word) {
...
    * Expand the S-box tables.
    *
    * @private
    */
-  _precompute: function () {
-   var encTable = this._tables[0], decTable = this._tables[1],
-       sbox = encTable[4], sboxInv = decTable[4],
-       i, x, xInv, d=[], th=[], x2, x4, x8, s, tEnc, tDec;
+  _precompute() {
+    let _tables = [[[], [], [], [], []], [[], [], [], [], []]];
```
dmlap: Sorry to do this... local variables don't need to be prefixed with `_`

PR #541, opened 2016-02-04
splitting up decrypter.js into multiple sub files

BrandonOCasey:


src/decrypter/async-stream.js
```diff
@@ -0,0 +1,27 @@
+import Stream from '../stream';
```
dmlap: This file needs a comment describing its contents.

dmlap:
My goals for modules in this project are:

- They are short enough you can read through them easily. 500 lines or so seems like a good target.
- They are not so short it is a pain to figure out how things fit together. Less than 20 lines seems like a warning sign to me.
- They export functionality that is well-defined and useful stand-alone.
- The functionality they export is actually used in the app.

The decrypt function and object, and the ntoh function don't to meet those criteria for me. AsyncStream is designed to be generic enough that it *might* be useful outside the decryption process so extracting that is fine by me. The other are not useful enough independently to be worth extracting and the combined module is still reasonably sized.

src/decrypter/decrypter.js
```diff
@@ -0,0 +1,144 @@
+import AES from './aes';
```
dmlap: Needs a file-level comment.

src/decrypter/async-stream.js
BrandonOCasey: added

src/decrypter/decrypter.js
BrandonOCasey: added

PR #543, opened 2016-02-05
only run precompute for tables once, slice copy after that

BrandonOCasey:
* move precompute function so that it is private to the module
* precompute only aesTables only once
* slice copy aesTables every other time
* JsPerf Comparison of our options: http://jsperf.com/aes-128-s-box-generation/7


dmlap:
Smoov.

PR #547, opened 2016-02-09
Duration and endOfStream fixes

imbcmdth:
Fixes `duration` reporting issues and corrects a bug in`updateEndHandler_` where we weren't always calling `mediaSource.endOfStream` when we should have been.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -297,6 +297,77 @@ videojs.Hls.findSoleUncommonTimeRangesEnd_ = function(original, update) {
   return result[0];
 };
 
+/**
+ * Updates segment with information about it's end-point in time and, optionally,
+ * the segment duration if we have enough information to determine a segment duration
+ * accurately.
+ * @param playlist {object} a media playlist object
+ * @param segmentIndex {number} the index of segment we last appended
+ * @param segmentEnd {number} the known of the segment referenced by segmentIndex
+ */
+videojs.HlsHandler.prototype.updateSegmentMetadata_ = function(playlist, segmentIndex, segmentEnd) {
+  var
+    segment,
+    previousSegment;
+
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -297,6 +297,72 @@ videojs.Hls.findSoleUncommonTimeRangesEnd_ = function(original, update) {
...
+  if (segmentEnd && segment) {
+    segment.end = segmentEnd;
+
+    // fix up segment durations based on segment end data
+    if (!previousSegment) {
+      // first segment is always has a start time of 0 making its duration
+      // equal to the segment end
+      segment.duration = segment.end;
+    } else if (previousSegment.end) {
+      segment.duration = segment.end - previousSegment.end;
```
dmlap: I'm still a *little* hesitant about overwriting duration during playback but let's see how it goes.

src/videojs-hls.js
```diff
@@ -297,6 +297,72 @@ videojs.Hls.findSoleUncommonTimeRangesEnd_ = function(original, update) {
...
+  // to read
+  appendedLastSegment = (segmentIndex === segments.length - 1);
+  bufferedToEnd = (currentBuffered.length &&
+    segments[segments.length - 1].end <= currentBuffered.end(0));
+
+  // if we've buffered to the end of the video, we need to call endOfStream
+  // so that MediaSources can trigger the `ended` event when it runs out of
+  // buffered data instead of waiting for me
+  return playlist.endList &&
+    this.mediaSource.readyState === 'open' &&
```
dmlap: Seems kind of weird that this function is checking `this.mediaSource`, given it's totally stateless outside this expression.

dmlap:
Two tiny questions but no blockers. LGTM!

PR #549, opened 2016-02-10
Fix buffer trimming so that it doesn't use seekable.start if it has a…

imbcmdth:
… value greater than the currentTime

We had a bug previously where if the seekable range was [10, 20] but currentTime was, say, 8 seconds we would remove up to seekable.start (in this case 10) and remove buffer we haven't even played yet. This would stall playback.


dmlap:
@gesinger: this would need to get migrated into the segment-loader PR right?

gesinger:
Yeah. I think all of the buffer trimming will need to be migrated in (I believe it might've gotten lost somewhere).

dmlap:
LGTM

PR #551, opened 2016-02-12
Avoid Firefox issue 548397

dmlap:
[getComputedStyle returns null](https://bugzilla.mozilla.org/show_bug.cgi?id=548397) when in an iframe with display: none so be careful dereferencing it.

### DIFF
--- a/test/videojs-hls_test.js
+++ b/test/videojs-hls_test.js
@@ -3106,6 +3106,26 @@ test('does not process update end until buffered value has been set', function()
   ok(!player.tech_.hls.pendingSegment_, 'pending segment cleared out');
 });
 
+// workaround https://bugzilla.mozilla.org/show_bug.cgi?id=548397
+test('selectPlaylist does not fail if getComputedStyle returns null', function() {
+  var oldGetComputedStyle = window.getComputedStyle;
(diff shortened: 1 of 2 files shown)


incompl:
LGTM

test/videojs-hls_test.js
```diff
@@ -3106,6 +3106,26 @@ test('does not process update end until buffered value has been set', function()
...
+  player.src({
+    src: 'master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(player);
+  standardXHRResponse(requests.shift()); // master
+  standardXHRResponse(requests.shift()); // media
+
+  player.tech_.hls.selectPlaylist();
+  ok(true, 'should not throw');
```
gesinger: Is this preferred over leaving a comment above selectPlaylist that an error should not be thrown?

gkatsev: qunit requires an assertion.

gesinger: Ah, gotcha. Thanks.

dmlap:
Weird Firefox test failure in Travis. Investigating.

PR #557, opened 2016-02-16
Change playlist loader to a class

BrandonOCasey:
remove internal loader variable
Change loader functions to class functions
Changed the code to use the es6 arrow


BrandonOCasey:
see #689

PR #558, opened 2016-02-16
browserify-p5: videojs-contrib-hls and bin-utils conversion

BrandonOCasey:
removed stub test and stub src files
updated build scripts to continue working
fixed several linting issues that were previously unnoticed
turn the linter back on
remove init function from stream
added ignore for playlist-loader as it will be finished later

> will not work without https://github.com/videojs/videojs-contrib-media-sources/pull/66


test/sanity.test.js
```diff
@@ -29,9 +29,11 @@ QUnit.test('the environment is sane', function(assert) {
   assert.strictEqual(typeof Array.isArray, 'function', 'es5 exists');
   assert.strictEqual(typeof sinon, 'object', 'sinon exists');
   assert.strictEqual(typeof videojs, 'function', 'videojs exists');
-  assert.strictEqual(typeof videojs.MediaSource, 'object', 'MediaSource is an object');
+  assert.strictEqual(typeof videojs.MediaSource, 'function', 'MediaSource is an object');
```
dmlap: > 'MediaSource is a __function__'

dmlap:
@BrandonOCasey: I'm still trying to work my way through the test cases. I noticed however, that you changed the order of operations in the afterEach() function of videojs-contrib-hls.test.js. This PR should not include any changes that would have required a re-ordering so I have to conclude that we've subtly changed behavior somewhere in videojs-hls.js. Also, you added a bunch of additional clock.tick()s to "selects the correct rendition by player dimensions". This also indicates a change in functionality.

dmlap:
I went ahead and made a number of tiny fixes as I've been reviewing this PR. You can review them in [my repo](https://github.com/dmlap/videojs-contrib-hls/tree/BrandonOCasey-browserify-p5).

BrandonOCasey:
@dmlap I will look into putting the order back to the way that it was or finding out why these things changed

PR #560, opened 2016-02-17
More comprehensive index/test page

BrandonOCasey:
added tech mode selection
added player options to form
added url content type to form
create player dynamically so that we can change options as we go

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,6 +21,10 @@
       min-width: 450px;
       padding: 5px;
     }
+    .video-container {
+      float: left;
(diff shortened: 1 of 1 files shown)


dmlap:
My thinking was that we would try to keep index.html relatively simple so it could serve as an example integration for people unfamiliar with how to use the project and we'd put the "smarts" into stats/index.html. What do you think about moving these additional options over there? We could link to stats from index.html so it's a bit easier for people to find, if that helps.

Issue #555, opened 2016-02-15
HLS videos do not always fire the 'ended' event


cdhowie:
This appears to be the same issue as #111.  I have had this issue with 1.3.5 and 1.3.7 (have not tested any other versions).  Sometimes the `ended` event does not fire when an HLS video ends, which causes major issues with our integration.

This only seems to be an issue on Chrome currently. I have tested on Firefox and found no issues.

ravi155:
+1

svenspielvogel:
i was able to fix it in https://github.com/videojs/videojs-contrib-hls/pull/561. Any better ideas?

FDiskas:
+1

cdhowie:
The v2.0.0 release has fixed this problem for me, but I don't know about others.

cdhowie:
I take it back; there are still some videos where the ended event doesn't fire, even with v2.0.0.

FDiskas:
what if dynamically #EXT-ENDLIST was added to playlist?
For me it looks like hls object is not updated correctly.
I don't have anything specific to this, but just the gues...

elmehdielkettani:
I have investigated the issue and noticed that it happens with a number of my streams. The common thread between the streams is that the last segment is very small (100ms or so). I hope that helps.

gesinger:
Sorry this ticket has gone a long time without much attention. We've made a lot of changes to videojs-contrib-hls in the meantime. If it's still a problem for anyone, would you be able to re-open the issue and provide a reduced test case using the latest version via [this template](http://jsbin.com/vokipos/8/edit?html,output)?

Thank you for reporting the issue, and for all of the added info.

Issue #550, opened 2016-02-11
Not working hls on server wowza


ROKAISAKKON:
Video.js 5.7.0
videojs-contrib-hls - v1.3.7
Chrome 48.0.2564.109
Wowza Streaming Engine 4 Monthly Edition 4.3.0 build16025

playlist.m3u8

```

#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM
INF:BANDWIDTH=1035985,CODECS="avc1.66.30,mp4a.40.2",RESOLUTION=480x320
chunklist_w2044689982.m3u8

```

chunklist_w2044689982.m3u8

```

#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:0
#EXTINF:9.6,
media_w2044689982_0.ts
#EXTINF:9.6,
media_w2044689982_1.ts
#EXTINF:9.6,
media_w2044689982_2.ts
...
#EXTINF:9.6,
media_w2044689982_92.ts
#EXTINF:7.019,
media_w2044689982_93.ts
#EXT-X-ENDLIST

```

.ts files stop downloading but new .ts files keep coming

url test : http://122.155.184.12/video_stream/_definst_/OTV2/lakorn/48_d_one_tv/02_wiman/ep3/03_wiman_part1_320.mp4/playlist.m3u8

srorry, i'm not good english

svenspielvogel:
I addressed this in my fork. The problem is, that wowza hls chunks seem to be too short, so the buffer misses some milliseconds. this results in a loop.
see https://github.com/mevango/videojs-contrib-hls

svenspielvogel:
see https://github.com/videojs/videojs-contrib-hls/pull/561

gak2:
Sven, I am experiencing exactly these issues live-streaming from Wowza: the browser seems to lose track of .ts files and just loops, loading the .m3u8 file.

I am very hopeful that your script might help, but - apologies for my inexperience - I'm have trouble applying it!  I am running 'video.js' from ZenCDN, and v.2.0.0 of 'videojs-contrib-hls.js' locally - all on a Windows server.  Could you please help me test your changes?  

I've tried replacing the latter file with yours from the 'src' folder on your fork, but it doesn't work.  Do I need more files from your fork?  Is non-Windows compilation required?  Even a working example would be great, just so I could follow the source code!

Thank you Sven ..

[...]

PR #561, opened 2016-02-19
Fix for #550 and #555

svenspielvogel:
this is a fix that resolves the wowza and the ended bug. Maybe there is a better solution, but it works fine.


dmlap:
Thanks for sharing your patch! It looks like there's a couple different tweaks going on in there. Would you mind explaining the extra behaviors you added in and the scenarios where they're helpful?

svenspielvogel:
Hey David. The problem is, that in some cases, especially when the video itself might be missing some audio frames the Wowza tends to pass the streams damaged as they are. This might lead to a gap between segments. What does that mean? Let's say Segment 1 Ends at Timestamp 5.300 - then it might happen, that the next segments starts at 5.350. This results in two timerange object. Or, what i discovered yesterday to even more TimeRanges. 
1. Problem: Video stops playing since there is a gap
Solve: if next timerange isnt far away (<0.5secs) jump over gap
2. Problem: Video loads segments in a loop: e.g. media_0.ts >> media_0.ts and so on.
Solve: Find real buffered end and get media index - my approach might still lead to a loop if we have 3 time ranges, but are still playing in the first time range. I'm gonna fix this pretty soon. I've to iterate through all timeranges. another solution, maybe the better one would be not to create a second timerange. Actually I wasn't able to figure it out in that short time I had, so this might be a good way. Any ideas on that?
3. The Plug-In doesn't always throw correct errors. e.g. loosing network or poor network (GPRS 64kBit/s)
Solve: setting up a retry count and throw error if it occurs.

svenspielvogel:
Another this. Travis fails. But this is because of 9d075a3. At this commit i set up my own 404 handling. Because the plugin throws an instant error even if enough buffer is present.
Just imagine followin use case: You sit in a Metro and at Station 1 you have 4G so you are buffering. Then you drive further and suddenly network falls away. Then, just 5 seconds later you get 4G again.

What happens at the old build and the current Tests: Even with 30 seconds in Buffer a miss on loading a segments leads to a error.
My Approach lets you still play the stream as log as you have buffer. (I love this behavior from the Amazon Instant Video Player). And as soon as you have Network the same segment gets buffered. After stalling too long, i will still throw a error 2.

nickygerritsen:
We seem to also have this issue, but (manually) applying the patch did not really solve it... 

I do get the following console errors after a short while:

```
VIDEOJS: WARN: Problem encountered with the current HLS playlist. No suitable alternatives found.
video.js:19152 VIDEOJS: ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported. MediaError {code: 4, message: "The media could not be loaded, either because the …rk failed or because the format is not supported."}_logType @ video.js:19152log.error @ video.js:19103error @ video.js:11654handleTechError_ @ video.js:10459ret @ video.js:18981ret @ video.js:18981data.dispatcher @ video.js:18629
video.js:19152 VIDEOJS: WARN: Problem encountered with the current HLS playlist. No suitable alternatives found.
video.js:19152 VIDEOJS: WARN: Problem encountered with the current HLS playlist. No suitable alternatives found.
17transmuxer-worker.js:49 Uncaught InvalidStateError: Failed to set the 'duration' property on 'MediaSource': The MediaSource's readyState is not 'open'.
```

Which I didn't get before.

svenspielvogel:
Can you provide the Video. I'd like to have a look into it. @nickygerritsen

nickygerritsen:
No problem, it is: http://hls.streamonecloud.net/livestream/amlst:h3kXcf2yyam/playlist.m3u8

cdhowie:
v2.0.0 seems to fix the "ended" issue for me (#555).  I'm not sure why, but it does.

cdhowie:
I take it back; there are still some videos where the ended event doesn't fire, even with v2.0.0.

forbesjo:
We've added a lot of fixes related to gaps, is this still an issue when using the latest version?

TylerZubke:
This is still happening for me. See #834. 
I tried @svenspielvogel fixes and the stream freezes with no error or anything indicating what happened. It also seems to happen more frequently than what occurs in my issue.

gesinger:
Thanks again for the PR @svenspielvogel . As to your comments above regarding what this fixes, we actually have made a lot of changes around those very areas (since we found them problematic as well). In order:

```
Problem: Video stops playing since there is a gap
Solve: if next timerange isnt far away (<0.5secs) jump over gap
```
We do a few different things now. In mux.js, we will fill some audio gaps with silent frames ( https://github.com/videojs/mux.js/pull/130 and https://github.com/videojs/mux.js/pull/143 ). In videojs-contrib-hls, we have https://github.com/videojs/videojs-contrib-hls/blob/master/src/playback-watcher.js which will skip over gaps in known error conditions.

```
Problem: Video loads segments in a loop: e.g. media_0.ts >> media_0.ts and so on.
Solve: Find real buffered end and get media index - my approach might still lead to a loop if we have 3 time ranges, but are still playing in the first time range. I'm gonna fix this pretty soon. I've to iterate through all timeranges. another solution, maybe the better one would be not to create a second timerange. Actually I wasn't able to figure it out in that short time I had, so this might be a good way. Any ideas on that?
```
We now have much better fetching behavior, and it has been resilient against looping segment requests. Please give it a shot and let us know if you see otherwise!

```
The Plug-In doesn't always throw correct errors. e.g. loosing network or poor network (GPRS 64kBit/s)
Solve: setting up a retry count and throw error if it occurs.
```
We've been working on better error reporting/events, and now will retry the last stream if it fails https://github.com/videojs/videojs-contrib-hls/pull/1039 .

As for the case @TylerZubke posted, it looks like that was solved.

Thanks again for the PR, and please let us know if you're still experiencing issues with the latest version of videojs-contrib-hls (as of this comment, v5.5.0). Thank you.

PR #573, opened 2016-02-29
Work around a bug with the buffered property in IE11 that prevented playback

imbcmdth:


dmlap:
Is it safe to assume findBufferedRanges_() is the only place this fix will be required?

dmlap:
Assuming your answer is "basically, yes", LGTM

PR #576, opened 2016-03-01
Work around a bug with the buffered property in IE11 that prevented playback

imbcmdth:
Same fix as #573 but for the `master` branch.


gkatsev:
LGTM

PR #580, opened 2016-03-03
Added an extra null check for mediaSource

forbesjo:
Added an extra null check for mediaSource, it looks like `this.mediaSource.mediaSource_` doesn't exist on an MSE-less browser causing the script to error out.

### DIFF
--- a/src/videojs-hls.js
+++ b/src/videojs-hls.js
@@ -892,6 +892,7 @@ var filterBufferedRanges = function(predicate) {
     // report a fully empty buffer until SourceBuffers have been created
     // which is after a segment has been loaded and transmuxed.
     if (!this.mediaSource ||
+        !this.mediaSource.mediaSource_ ||
         !this.mediaSource.mediaSource_.sourceBuffers.length) {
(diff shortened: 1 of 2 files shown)


src/videojs-hls.js
```diff
@@ -892,6 +892,7 @@ var filterBufferedRanges = function(predicate) {
     // report a fully empty buffer until SourceBuffers have been created
     // which is after a segment has been loaded and transmuxed.
     if (!this.mediaSource ||
+        !this.mediaSource.mediaSource_ ||
```
dmlap: I don't think we should be looking at `this.mediaSource.mediaSource_` in this project at all (which I know the next line is violating). Can this be fixed in contrib-mse?

forbesjo: I could check `videojs.MediaSource.supportsNativeMediaSources()` or do we want to remove the next line as well?

forbesjo: I've added a story to our backlog to come back to this code later and figure out a way to do this without accessing private properties.

imbcmdth:
LGTM

PR #581, opened 2016-03-04
Added a null check

forbesjo:
Same as #580


dmlap:
Should we fix this one in a new release of contrib-mse?

PR #585, opened 2016-03-07
Segment Loader, on development

dmlap:
Pull out segment fetching logic into it's own object (again...).


test/plugin-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+    this.sourceBuffers.push(sourceBuffer);
+    return sourceBuffer;
+  };
+
+  return mediaSource;
+};
+
+// patch over some methods of the provided tech so it can be tested
+// synchronously with sinon's fake timers
+let Helper = {
```
dmlap: This looks copy-pasted from the main test file (and it looks like we didn't get rid of it back there).

test/plugin-helpers.js
```diff
@@ -0,0 +1,267 @@
...
+    this.updates_.push({
+      remove: [start, end]
+    });
+    this.updating = true;
+  },
+
+  updating: false
+});
+
+export const MockMediaSource = videojs.extend(videojs.EventTarget, {
```
dmlap: This appears to be unused.

dmlap:
Looks like retrying key requests didn't make the move into the segment loader. I don't think it's a big deal now that we have variant blacklisting so I'm removing the test cases.

dmlap:
Two additional issues discovered:
- Source Updaters hard-code the underlying SourceBuffer's MIME type to `video/mp4`
- Source Updaters create their SourceBuffer as soon as their MediaSource opens, which may be before the media playlist has provided the codec configuration

PR #587, opened 2016-03-09
Development rebased onto segment loader

dmlap:
Experimenting to see if this is easier than rebasing segment loader on top of development.


dmlap:
This was interesting as a double-check on the other branch that brought over segment loader but won't be merged.

PR #599, opened 2016-03-14
[WIP] Master Playlist Controller + Support for Alternate Audio Track Loading

gesinger:
The first commit is the Master Playlist Controller (original PR: https://github.com/videojs/videojs-contrib-hls/pull/591 ).

The third commit includes changes to support alternate audio track loading.

CODE MOVES:
- A lot of moves were done from videojs-contrib-hls to master-playlist-controller to better confine the logic of the main plugin file (dealing with videojs interfaces), and the HLS specific logic.
- Some old unused code was removed (including sourceBuffer handling and creation within videojs-contrib-hls, as that is now the responsibility of each segment loader's source updater).

TODOs (included in code):
- loadAlternateAudioPlaylist is unused, and should be used when we start listening to change events on AudioTracksList
- within loadAlternateAudioPlaylists, on error events, there are TODOs to re-enable the main's audio
- Some methods within videojs-contrib-hls are unused within the project, and should either be removed or used where appropriate (TODOs are labeled above excludeIncompatibleVariants, playlistUriToUrl, and setBandwidth).

NOTES:
- We're not listening to audio segment loader progress events. This is because we do not need to switch renditions, and don't gain much by keeping an eye on the bandwidth (instead we rely on the bandwidth estimates from the combined video/audio stream).
- We're not pausing the alternate audio segment loading on media changes, as we might as well continue loading the same audio track.
- We're only tracking expired time for the main stream.

There are currently no tests, and old tests were not moved/fixed. These need to be added.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1,4 +1,3 @@
-import Stream from './stream';
 import PlaylistLoader from './playlist-loader';
 import SegmentLoader from './segment-loader';
 import Ranges from './ranges';
@@ -155,7 +154,7 @@ const selectPlaylist = function() {
     sortedPlaylists[0];
 };
 
-export default class MasterPlaylistController extends Stream {
+export default class MasterPlaylistController extends videojs.EventTarget {
   constructor({url, withCredentials, currentTimeFunc, mediaSourceMode, hlsHandler,
     externHls}) {
     super();
@@ -221,8 +220,8 @@ export default class MasterPlaylistController extends Stream {
       this.updateDuration(this.masterPlaylistLoader_.media());
 
       // update seekable
-      seekable = this.hlsHandler.seekable();
-      if (this.hlsHandler.duration() === Infinity && seekable.length !== 0) {
+      seekable = this.seekable();
+      if (!updatedPlaylist.endList && seekable.length !== 0) {
         this.mediaSource.addSeekableRange_(seekable.start(0), seekable.end(0));
       }
     });
@@ -318,8 +317,8 @@ export default class MasterPlaylistController extends Stream {
 
     // check that everything is ready to begin buffering
 
-    // 1) the video is a live stream of unknown duration
-    if (this.hlsHandler.duration() === Infinity &&
+    // 1) the video is a live stream
+    if (!media.endList &&
 
         // 2) the player has not played before and is not paused
         this.hlsHandler.tech_.played().length === 0 &&
@@ -338,7 +337,7 @@ export default class MasterPlaylistController extends Stream {
         this.masterPlaylistLoader_.trigger('firstplay');
 
         // seek to the latest media position for live videos
-        seekable = this.hlsHandler.seekable();
+        seekable = this.seekable();
         if (seekable.length) {
           this.hlsHandler.tech_.setCurrentTime(seekable.end(0));
         }
@@ -396,7 +395,7 @@ export default class MasterPlaylistController extends Stream {
     return this.mediaSource.endOfStream('network');
   }
 
-  pause() {
+  pauseLoading() {
     this.mainSegmentLoader_.pause();
     if (this.audioPlaylistLoader_) {
       this.audioSegmentLoader_.pause();
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -211,12 +211,17 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       throw new Error('A non-empty playlist URL is required');
     }
 
-    this.masterPlaylistLoader_ = new PlaylistLoader(url, this.withCredentials);
+    this.masterPlaylistLoader_ = new PlaylistLoader({
+      srcUrl: url,
+      withCredentials: this.withCredentials,
+      startLoading: true
+    });
     this.hlsHandler.playlists = this.masterPlaylistLoader_;
 
     this.masterPlaylistLoader_.on('loadedmetadata', () => {
       let media = this.masterPlaylistLoader_.media();
       let master = this.masterPlaylistLoader_.master;
+      let alternateAudioUri;
 
       // if this isn't a live video and preload permits, start
       // downloading segments
@@ -227,14 +232,15 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         this.mainSegmentLoader_.load();
       }
 
-      this.audioPlaylistLoaders_ = [];
+      this.audioPlaylistLoaders_ = {};
       if (master.mediaGroups && master.mediaGroups.AUDIO) {
         for (let groupKey in master.mediaGroups.AUDIO) {
           for (let labelKey in master.mediaGroups.AUDIO[groupKey]) {
-            this.audioPlaylistLoaders_.push(new PlaylistLoader(
-              master.mediaGroups.AUDIO[groupKey][labelKey].uri,
-              this.withCredentials,
-              true));
+            alternateAudioUri = master.mediaGroups.AUDIO[groupKey][labelKey].uri;
+            this.audioPlaylistLoaders_[alternateAudioUri] = new PlaylistLoader({
+              srcUrl: alternateAudioUri,
+              withCredentials: this.withCredentials
+            });
           }
         }
       }
@@ -294,10 +300,14 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.audioSegmentLoader_.on('error', () => {
+      videojs.log.warn('Problem encountered with the current alternate audio track' +
+                       '. Switching back to default.');
       this.audioSegmentLoader_.abort();
       this.audioPlaylistLoader_ = null;
       this.useAudio('main');
     });
+
+    this.masterPlaylistLoader_.load();
   }
 
   load() {
@@ -325,16 +335,12 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     audioEntry =
       this.masterPlaylistLoader_.master.mediaGroups.AUDIO[mediaGroupName][label];
 
-    this.audioPlaylistLoaders_.forEach((audioPlaylistLoader) => {
-      if (audioPlaylistLoader.srcUrl === audioEntry.uri) {
-        newAudioPlaylistLoader = audioPlaylistLoader;
-      }
-    });
+    newAudioPlaylistLoader = this.audioPlaylistLoaders_[audioEntry.uri];
 
     if (!newAudioPlaylistLoader.started) {
       this.loadAlternateAudioPlaylistLoader_(newAudioPlaylistLoader);
     } else {
-      newAudioPlaylistLoader.resume();
+      newAudioPlaylistLoader.load();
       this.audioSegmentLoader_.load();
     }
   }
@@ -376,6 +382,8 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.audioPlaylistLoader_.on('error', () => {
+      videojs.log.warn('Problem encountered loading the alternate audio track' +
+                       '. Switching back to default.');
       this.audioSegmentLoader_.abort();
       this.audioPlaylistLoader_ = null;
       this.useAudio('main');
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -211,11 +211,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       throw new Error('A non-empty playlist URL is required');
     }
 
-    this.masterPlaylistLoader_ = new PlaylistLoader({
-      srcUrl: url,
-      withCredentials: this.withCredentials,
-      startLoading: true
-    });
+    this.masterPlaylistLoader_ = new PlaylistLoader(url, this.withCredentials);
     this.hlsHandler.playlists = this.masterPlaylistLoader_;
 
     this.masterPlaylistLoader_.on('loadedmetadata', () => {
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -216,6 +216,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
 
     this.masterPlaylistLoader_.on('loadedmetadata', () => {
       let media = this.masterPlaylistLoader_.media();
+      let master = this.masterPlaylistLoader_.master;
 
       // if this isn't a live video and preload permits, start
       // downloading segments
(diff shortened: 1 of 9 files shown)


dmlap:
>NOTES:

> - We're not listening to audio segment loader progress events. This is because we do not need to switch renditions, and don't gain much by keeping an eye on the bandwidth (instead we rely on the bandwidth estimates from the combined video/audio stream).
> - We're not pausing the alternate audio segment loading on media changes, as we might as well continue loading the same audio track.

Makes sense.

> - We're only tracking expired time for the main stream.

Seems like we might run into live-sync issues without per-stream expired tracking but we could layer that in later.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  // fallback chain of variants
+  return resolutionPlusOne ||
+    resolutionBestVariant ||
+    bandwidthBestVariant ||
+    sortedPlaylists[0];
+};
+
+export default class MasterPlaylistController extends Stream {
```
dmlap: It might be more appropriate for this prototype to extend videojs.EventTarget.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+  constructor({url, withCredentials, currentTimeFunc, mediaSourceMode, hlsHandler,
+    externHls}) {
```
dmlap: With this many constructor arguments, it's probably time to switch over to options-object style.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+    super();
+
+    Hls = externHls;
```
dmlap: We should be switching to using `import` for this.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+      withCredentials: this.withCredentials
+    });
+    // alternate audio track
+    this.audioSegmentLoader_ = new SegmentLoader({
+      mediaSource: this.mediaSource,
+      currentTime: this.currentTimeFunc,
+      withCredentials: this.withCredentials
+    });
+
+    this.hlsHandler = hlsHandler;
```
dmlap: One goal that I have for this refactor is to reduce bi-directional dependencies. Passing the HlsHandler along here seems very close to the dependency between techs and the player object that we just untangled with great effort in video.js 5. This isn't any worse than the status quo but we should think long and hard about how to avoid two complex objects that mutually and directly invoke one another. It seems possible to reformulate a lot of the interaction with HlsHandler as it driving MasterPlaylistController, and allow communication back to happen through events.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  load() {
+    this.mainSegmentLoader_.load();
+    if (this.audioPlaylistLoader_) {
+      this.audioSegmentLoader_.load();
+    }
+  }
+
+  loadAlternateAudioPlaylist(playlistLoader) {
```
dmlap: I think it would be cool if we could encapsulate PlaylistLoader creation in MasterPlaylistController. What about taking a playlist object or an index instead?

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+   * player and video are loaded and initialized.
+   */
+  setupFirstPlay() {
+    let seekable;
+    let media = this.masterPlaylistLoader_.media();
+
+    // check that everything is ready to begin buffering
+
+    // 1) the video is a live stream of unknown duration
+    if (this.hlsHandler.duration() === Infinity &&
```
dmlap: Many of the calls to `hlsHandler` in this method could be converted to queries against playlists or playlist loaders. `!this.mainPlaylistLoader_.media().endList` is equivalent to this comparison, for instance.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+      return this.masterPlaylistLoader_.media(nextPlaylist);
+    }
+    videojs.log.warn('Problem encountered with the current ' +
+                     'HLS playlist. No suitable alternatives found.');
+    // We have no more playlists we can select so we must fail
+    this.hlsHandler.error = error;
+    return this.mediaSource.endOfStream('network');
+  }
+
+  pause() {
```
dmlap: The name of this method might be confusing. You wouldn't want to pause the segment loaders if the player is paused.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  seekable() {
+    let media;
+    let seekable;
+
+    if (!this.masterPlaylistLoader_) {
+      return videojs.createTimeRanges();
+    }
+    media = this.masterPlaylistLoader_.media();
```
dmlap: The MasterPlaylistController will need to take all active loaders into account for this to work correctly.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,519 @@
...
+
+    // seekable has been calculated based on buffering video data so it
+    // can be returned directly
+    return seekable;
+  }
+
+  /**
+   * Update the player duration
+   */
+  updateDuration(playlist) {
```
dmlap: In putting the segment loader together, I came to the conclusion that we're doing this totally wrong :) I think we should set the duration when the first media playlist loads and then let media sources manage it from that point out. As it stands, I think we may be invoking the coded frame removal algorithm unnecessarily.

src/videojs-contrib-hls.js
```diff
@@ -207,124 +114,41 @@ export default class HlsHandler extends Component {
...
-      // with the updated bandwidth information
-      this.bandwidth = this.segments.bandwidth;
-      this.playlists.media(this.selectPlaylist());
-
-      this.tech_.trigger('progress');
-    });
-    this.segments.on('error', () => {
-      this.blacklistCurrentPlaylist_(this.segments.error());
-    });
+    this.tech_.one('canplay',
```
dmlap: `canplay` indicates [media data is available in the buffer](https://html.spec.whatwg.org/multipage/embedded-content.html#event-media-canplay). That should never happen until we begin buffering and we don't begin buffering live streams until you hit play. Seems like this condition shouldn't be hit.

src/videojs-contrib-hls.js
```diff
@@ -861,6 +329,70 @@ const HlsSourceHandler = function(mode) {
...
+/**
+ * A comparator function to sort two playlist object by bandwidth.
+ * @param left {object} a media playlist object
+ * @param right {object} a media playlist object
+ * @return {number} Greater than zero if the bandwidth attribute of
+ * left is greater than the corresponding attribute of right. Less
+ * than zero if the bandwidth of right is greater than left and
+ * exactly zero if the two are equal.
+ */
+Hls.comparePlaylistBandwidth = function(left, right) {
```
dmlap: Were these moved or is it just the diff?

dmlap:
Looks like a very promising start. My main suggestion is we work to be clear at  the outset the responsibilities of each part of the system so it's clear where functionality belongs. I'd suggest:

- MasterPlaylistController: very m3u8/HLS oriented properties and methods. Mostly focused with managing and aggregating playlists loaders.
- hls: very media-element oriented. Drives MasterPlaylistController and adapts m3u8/HLS events and methods to HTML events and methods.

src/master-playlist-controller.js
gesinger: I agree. I'll take a look at removing it.

src/master-playlist-controller.js
gesinger: Over destructuring?

src/master-playlist-controller.js
gesinger: We might be able to do a playlist object, but we'd have to have to have the reference to the playlist loader to deal with refreshes. Were you thinking that MasterPlaylistController can do the listening to the active playlist loader and appropriately update the playlist, and by the time we get to loadAlternateAudioPlaylist the MPC would already have started listening and doing the refreshes of the playlist?

src/master-playlist-controller.js
gesinger: Good call. Will change.

src/master-playlist-controller.js
gesinger: For now I figured we would be deferring duration/seekable/other timing related information to the video/combined track only. Should we do both to start (i.e., take the longest of both)?

src/master-playlist-controller.js
gesinger: Good call.

src/master-playlist-controller.js
gesinger: The main reason we are passing hlsHandler right now is to allow users to override what they previously could override. For instance, the selectPlaylist function. At the moment people can set that to be whatever function they want, and expect it to be called with the HLS Handler. We could pass the bound function down instead, which prevents MasterPlaylistController from having the entire hlsHandler reference, but it might not buy us too much in terms of separation, and if the user updates it again, we'd have to call back in to MPC and modify it (which is fine, we'd just have to agree on that interaction). Were you thinking of a different relationship?

src/master-playlist-controller.js
dmlap: Oh wow. I haven't been looking at es6 enough yet. Nevermind!

src/master-playlist-controller.js
dmlap: I'd want to keep the existing overrides around. I think your proposal for passing selectPlaylist() down and keeping it updated sounds like a good one.

src/master-playlist-controller.js
dmlap: I think so but let me try restating it just to verify we're on the same page. I was thinking the MPC would listen for refreshes and signal them to the segment loaders. When it's instructed to load alternate audio, it would start up an alt audio playlist loader and activate the appropriate segment loader once the alt audio playlist was ready. 

Since the MPC handles the entire playlist loader lifecycle, it doesn't have to expose them to hls/the media element adapter. HLS could signal the MPC to begin loading alternate audio by name or URI:
```js
// by name:
mpc.useAudio('Espanol');
// by index:
mpc.useAudio(2);
```

src/master-playlist-controller.js
dmlap: I think we'd need to do the intersection-- if one playlist is missing content we'd stall by seeking there.

src/master-playlist-controller.js
gesinger: Will try for that.

src/master-playlist-controller.js
gesinger: I see what you're saying now. It lets us encapsulate the selection logic (rather, the listening to selected audio tracks) within HLS, instead of MPC having to keep track of that as an external reference to the player's audio track list change events.

That makes a lot of sense, and I'll modify it to allow for that interaction.

src/master-playlist-controller.js
gesinger: That's a fair point. Though do you know if it's valid for an alternate audio track to have a longer or shorter duration than the video/combined track? I could definitely see the video stalling if it is pending information from one source, but if that is allowed it could also create other sorts of problems (from usage point of view, a shorter audio track could potentially cut off portions of video, and a longer audio track would be cut off as well).

dmlap: It's legal to add variants to a live stream while it's in progress which is one way this scenario could show up. If you're running a 24/7 news channel, you might decide to add a 4k variant to your stream while it's running and there would be less video available for that variant while it filled out the live window.

src/master-playlist-controller.js
gesinger: Turns out that in this case, using import will create a cyclical dependency (a sign of another problem, but that should go away over time).

src/master-playlist-controller.js
```diff
@@ -226,6 +227,17 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         this.mainSegmentLoader_.load();
       }
 
+      this.audioPlaylistLoaders_ = [];
+      if (master.mediaGroups && master.mediaGroups.AUDIO) {
+        for (let groupKey in master.mediaGroups.AUDIO) {
+          for (let labelKey in master.mediaGroups.AUDIO[groupKey]) {
+            this.audioPlaylistLoaders_.push(new PlaylistLoader(
+              master.mediaGroups.AUDIO[groupKey][labelKey].uri,
```
dmlap: For later: we will need to resolve these URIs

src/master-playlist-controller.js
```diff
@@ -280,6 +292,12 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     this.mainSegmentLoader_.on('error', () => {
       this.blacklistCurrentPlaylist_(this.mainSegmentLoader_.error());
     });
+
+    this.audioSegmentLoader_.on('error', () => {
+      this.audioSegmentLoader_.abort();
+      this.audioPlaylistLoader_ = null;
+      this.useAudio('main');
```
dmlap: Log a warning of some sort here

src/master-playlist-controller.js
```diff
@@ -226,6 +227,17 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         this.mainSegmentLoader_.load();
       }
 
+      this.audioPlaylistLoaders_ = [];
+      if (master.mediaGroups && master.mediaGroups.AUDIO) {
+        for (let groupKey in master.mediaGroups.AUDIO) {
+          for (let labelKey in master.mediaGroups.AUDIO[groupKey]) {
+            this.audioPlaylistLoaders_.push(new PlaylistLoader(
```
dmlap: I think it's worth considering making this an object so that you can index into them by URI below.

src/playlist-loader.js
```diff
@@ -102,7 +102,7 @@ const updateMaster = function(master, media) {
   return changed ? result : null;
 };
 
-const PlaylistLoader = function(srcUrl, withCredentials) {
+const PlaylistLoader = function(srcUrl, withCredentials, startsPaused) {
```
dmlap: I would rather we adopt the convention that loaders always start paused (a la segment loader) and provide a mechanism to begin loading explicitly. Having the two loader types share a basic lifecycle sounds like a nice simplification to me.

src/playlist-loader.js
```diff
@@ -338,76 +340,90 @@ const PlaylistLoader = function(srcUrl, withCredentials) {
...
-    let playlist;
-    let i;
+  loader.pause = () => {
+    loader.stopRequest();
+    window.clearTimeout(mediaUpdateTimeout);
+  };
 
-    // clear the loader's request reference
-    request = null;
+  loader.resume = () => {
```
dmlap: Similar to my comment above, what do you think about calling this `load()` to align with segment loader?

src/playlist-loader.js
```diff
@@ -338,76 +340,90 @@ const PlaylistLoader = function(srcUrl, withCredentials) {
...
+          // MEDIA_ERR_NETWORK
+          code: 2
+        };
+        return loader.trigger('error');
+      }
 
-    // loaded a master playlist
-    if (parser.manifest.playlists) {
-      loader.master = parser.manifest;
+      parser = new m3u8.Parser();
```
dmlap: Looks like the changes below are to remove the master-playlist fetching capabilities of this object and are not quite finished. Is that correct?

gesinger: I think that is just the diff. What I did here was just move the initial loading into the function loader.start and changed the indentation level.

src/master-playlist-controller.js
```diff
@@ -0,0 +1,674 @@
...
+        // select the initial variant
+        this.masterPlaylistLoader_.media(this.hlsHandler.selectPlaylist());
+        return;
+      }
+
+      this.mainSegmentLoader_.playlist(updatedPlaylist);
+      this.updateDuration(this.masterPlaylistLoader_.media());
+
+      // update seekable
+      seekable = this.seekable();
```
dmlap: seekable() is a media-element thing, so it seems like it should be calculated by hls.js

imbcmdth:
Superseded by #618

PR #601, opened 2016-03-16
Add Audio Tracks to video.js tracklist

BrandonOCasey:


test/videojs-contrib-hls.test.js
```diff
@@ -2072,6 +2061,54 @@ QUnit.test('resolves relative key URLs against the playlist', function() {
               'resolves the key URL');
 });
 
+QUnit.test('adds 1 empty audio track if we have not parsed any, and the playlist is loaded', function() {
```
gesinger: Is it actually empty?

src/videojs-contrib-hls.js
```diff
@@ -278,8 +279,42 @@ export default class HlsHandler extends Component {
...
       if (!updatedPlaylist) {
+        let selectedPlaylist = this.selectPlaylist();
+
+        if (!selectedPlaylist.attributes.AUDIO) {
+          // from safari defaults
+          this.tech_.audioTracks().addTrack(new AudioTrack({
+            enabled: true,
+            id: '1',
+            kind: 'main',
+            tech: this.tech_
```
dmlap: The tech is a required argument for new AudioTracks?

src/videojs-contrib-hls.js
```diff
@@ -278,8 +279,42 @@ export default class HlsHandler extends Component {
...
+          }));
+        } else {
+          let audioMediaGroupName = selectedPlaylist.attributes.AUDIO;
+          let mg = this.playlists.master.mediaGroups.AUDIO[audioMediaGroupName];
+
+          for (let label in mg) {
+            let language = mg[label].language || '';
+            /* eslint-disable dot-notation */
+            // we need to use non dot notation around defaul for ie8
```
dmlap: This project doesn't support IE8 so as long as this doesn't fail at parse-time, it's ok to skip IE8 compatibility stuff.

gesinger:
Maybe this is out of scope for now, but if a playlist is switched, should we clear out the alternate audio tracks and re-add them with the associated ones from the new playlist?

src/videojs-contrib-hls.js
```diff
@@ -278,8 +279,42 @@ export default class HlsHandler extends Component {
...
+
+        if (!selectedPlaylist.attributes.AUDIO) {
+          // from safari defaults
+          this.tech_.audioTracks().addTrack(new AudioTrack({
+            enabled: true,
+            id: '1',
+            kind: 'main',
+            tech: this.tech_
+          }));
+        } else {
```
dmlap: No action necessary but when the branches of conditionals get this big I tend to favor short-circuit `return`s so you don't have to remember what block you're in for so long while reading.

test/videojs-contrib-hls.test.js
BrandonOCasey: It will look like this: 
```JavaScript
  {
    language: '',
    label: '',
    enabled: true,
    id: '1',
    kind: 'main',
    tech: this.tech_
}));
``` JavaScript

File: src/videojs-contrib-hls.js [hunk 5]
```diff
@@ -309,7 +345,7 @@ export default class HlsHandler extends Component {
     });
 
     this.segments = new SegmentLoader({
-      currentTime: this.tech_.currentTime.bind(this.tech_),
+      currentTime: () => this.tech_.currentTime(),
```
dmlap: No objection to this change but was there a particular reason for it?

test/videojs-contrib-hls.test.js
```diff
@@ -315,25 +315,17 @@ QUnit.module('HLS', {
...
-  // this.player.tech_.triggerReady();
-  // this.clock.tick(1);
   // make sure play() is called *after* the media source opens
-  this.player.tech_.hls.play = function() {
-    plays++;
-  };
   openMediaSource(this.player, this.clock);
 
   standardXHRResponse(this.requests[0]);
-  QUnit.strictEqual(1, plays, 'play was called');
```
dmlap: This assertion still seems valid. Why remove it?

src/videojs-contrib-hls.js
gkatsev: `Function#bind` is an issue on pages with broken polyfills like prototypejs and mootools.

src/videojs-contrib-hls.js
BrandonOCasey: I think that we were having difficulties with webworkify not quoting their default string (something we have yet to come up with a workaround for)

test/videojs-contrib-hls.test.js
```diff
@@ -2072,6 +2061,54 @@ QUnit.test('resolves relative key URLs against the playlist', function() {
...
+  // master
+  standardXHRResponse(this.requests.shift());
+  let at = this.player.audioTracks();
+
+  QUnit.equal(at.length, 3, `three audio tracks after load`);
+
+  QUnit.equal(at[0].label, 'English', `track 1 - label = NAME`);
+  QUnit.equal(at[0].enabled, true, `track 1 - enabled = DEFAULT`);
+  QUnit.equal(at[0].language, 'eng', `track 1 - language = LANG`);
+  QUnit.equal(at[0].kind, 'main', `track 1 - kind = main if DEFAULT is YES`);
```
dmlap: Another non-critical comment: you can use `deepEqual` to cut down on the amount of typing you need to do for these sorts of tests.

src/videojs-contrib-hls.js
BrandonOCasey: yes

test/videojs-contrib-hls.test.js
BrandonOCasey: will deepEqual work with Object.defineProperties?

test/videojs-contrib-hls.test.js
dmlap: +1 for "default"

src/videojs-contrib-hls.js
dmlap: webworkify only runs agains mux.js so that shouldn't cause problems here.

test/videojs-contrib-hls.test.js
BrandonOCasey: I think that some of these unit test changes may have been from an old version of my-rebase, as I did not change any of them.

test/videojs-contrib-hls.test.js
dmlap: Ohhh... no idea. Would be fun to find out :)

src/videojs-contrib-hls.js
BrandonOCasey: well in our case we would also not be quoting a key called `default` and I don't think IE would like it but I can test to make sure.

src/videojs-contrib-hls.js
dmlap: It's not clear to me why it should be. If the tech owns the audioTracks and the audioTracks is called to add a track, it seems like the audio tracks could pass along the tech instead of requiring it from the caller.

gkatsev: Another question, do audioTracks and videoTracks actually need a reference to the tech inside them? TextTracks do because that's how they update activeCues but TextTracks are also not generally created by hand but via helper methods.
Maybe only TextTracks need to require techs?

test/videojs-contrib-hls.test.js
BrandonOCasey: yeah deepEquals will not work quite right as AudioTracks are an object with lots of functions and internal variables etc.

src/videojs-contrib-hls.js
BrandonOCasey: Yeah I don't think that AudioTracks or VideoTracks should actually require tech_

test/videojs-contrib-hls.test.js
gkatsev:
```js
QUnit.deepEquals(_.pick(at[0], ['label, 'enabled', 'language', 'kind']), {label: 'English', enabled: true, langauge: 'eng', kind: 'main'});
```
or something.

PR #612, opened 2016-03-23
dispose audio playlist rather than set null.

BrandonOCasey:
this fixes playlist switching
move audioTrack addition back into hls

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -295,7 +295,6 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         type: 'mediachange',
         bubbles: true
       });
-      this.trigger('mediachange');
     });
 
     this.mainSegmentLoader_.on('progress', () => {
@@ -315,7 +314,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       videojs.log.warn('Problem encountered with the current alternate audio track' +
                        '. Switching back to default.');
       this.audioSegmentLoader_.abort();
-      this.audioPlaylistLoader_.dispose();
+      this.audioPlaylistLoader_ = null;
       this.useAudio();
     });
 
@@ -337,7 +336,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     // Pause any alternative audio
     if (this.audioPlaylistLoader_) {
       this.audioPlaylistLoader_.pause();
-      this.audioPlaylistLoader_.dispose();
+      this.audioPlaylistLoader_ = null;
       this.audioSegmentLoader_.pause();
     }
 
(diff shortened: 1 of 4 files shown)


src/master-playlist-controller.js
```diff
@@ -370,7 +336,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     // Pause any alternative audio
     if (this.audioPlaylistLoader_) {
       this.audioPlaylistLoader_.pause();
-      this.audioPlaylistLoader_ = null;
+      this.audioPlaylistLoader_.dispose();
```
gesinger: The problem with using dispose here is that we reuse the alternate audio playlist loader the if it is selected in the future (we don't create them when they are selected).

It would also still need to be set to null for some other checks to work (it isn't the optimal approach, but right now we use if (this.audioPlaylistLoader_) to determine if we are using an alternate audio playlist.

src/videojs-contrib-hls.js
```diff
@@ -134,6 +135,39 @@ export default class HlsHandler extends Component {
...
+      for (let label in attributes.audio) {
+        let hlstrack = attributes.audio[label];
+        let kind = 'alternative';
+        let enabled = hlstrack.default || false;
+        let language = hlstrack.language || '';
+        if (enabled) {
+          kind = 'main';
+        }
+
+        let track = new AudioTrack({
```
gesinger: This might just be a style preference (so feel free to ignore), but I think it would look cleaner (and save variables) as:

```javascript
audioTracks.addTrack(new AudioTrack({
  kind: hlstrack.default ? 'main' : 'alternative',
  enabled: hlstrack.default || false,
  language: hlstrack.language || '',
  label,
});
```

src/master-playlist-controller.js
BrandonOCasey: the problem comes in when we set an `audioPlaylistLoader_` to null. which sets an entry in `audioPlaylistLoaders_` to null. Then when we go to switch back to the audio a second time there is no `PlaylistLoader` to use for that track. Maybe we could not create all the playlist loaders at the start? and only create them when we use them?

src/master-playlist-controller.js
```diff
@@ -329,6 +294,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         type: 'mediachange',
         bubbles: true
       });
+      this.trigger('mediachange');
```
gesinger: Is anyone listening to mediachange?

BrandonOCasey: at first I thought we could watch for it in hls and then update video.js audiotracks. but it turned out that that was the wrong event. I will remove this line good catch.

src/master-playlist-controller.js
gesinger: Isn't audioPlaylistLoader_ a separate reference? Setting it to null shouldn't modify the entry in audioPlaylistLoaders_. It should just stop pointing to the entry.

The benefit to keeping them around and not re-creating them is that we save on time to start if the playlist is already loaded. Not a 100% requirement, but it is helpful.

BrandonOCasey: yeah your right not sure what I was thinking. Will be committing an actual fix in a minute or two

PR #615, opened 2016-03-24
Fix the IE11 bug without introducing a massive bandwidth leak into Flash playback

imbcmdth:
The last release had a bug that meant that if the virtual MediaSource did not contain a *real* MediaSource, HLS would always consider the buffer empty. In a Flash fallback scenario (such as IE11 on Windows 7) we would download segments as quickly as possible repeated while playing back. Not only was this wasting massive amounts of bandwidth but it meant that the transmuxer we processing segments as quickly as possible and negatively affecting playback performance.

Now the condition reads: The *real* MediaSource **must** exist AND the sourceBuffers array **must** be empty in order to return an empty buffer as a work around the IE11 bug. Otherwise, continue processing as normal.


forbesjo:
Looks good to me.

PR #616, opened 2016-03-25
Support EXT-X-KEY:IV prefixed with 0X

makotokw:
I use the Adobe Media Server 5. It publish a manifest contains ``IV=0X1234567890abcdef1234567890abcdef``

I found brew specification.
https://tools.ietf.org/html/draft-pantos-http-live-streaming-18#ref-AES_128
> hexadecimal-sequence: an unquoted string of characters from the
set [0..9] and [A..F] that is prefixed with 0x or 0X

### DIFF
--- a/src/m3u8/parse-stream.js
+++ b/src/m3u8/parse-stream.js
@@ -280,7 +280,7 @@ export default class ParseStream extends Stream {
         event.attributes = parseAttributes(match[1]);
         // parse the IV string into a Uint32Array
         if (event.attributes.IV) {
-          if (event.attributes.IV.substring(0, 2) === '0x') {
+          if (event.attributes.IV.substring(0, 2).toLowerCase() === '0x') {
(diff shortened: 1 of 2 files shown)


src/m3u8/parse-stream.js
```diff
@@ -280,7 +280,9 @@ export default class ParseStream extends Stream {
         event.attributes = parseAttributes(match[1]);
         // parse the IV string into a Uint32Array
         if (event.attributes.IV) {
-          if (event.attributes.IV.substring(0, 2) === '0x') {
+          let prefix = event.attributes.IV.substring(0, 2);
+
+          if (prefix === '0x' || prefix === '0X') {
```
imbcmdth: You should do `prefix = event.attributes.IV.substring(0, 2).toLowerCase()` and simplify the condition here to `prefix === '0x'`.

PR #619, opened 2016-03-28
Wip alt audio combined test fixes

gesinger:


src/master-playlist-controller.js
```diff
@@ -366,7 +333,15 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
-    let mediaGroupName = this.masterPlaylistLoader_.media().attributes.AUDIO;
+    let mediaGroupName;
+    let media = this.masterPlaylistLoader_.media();
+
+    if (!media.attributes) {
+      // source URL was playlist manifest, not master
+      return;
+    }
+
+    mediaGroupName = this.masterPlaylistLoader_.media().attributes.AUDIO;
```
tenacex: can't this line be shortened to `mediaGroupName = media.attributes.AUDIO`?

src/segment-loader.js
```diff
@@ -155,6 +155,9 @@ export default videojs.extend(videojs.EventTarget, {
 
   dispose() {
     this.abort_();
+    if (this.sourceUpdater_) {
+      this.sourceUpdater_.dispose();
```
tenacex: idk if this would be any better, but could shorten to one line with 
```javascript
if (this.sourceUpdater_ && this.sourceUpdater_.dispose()){}
```
Not sure if this is bad practice tho :p

gkatsev: no, don't do this :P

src/segment-loader.js
```diff
@@ -653,6 +660,7 @@ export default videojs.extend(videojs.EventTarget, {
   updateTimeline_(segmentInfo) {
     let segment;
     let timelineUpdate;
+    let timelineUpdated;
```
tenacex: Variables might get confusing...

src/master-playlist-controller.js
```diff
@@ -366,7 +333,15 @@ export default class MasterPlaylistController extends videojs.EventTarget {
   useAudio(label) {
     let audioEntry;
     let newAudioPlaylistLoader;
-    let mediaGroupName = this.masterPlaylistLoader_.media().attributes.AUDIO;
```
tenacex: Shouldn't all of these be declared with just one let, like:
```javascript
let var1,
    var2 = 'meh',
    var3;
```

src/segment-loader.js
tenacex: One let?

src/segment-loader.js
```diff
@@ -667,17 +675,15 @@ export default videojs.extend(videojs.EventTarget, {
...
 
     // the last segment append must have been entirely in the
     // already buffered time ranges. adjust the timestamp offset to
     // fetch forward until we find a segment that adds to the
     // buffered time ranges and improves subsequent media index
     // calculations.
-//    if (!timelineUpdated) {
-//      this.expired_ -= segment.duration;
-//    }
+    if (!timelineUpdated) {
```
tenacex: do you need the timelineUpdated variable if you only use it here?

src/videojs-contrib-hls.js
```diff
@@ -107,7 +103,8 @@ export default class HlsHandler extends Component {
     this.tech_.audioTracks().addEventListener('change', () => {
       for (let i = 0; i < this.tech_.audioTracks().length; i++) {
         if (this.tech_.audioTracks()[i].enabled) {
```
tenacex: I think you should make a variable for `this.tech_.audiTracks()[i]`

src/videojs-contrib-hls.js
```diff
@@ -188,41 +188,34 @@ export default class HlsHandler extends Component {
...
-    if (this.playlists.media().uri === this.source_.src) {
-      playListUrl = resolveUrl(this.source_.src, segmentRelativeUrl);
-    } else {
-      playListUrl =
-        resolveUrl(resolveUrl(this.source_.src, this.playlists.media().uri || ''),
-                   segmentRelativeUrl);
+  fillAudioTracks_() {
+    let master = this.masterPlaylistController_.master();
+    let media = this.masterPlaylistController_.initialMedia();
+    let mediaGroupName = media.attributes.AUDIO;
```
tenacex: one let.

src/videojs-contrib-hls.js
```diff
@@ -188,41 +188,34 @@ export default class HlsHandler extends Component {
...
-   */
-  setBandwidth(localXhr) {
-    // calculate the download bandwidth
-    this.segmentXhrTime = localXhr.roundTripTime;
-    this.bandwidth = localXhr.bandwidth;
-    this.bytesReceived += localXhr.bytesReceived || 0;
-
-    this.tech_.trigger('bandwidthupdate');
+    mediaGroup = master.mediaGroups.AUDIO[mediaGroupName];
+    for (let key in mediaGroup) {
```
tenacex: key should be instantiated at top of function

src/master-playlist-controller.js
gesinger: Nice catch. Will do.

src/master-playlist-controller.js
gesinger: With our ES6 styling, each let gets its own line.

src/segment-loader.js
gesinger: Same as other.

src/segment-loader.js
gesinger: Might not be strictly necessary, but it seemed to read a little better. That said, maybe that isn't true. I'll try it out without the variable and see how it looks.

src/videojs-contrib-hls.js
gesinger: Yeah, a variable for this.tech_.audioTracks() looks like a good idea.

src/videojs-contrib-hls.js
gesinger: Same as above.

src/videojs-contrib-hls.js
gesinger: For ES6, it actually is block scoped here.

src/segment-loader.js
tenacex: so sexy tho

src/master-playlist-controller.js
tenacex: My bad

src/segment-loader.js
gkatsev: nah, there's no benefit, at that point, you might as well just drop the if and just do
```js
this.sourceUpdater_ && this.sourceUpdater_.dispose()
```

src/videojs-contrib-hls.js
tenacex: I really gotten learn my es6

src/segment-loader.js
tenacex: ooooh, do it @gesinger :p

src/master-playlist-controller.js
gesinger: No worries. I was doing it until the linter caught me.

src/segment-loader.js
```diff
@@ -667,17 +675,15 @@ export default videojs.extend(videojs.EventTarget, {
...
+      this.expired_ -= segment.duration;
```
imbcmdth: This was comment out for a reason. I decided that we shouldn't overload the `expired` time tracking because it just makes things more difficult to reason about. I have another PR #626 that adds this back but uses a separate value for this purpose that we reset once we have "gotten back on track".

gesinger: I'll comment it back out again for now then. I think I added it back to fix some tests, but I'll skip them for the time being.

src/videojs-contrib-hls.js
```diff
@@ -10,10 +10,8 @@ import {Decrypter, AsyncStream, decrypt} from './decrypter';
 import utils from './bin-utils';
 import {MediaSource, URL} from 'videojs-contrib-media-sources';
 import m3u8 from './m3u8';
-import videojs from 'video.js';
-import resolveUrl from './resolve-url';
+import {default as videojs, AudioTrack} from 'video.js';
```
BrandonOCasey: didn't know you could do this! Awesome

test/master-playlist-controller.test.js
```diff
@@ -101,7 +105,10 @@ QUnit.test('creates combined and audio only SegmentLoaders', function() {
               'created alternate audio track segment loader');
 });
 
-QUnit.test('if buffered, will request second segment byte range', function() {
+/* eslint-disable no-warning-comments */
+// TODO add back in once logic is returned in segment-loader
+/* eslint-enable no-warning-comments */
+QUnit.skip('if buffered, will request second segment byte range', function() {
```
BrandonOCasey: Should we still be skipping this one? Not that you have to add it back in right now, just curious.

gesinger: Yeah, for now. Once logic similar to the timelineUpdated check is added back in, we can return it: https://github.com/videojs/videojs-contrib-hls/pull/619/files#diff-d5c992e639e8f39a5f94c4a3a6582a00R677

src/playlist-loader.js
```diff
@@ -207,9 +207,11 @@ const PlaylistLoader = function(srcUrl, withCredentials) {
 
   loader.stopRequest = () => {
     if (request) {
-      request.onreadystatechange = null;
-      request.abort();
+      let oldRequest = request;
```
BrandonOCasey: why oldRequest here?

gesinger: I couldn't think of a better name. I was thinking that "r" was too ambiguous, but I suppose it is better than "oldRequest." Any suggestions though?

BrandonOCasey: ehh I think oldRequest works when I look at it some more. The only other thing I could think of naming it would be `req`

BrandonOCasey:
LGTM

src/source-updater.js
```diff
@@ -139,4 +133,10 @@ export default class SourceUpdater {
       callbacks[0]();
     }
   }
+
+  dispose() {
+    if (this.sourceBuffer_ && this.mediaSource.readyState === 'open') {
+      this.sourceBuffer_.abort();
```
tenacex: No expert on this pr, but should you maybe check to see if we have `this.mediaSource`?

gesinger: In this case no one should be creating a source updater without passing in a media source (I am pretty sure).

src/segment-loader.js
```diff
@@ -667,18 +672,15 @@ export default class SegmentLoader extends videojs.EventTarget {
...
     // already buffered time ranges. adjust the timestamp offset to
     // fetch forward until we find a segment that adds to the
     // buffered time ranges and improves subsequent media index
     // calculations.
-//    if (!timelineUpdated) {
-//      this.expired_ -= segment.duration;
-//    }
+    // if (!timelineUpdated) {
+    //   this.expired_ -= segment.duration;
+    // }
```
imbcmdth: This kind of change is just going to create merge conflicts. Specifically, when #626 is merged..

src/segment-loader.js
```diff
@@ -667,18 +672,15 @@ export default class SegmentLoader extends videojs.EventTarget {
                                                    this.sourceUpdater_.buffered());
 
     // Update segment meta-data (duration and end-point) based on timeline
-    // let timelineUpdated =
-    updateSegmentMetadata(playlist,
-                          currentMediaIndex,
-                          timelineUpdate);
+    updateSegmentMetadata(playlist, currentMediaIndex, timelineUpdate);
```
imbcmdth: This kind of change is just going to create merge conflicts. Specifically, when #626 is merged..

src/source-updater.js
```diff
@@ -99,23 +98,18 @@ export default class SourceUpdater {
...
   }
 
-  /**
-   * Queue an update to remove a time range from the buffer.
-   * @see http://www.w3.org/TR/media-source/
-   *      #widl-SourceBuffer-remove-void-double-start-unrestricted-double-end
-   */
   updating() {
-    return this.sourceBuffer_.updating;
+    return this.sourceBuffer_ && this.sourceBuffer_.updating;
```
imbcmdth: I'm wondering if this should be 

```
return !this.sourceBuffer_ || this.sourceBuffer_.updating;
```
Since code outside of sourceUpdater to determine when it is ok to load segments. Right now, the code outside will get the "all clear" when there is no `sourceBuffer_`. With the change above, outside code would think we were busy (which we kind of are; in a way) and not do anything silly quite yet.

src/master-playlist-controller.js
```diff
@@ -678,17 +682,24 @@ export default class MasterPlaylistController extends videojs.EventTarget {
 
   dispose() {
     this.masterPlaylistLoader_.dispose();
-
```
imbcmdth: Stop removing my beautify whitespace! :laughing:

src/master-playlist-controller.js
```diff
@@ -238,8 +238,10 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       if (master.mediaGroups && master.mediaGroups.AUDIO) {
         for (let groupKey in master.mediaGroups.AUDIO) {
           for (let labelKey in master.mediaGroups.AUDIO[groupKey]) {
+            /* eslint-disable no-warning-comments */
```
imbcmdth: We should probably keep the comment warnings. They don't break tests and it is good to be reminded that TODOs exist.

src/videojs-contrib-hls.js
```diff
@@ -177,7 +178,10 @@ export default class HlsHandler extends Component {
         /* eslint-enable dot-notation */
       }
       this.masterPlaylistController_.useAudio();
```
imbcmdth: This line must be moved to `loadedmetadata` handler for it to have any effect. Right now we don't select an audio track if the source is demuxed (pretty much everything but Advanced BipBop.)

gesinger:
Made changes.

src/master-playlist-controller.js
tenacex: mwahaaha

PR #623, opened 2016-03-29
maat cleanup

BrandonOCasey:
use class syntax for SegmentLoader and SourceUpdater
this fixes playlist switching
move audioTrack addition back into hls
we now set this.audioPlaylistLoader_ to newPlaylistLoader right away
improved the code for adding an AudioTrack to video.js
ran the linter
removed loadaltaudio function
fixed seperate video/audio buffers

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -331,7 +331,6 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     let media = this.masterPlaylistLoader_.media();
     let master = this.masterPlaylistLoader_.master;
 
-    // we have been called to early
     if (!media || !master) {
       videojs.log.warn('useAudio() was called before playlist was loaded');
       return;
@@ -369,7 +368,8 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       }
     }
 
-    // all audio tracks are disabled somehow
+    // all audio tracks are disabled somehow, safari keeps playing
+    // so should we
     if (!label) {
       return;
     }
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1,9 +1,7 @@
-/* eslint-disable */
 import PlaylistLoader from './playlist-loader';
 import SegmentLoader from './segment-loader';
 import Ranges from './ranges';
 import videojs from 'video.js';
-import {AudioTrack} from 'video.js';
 
 // 5 minute blacklist
 const BLACKLIST_DURATION = 5 * 60 * 1000;
@@ -225,7 +223,6 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     this.masterPlaylistLoader_.on('loadedmetadata', () => {
       let media = this.masterPlaylistLoader_.media();
       let master = this.masterPlaylistLoader_.master;
-      let alternateAudioUri;
 
       // if this isn't a live video and preload permits, start
       // downloading segments
@@ -244,6 +241,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
             // TODO: use one playlist loader for alternate audio and
             // update the src when it is being used
             let audio = master.mediaGroups.AUDIO[groupKey][labelKey];
+
             if (!audio.resolvedUri) {
               continue;
             }
@@ -255,8 +253,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
 
       this.setupSourceBuffer_();
       this.setupFirstPlay();
-      this.hlsHandler.tech_.trigger('loadedmetadata');
-      this.useAudio();
+      this.trigger('loadedmetadata');
     });
 
     this.masterPlaylistLoader_.on('loadedplaylist', () => {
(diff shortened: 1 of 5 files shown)


src/master-playlist-controller.js
```diff
@@ -369,10 +327,20 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
 
-  useAudio(label) {
-    let audioEntry;
-    let newAudioPlaylistLoader;
-    let mediaGroupName = this.masterPlaylistLoader_.media().attributes.AUDIO;
+  useAudio() {
+    let media = this.masterPlaylistLoader_.media();
+    let master = this.masterPlaylistLoader_.master;
+
+    if (!media || !media.attributes || !media.attributes.AUDIO ||
```
gesinger: I think in this case it may be better for us to not do some guards. They are definitely safe to have, but it might hide a bug later on (for instance, if someone does try setting useAudio before they are supposed to, we'd simply return here, instead of an error being thrown which could clue someone into the problem--that they are calling useAudio too early, before they are supposed to--immediately).

gesinger: That said, we definitely need some guards (for instance, the !media.attributes for the case of the src URL being a playlist, and not a master)

src/master-playlist-controller.js
```diff
@@ -369,10 +327,20 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+        !master.mediaGroups || !master.mediaGroups.AUDIO) {
+      return;
+    }
+    let mediaGroupName = media.attributes.AUDIO;
+
+    if (!master.mediaGroups.AUDIO[mediaGroupName]) {
+      return;
+    }
+    let audioEntries = master.mediaGroups.AUDIO[mediaGroupName];
```
gesinger: This is more for me, but do you know if we're doing all of our let declarations at the top of the function, except for variables declared within inner blocks, or are we doing the declaration when we define it within the function?

src/videojs-contrib-hls.js
```diff
@@ -134,6 +133,46 @@ export default class HlsHandler extends Component {
       externHls: Hls
     });
 
+    this.masterPlaylistController_.on('loadedmetadata', () => {
+      let audioTracks = this.tech_.audioTracks();
```
gesinger: This is probably too picky (and introduces shadowing), but we have a variable called audioTrackList that is the same thing above as audioTracks is here (same as in different names representing the same thing, though their values will be different).

gesinger:
I like the cleanup (and it was about time those Ranges didn't have the "private" underscore at the end of them). And the adding audio tracks function on loadedmetadata in HlsHandler looks really clean. A couple general notes:

- I know we discussed doing the audio track setting logic on loadedplaylist instead of loadedmetadata. I think we can cover that in a separate PR. Or here. Up to you.
- There are a couple minor items that I did in https://github.com/videojs/videojs-contrib-hls/pull/619 that I'll copy over when doing a rebase (e.g., an additional method in MasterPlaylistController to get some of the internals like the current media instead of using masterPlaylistLoader_). Will save some time rather than worrying about them here.

src/master-playlist-controller.js
BrandonOCasey: from what I understood we only did that when we used var to declare all variables at the top, comma separated. I don't think we need them all to be at the top with let.

src/videojs-contrib-hls.js
BrandonOCasey: Not to picky at all, they will actually have the same value here, we can just re-use that old variable, as they will both return the same video.js tracklist (it never gets deleted or re-created just cleared out)

BrandonOCasey:
I think we will leave changing the logic to loadedplaylist for another PR. I will work on addressing the other concerns. Thanks for the review!

src/master-playlist-controller.js
BrandonOCasey: should we add a warning log in some cases here rather than just a return? (but only do it for the ones that make sense)

src/videojs-contrib-hls.js
BrandonOCasey: I will just change the name re-use over the scope won't work to well. I guess I misunderstood.

src/master-playlist-controller.js
```diff
@@ -358,6 +368,8 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         break;
       }
     }
+
+    // all audio tracks are disabled somehow
```
gesinger: Is this an error condition?

src/master-playlist-controller.js
```diff
@@ -331,13 +331,23 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     let media = this.masterPlaylistLoader_.media();
     let master = this.masterPlaylistLoader_.master;
 
-    if (!media || !media.attributes || !media.attributes.AUDIO ||
+    // we have been called to early
```
gesinger: I think the warn log is enough (without the comment).

src/master-playlist-controller.js
BrandonOCasey: from the api it is possible, but I am not sure what we should do if it actually happens

gesinger: Interesting, it seems Safari keeps playing the audio.

BrandonOCasey: clarified that we will do the same in the same comment this line points to

gesinger:
LGTM

PR #626, opened 2016-03-30
Re-introduce the concept of time correction

imbcmdth:
A few changes:
* When fetches aren't productive (the append didn't change the buffer), use a time-correction to "bump" the segment fetching logic forward until we find a segment that adds data to the buffer.
* When currentTime is 0, add a fudge factor to the currentTime to account for streams that start a little bit after 0. Because MSE specifies that the `buffered` property is the *intersection* of all source buffers, if the audio source buffer starts a little bit after the video, we end up with a combined `buffered` property that does not start at 0.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -140,6 +140,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.checkBufferTimeout_ = null;
     this.error_ = void 0;
     this.expired_ = 0;
+    this.timeCorrection_ = 0;
     this.currentTimeline_ = -1;
     this.xhr_ = null;
     this.pendingSegment_ = null;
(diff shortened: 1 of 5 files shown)


BrandonOCasey:
seems like this needs a rebase on development

BrandonOCasey:
Once #619 is merged we should try to get the skipped tests in Segment Loader and Master Playlist Controller working for this PR

src/segment-loader.js
```diff
@@ -674,12 +683,14 @@ export default videojs.extend(videojs.EventTarget, {
...
-    // calculations.
-//    if (!timelineUpdated) {
-//      this.expired_ -= segment.duration;
-//    }
+    // already buffered time ranges. adjust the timeCorrection
+    // offset to fetch forward until we find a segment that adds
+    // to the buffered time ranges and improves subsequent media
+    // index calculations.
+    if (!timelineUpdated) {
+      this.timeCorrection_ -= segment.duration;
```
dmlap: Where will this be used? Why do we need a separate variable for it?

src/segment-loader.js
```diff
@@ -674,12 +683,14 @@ export default videojs.extend(videojs.EventTarget, {
...
-    // buffered time ranges and improves subsequent media index
-    // calculations.
-//    if (!timelineUpdated) {
-//      this.expired_ -= segment.duration;
-//    }
+    // already buffered time ranges. adjust the timeCorrection
+    // offset to fetch forward until we find a segment that adds
+    // to the buffered time ranges and improves subsequent media
+    // index calculations.
+    if (!timelineUpdated) {
```
BrandonOCasey: we need to change the rest of the variables in this scope to `timelineUpdated` or change this variable to `timelineUpdate`

imbcmdth: On line 681, `timelineUpdated` is defined.

src/segment-loader.js
imbcmdth: We talked about this in person. The consensus was that `timeCorrection` has a different lifespan from `expired` and conflating the two variables would lead to heartache.

PR #629, opened 2016-03-30
Don't try to load segments that are likely already buffered

imbcmdth:
Don't fetch a segment if it's time-range is already >=90% buffered.

Intersect a TimeRanges object created for the segment with the existing TimeRanges returned by `buffered` and determine how much of the segment is currently `buffered`. If that value is greater than or equal to 90%, don't fetch this segment but repeat the check against the next segment. 

Return if neither segment is less than 90% buffered.

### DIFF
--- a/src/ranges.js
+++ b/src/ranges.js
@@ -113,9 +113,100 @@ const findSoleUncommonTimeRangesEnd = function(original, update) {
   return result[0];
 };
 
+/**
+ * Calculate the intersection of two TimeRanges
+ * @param bufferA {TimeRanges}
+ * @param bufferB {TimeRanges}
+ * @return {TimeRanges} The interesection of `bufferA` with `bufferB`
+ */
+const bufferIntersection = function(bufferA, bufferB) {
+    let start = null;
+    let end = null;
+    let arity = 0;
+    let extents = [];
+    let ranges = [];
+
+    if (!bufferA || !bufferA.length || !bufferB || !bufferB.length) {
+      return videojs.createTimeRange();
+    }
+
+    // Handle the case where we have both buffers and create an
+    // intersection of the two
+    let count = bufferA.length;
+
+    // A) Gather up all start and end times
+    while (count--) {
+      extents.push({time: bufferA.start(count), type: 'start'});
+      extents.push({time: bufferA.end(count), type: 'end'});
+    }
+    count = bufferB.length;
+    while (count--) {
+      extents.push({time: bufferB.start(count), type: 'start'});
+      extents.push({time: bufferB.end(count), type: 'end'});
+    }
+    // B) Sort them by time
+    extents.sort(function(a, b) {
+      return a.time - b.time;
+    });
+
+    // C) Go along one by one incrementing arity for start and decrementing
+    //    arity for ends
+    for (count = 0; count < extents.length; count++) {
+      if (extents[count].type === 'start') {
+        arity++;
+
+        // D) If arity is ever incremented to 2 we are entering an
+        //    overlapping range
+        if (arity === 2) {
+          start = extents[count].time;
+        }
+      } else if (extents[count].type === 'end') {
+        arity--;
+
+        // E) If arity is ever decremented to 1 we leaving an
+        //    overlapping range
+        if (arity === 1) {
+          end = extents[count].time;
+        }
+      }
+
+      // F) Record overlapping ranges
+      if (start !== null && end !== null) {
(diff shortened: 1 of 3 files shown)


src/ranges.js
```diff
@@ -113,9 +113,100 @@ const findSoleUncommonTimeRangesEnd = function(original, update) {
   return result[0];
 };
 
+/**
+ * Calculate the intersection of two TimeRanges
+ * @param bufferA {TimeRanges}
```
BrandonOCasey: Type comes before parameter name and is followed by description (not that its really needed in this case) so it would look something like this. (some of our current doc comments do it backwards, which I will fix later)
``` JavaScript
* @param {TimeRanges} bufferA the first buffer
* @param {TImeRanges} bufferB the second buffer

File: src/ranges.js [hunk 2]
```diff
@@ -113,9 +113,100 @@ const findSoleUncommonTimeRangesEnd = function(original, update) {
...
+
+    return videojs.createTimeRanges(ranges);
+};
+
+/**
+ * Calculates the percentage of `segmentRange` that overlaps the
+ * `buffered` time ranges.
+ * @param segmentRange {TimeRanges} the time range that the segment covers
+ * @param buffered {TimeRanges} the currently buffered time ranges
+ * @return {Number} percent of the segment currently buffered
```
BrandonOCasey: same as the previous comment https://github.com/videojs/videojs-contrib-hls/pull/629/files#r58073036

src/ranges.js
```diff
@@ -113,9 +113,100 @@ const findSoleUncommonTimeRangesEnd = function(original, update) {
...
+ */
+const calculateBufferedPercent = function(segmentRange, buffered) {
+  let segmentDuration = segmentRange.end(0) - segmentRange.start(0);
+
```
BrandonOCasey: Maybe it personal preference here but its nice not to have so many spaces between lets.

src/segment-loader.js
```diff
@@ -258,6 +258,24 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.checkBufferTimeout_ = window.setTimeout(this.monitorBuffer_.bind(this),
                                                  CHECK_BUFFER_DELAY);
   }
+
+  /**
+   * Return the amount of a segment specified by the mediaIndex overlaps
+   * the current buffered content.
+   * @param playlist {object} the playlist object to fetch segments from
+   * @param mediaIndex {number} the index of the segment in the playlist
```
BrandonOCasey: same as the previous comment https://github.com/videojs/videojs-contrib-hls/pull/629/files#r58073036

src/segment-loader.js
```diff
@@ -299,7 +317,30 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    if (percentBuffered >= 90) {
+      // Retry the buffered calculation with the next segment if there is another
+      // segment after the currently selected segment
+      if (mediaIndex + 1 < playlist.segments.length) {
+        percentBuffered = this.getSegmentBufferedPercent_(playlist, mediaIndex + 1, buffered);
+      }
+
+      // If both checks failed return and don't load anything
+      if (percentBuffered >= 90) {
+        return;
```
BrandonOCasey: do we want to log some kind of warning here?

BrandonOCasey:
tests? or are we just trying to get something working for now?

gquinones:
Rado and I can provide some assistance in evaluating how this change might
impact bandwidth consumption by the player.

Let us know if you want us to help.


On Thu, Mar 31, 2016 at 10:32 AM, Brandon Casey <notifications@github.com>
wrote:

> tests? or are we just trying to get something working for now?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/videojs/videojs-contrib-hls/pull/629#issuecomment-203988103>
>



-- 
- Gerardo Quiñones

src/ranges.js
imbcmdth: Good point!

src/segment-loader.js
imbcmdth: Nah, we should be quiet unless something has actually gone wrong. I think we are too verbose as it is. We'll need to remove some of the logging at some point. (Or hide it behind a debug flag of some kind if it's useful.)

src/ranges.js
imbcmdth: I want to blame someone else for these broken style comments. It wasn't me, I don't even write comments! :wink:

PR #641, opened 2016-04-05
Workaround for the firefox audio sample/channel change crash

BrandonOCasey:
MPC re-emits audiotrackchanged and HLS picks it up
unify the interface to get parsed playlist attributes in MPC

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -189,6 +189,9 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     this.mediaSource = new videojs.MediaSource({ mode: this.mediaSourceMode });
     // load the media source into the player
     this.mediaSource.addEventListener('sourceopen', this.handleSourceOpen_.bind(this));
+    this.mediaSource.addEventListener('audioinfochanged', (e) => {
+      this.trigger(e);
+    });
 
     this.hlsHandler = hlsHandler;
     this.hlsHandler.mediaSource = this.mediaSource;
@@ -326,6 +329,30 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     if (this.audioPlaylistLoader_) {
       this.audioSegmentLoader_.load();
     }
+  }
+
+  getPlaylistAttributes_() {
+    let media = this.media();
+    let master = this.master();
+    let mediaGroups = master.mediaGroups;
+    let attributes = {
+      audio: {main: {default: true}}
+    };
+
+    if (!media.attributes) {
+      // source URL was playlist manifest, not master
+      // no audio tracks to add
+      return;
+    }
+
+    // only do alternative audio tracks in html5 mode, and if we have them
+    if (this.mediaSourceMode === 'html5' &&
+        media.attributes &&
+        media.attributes.AUDIO &&
+        mediaGroups.AUDIO[media.attributes.AUDIO]) {
+      attributes.audio = mediaGroups.AUDIO[media.attributes.AUDIO];
+    }
+    return attributes;
   }
 
   useAudio() {
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -532,7 +532,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
    * making it unavailable for selection by the rendition selection algorithm
    * and then forces a new playlist (rendition) selection.
    */
-  blacklistCurrentPlaylist(error) {
+  blacklistCurrentPlaylist(error = {}) {
     let currentPlaylist;
     let nextPlaylist;
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -284,7 +284,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.masterPlaylistLoader_.on('error', () => {
-      this.blacklistCurrentPlaylist_(this.masterPlaylistLoader_.error);
+      this.blacklistCurrentPlaylist(this.masterPlaylistLoader_.error);
     });
 
     this.masterPlaylistLoader_.on('mediachanging', () => {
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
This still needs tests but is waiting on #619

src/master-playlist-controller.js
```diff
@@ -328,6 +331,30 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     }
   }
 
+  getPlaylistAttributes_() {
```
gesinger: If this is used in videojs-contrib-hls, should we remove the "private" underscore?

src/videojs-contrib-hls.js
```diff
@@ -134,27 +134,73 @@ export default class HlsHandler extends Component {
...
+
+      let trackToRemove;
+      let mainTrack;
+
+      for (let i = 0; i < audioTrackList.length; i++) {
+        let track = audioTrackList[i];
+
+        if (track.enabled) {
+          trackToRemove = track;
+        } else if (track.label === mainLabel) {
```
gesinger: I guess it would signal a bigger problem if the main track caused the change and had the issue. If it is possible to have that condition though, it looks like this might throw an error (down when we call mainTrack.enabled).

gesinger: Nevermind, I noticed the guard below. Looks safe!

src/videojs-contrib-hls.js
```diff
@@ -175,29 +176,31 @@ export default class HlsHandler extends Component {
       // they did not switch audiotracks
       // blacklist the current playlist
       if (trackToRemove.label === mainLabel) {
-        error = 'The rendition that we tried to switch to ' + error +
+        error = `The rendition that we tried to switch to ${error}` +
                 'Unfortunately that means we will have to blacklist ' +
                 'the current playlist and switch to another. Sorry!';
-        this.masterPlaylistController_.blacklistCurrentPlaylist(error);
-        return;
+        change = () => this.masterPlaylistController_.blacklistCurrentPlaylist();
```
gesinger: I think this will break if error is undefined.

gesinger:
LGTM

gesinger:
(Though Travis has some complaints)

src/videojs-contrib-hls.js
```diff
@@ -134,27 +134,76 @@ export default class HlsHandler extends Component {
...
+                  'https://bugzilla.mozilla.org/show_bug.cgi?id=1247138\n\n';
+
+      let attributes = this.masterPlaylistController_.getPlaylistAttributes();
+      let mainLabel;
+
+      for (let label in attributes.audio) {
+        let hlstrack = attributes.audio[label];
+
+        /* eslint-disable dot-notation */
+        // dot notation used on the key default to support ie8
```
dmlap: This code would never be run in IE8.

src/master-playlist-controller.js
```diff
@@ -505,7 +532,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
    * making it unavailable for selection by the rendition selection algorithm
    * and then forces a new playlist (rendition) selection.
    */
-  blacklistCurrentPlaylist_(error) {
+  blacklistCurrentPlaylist(error = {}) {
```
dmlap: Why add the default value?

src/videojs-contrib-hls.js
```diff
@@ -134,27 +134,76 @@ export default class HlsHandler extends Component {
...
+        if (hlstrack['default']) {
+          mainLabel = label;
+          break;
+        }
+        /* eslint-enable dot-notation */
+      }
+
+      let trackToRemove;
```
dmlap: I would be inclined to mark it as unusable somehow, rather than removing it. I think it would be clearer in a hypothetical UI to mark an audio track as "bad" than having it just disappear.

dmlap:
I don't see anything that would pause the segment loader if a bad audio track is encountered. Am I just missing the connection?

src/videojs-contrib-hls.js
BrandonOCasey: We do need IE8 to not throw an error when loading the script though. I can do some testing to verify that we need these. When we were working with webworkify they added an unquoted parameter called `default` that would error throw an error on IE8 while loading the script.

src/videojs-contrib-hls.js
BrandonOCasey: Right now don't have an api for an audio button yet, and we don't have anything in the official API on video.js to disable or mark a track as bad. Do we want to change this functionality when something like that comes around, save this PR for later when we can get an API addition, or do something else that would probably be slightly hacky.

src/master-playlist-controller.js
BrandonOCasey: Basically so that we can call `blacklistCurrentPlaylist` without passing an error which we only used for the `error.playlist` value anyway. I can switch it back to not using the default but then I will just be doing the same thing the code does when it finds that error has no playlist:  `currentPlaylist = error.playlist || this.masterPlaylistLoader_.media();`

dmlap: This method name is very confusing. blacklistCurrentPlaylist() optionally takes a non-current playlist to blacklist? Ugh.

BrandonOCasey: agreed it is rather confusing

src/videojs-contrib-hls.js
dmlap: I think that only occurs if it's in the parse-time execution path (but it's definitely worth checking).

src/videojs-contrib-hls.js
dmlap: I was thinking add a property like `error_` to the audio track and then revert any attempts to switch to that track in the future. That might fall into the "hacky" category for you :) The advantage (IMO) is API consumers can introspect on the audio tracks and discover something is wrong more easily. It would probably take some thinking to figure out why an audio track disappeared if you weren't familiar with our code and the complexity of the problem.

BrandonOCasey: well we do log an error, with lots of detail in the console

PR #643, opened 2016-04-06
Just upload source files to npm

forbesjo:
This PR is to reduce the amount of files there are on npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,15 +80,13 @@
   },
   "files": [
     "CONTRIBUTING.md",
-    "dist-test/",
     "dist/",
     "docs/",
     "es5/",
     "index.html",
     "scripts/",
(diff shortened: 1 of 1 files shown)


gkatsev:
This will break everyone using contrib-hls. Those that include it via browserify will break because the es5 folder is missing. Those that include it via the dist folder would break because that folder is missing.

gkatsev:
Oh, too much read obscured things, while it should work, I think it's nicer to have all the files included in npm.

gkatsev:
removing `dist-test` sounds fine. I think keeping the rest would be good.

forbesjo:
I'll just remove `dist-test` and `utils` (20mb) in this PR.

PR #645, opened 2016-04-07
Added a unit test to verify switching media groups during playback works

BrandonOCasey:
silenced videojs.log during tests
updated multipleAudioGroups files to be easier to understand
changed caret character to single quote where we were not using it as a template string
Pulls in changes from #642

### DIFF
--- a/test/test-helpers.js
+++ b/test/test-helpers.js
@@ -1,6 +1,7 @@
 import document from 'global/document';
 import sinon from 'sinon';
 import videojs from 'video.js';
+import QUnit from 'qunit';
 /* eslint-disable no-unused-vars */
 // needed so MediaSource can be registered with videojs
 import MediaSource from 'videojs-contrib-media-sources';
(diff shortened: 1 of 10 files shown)


test/test-helpers.js
```diff
@@ -112,12 +112,21 @@ let fakeEnvironment = {
...
+    });
+
   }
 };
 
 export const useFakeEnvironment = function() {
   fakeEnvironment.clock = sinon.useFakeTimers();
-
+  ['warn', 'error'].forEach((level) => {
+    sinon.stub(videojs.log, level);
```
gesinger: I agree that the warnings are a pain in the tests, but are they something we should address? Or can we never get around them, as we need to use the internals of videojs?

test/videojs-contrib-hls.test.js
```diff
@@ -1793,6 +1793,81 @@ QUnit.test('cleans up the buffer when loading VOD segments', function() {
...
+  let hls = this.player.tech_.hls;
+  let mpc = hls.masterPlaylistController_;
+  let at = this.player.audioTracks();
+
+  QUnit.equal(at.length, 3, 'three audio tracks after load');
+  let trackOne = at[0];
+  let trackTwo = at[1];
+  let trackThree = at[2];
+
+  QUnit.equal(trackOne.enabled, true, 'track on enabled after load');
```
gesinger: Minor type (on => one)

gesinger: I even made a typo in the last comment...

test/videojs-contrib-hls.test.js
```diff
@@ -1793,6 +1793,81 @@ QUnit.test('cleans up the buffer when loading VOD segments', function() {
...
+
+  let oldMediaGroup = mpc.media().attributes.AUDIO;
+
+  // force mpc to select a new media group
+  mpc.hlsHandler.selectPlaylist = () => {
+    let select = null;
+
+    mpc.master().playlists.forEach((playlist) => {
```
gesinger: We can probably use:

```javascript
return mpc.master().playlists.find(playlist => playlist.attributes.AUDIO !== oldMediaGroup);
```

gesinger:
Nice test addition. Some minor comments, but otherwise, looks good.

test/test-helpers.js
BrandonOCasey: I think we may be able to get rid of this issue without silencing the log, took a few minutes to narrow it down, I will write a test around it to verify that nothing changes basically it involves the depreciation warning when we use `player.hls`. when we were checking if `player.hls` existed when it did exist we inadvertently called it and caused the warning to showup.

test/videojs-contrib-hls.test.js
BrandonOCasey: :)

test/videojs-contrib-hls.test.js
imbcmdth: I don't think `find` is polyfilled.. it'll work in Chrome but not everywhere.

BrandonOCasey:
ok so ultimately I decided that it would be good to force us to make sure that we are logging. So I mock the log as a did previously and on this.env.restore() we verify that log functions were called 0 times. I added a getter to the log stubbing that will reset it to zero. Basically if you check how many calls happened with that then we know you checked (I guess you could still throw it to the void but CR's would catch it).

test/videojs-contrib-hls.test.js
BrandonOCasey: seems to work on Firefox/Chrome

PR #646, opened 2016-04-08
fixed bug :  if m3u8 link is a redirect url ,it's not working

YouYII:
I use this link : http://zy160325.podbean.org/hls/web/d6gidx/2.mp4,
It works on Safari  and  [hlsjs](https://github.com/dailymotion/hls.js) , 
But it does not work on this project , 
I fixed this problem.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -338,7 +338,10 @@ export default class PlaylistLoader extends Stream {
 
       loader.state = 'HAVE_MASTER';
 
-      parser.manifest.uri = srcUrl;
+      //follow redirect uri
+      if (parser.manifest.uri != req.responseURL) {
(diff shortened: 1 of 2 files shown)


samward1985:
Ha! It looks like we were doing this at the exact same time, and came up with the same solution: https://github.com/videojs/videojs-contrib-hls/pull/647

:D

dmlap:
Mentioned it in #647 but IE doesn't support `responseURL`

forbesjo:
Could you add a check for responseURL to handle IE?

dmlap:
We're trying to support IE and (as written) this PR wouldn't work because of its use of responseURL. If you have another approach to solve this problem, don't hesitate to re-open this!

Issue #548, opened 2016-02-10
Base URL is not updated if url was redirected


Naouak:
If the HLS playlist source url return a 302 HTTP code to another location, videojs-contrib-hls doesn't update the url for subsequent playlist elements.

Example:
http://example1.com/playlist.m3u8 > 302 to http://example2.com/playlist.m3u8
http://example2.com/playlist.m3u8 > first chunk: chunk1.ts
The player should query : http://example2.com/chunk1.ts
The player is currently querying : http://example1.com/chunk1.ts

gkatsev:
Hm... that's an interesting problem. The spec for the 302 status code tells us to not use the new URI for the new requests, 301 tells us to use the new URI. However, segments and media manifests are relative to the location of the m3u8 file, so, if the playlist was redirected, do we use the new URI for the playlist or our original one?
I assume that Safari/iOS is able to play this correctly?

Naouak:
Safari and Edge play this correctly.

I look if there is an easy way to get this issue right but it seems XHR
doesn't tell if there was a redirection.

gkatsev:
Yeah, it looks like we cannot read the value without making a separate HEAD request directly and looking at the headers: https://dvcs.w3.org/hg/xhr/raw-file/tip/Overview.html#infrastructure-for-the-send%28%29-method

dmlap:
I'm going to close this because it seems like a browser limitation that we won't be able to work around. If anyone has any creative solutions though, please comment!

dcamillo:
Hi Naouak,
Are you sure you are able to play this on EDGE ?
I have this issue on IE-EDGE.
So I am testing the index.html
https://github.com/videojs/videojs-contrib-hls/blob/master/index.html

I clone the project using git and at first attemp does not matter URL I place in the TextBox it does not play the new URL, always the one was loaded in the Video tag.

Am I doing something wrong ? Are you guys able to play it successfuly ?

PR #647, opened 2016-04-08
Allow dynamic source switching

samward1985:
To preface, I'm fairly sure that this isn't the best solution to this problem, but it works for my case, and thought it might at least be helpful to look at.

#### The problem
We are using a live video provider that doesn't give us a raw m3u8 URL, but rather, an API URL that redirects to the m3u8. For example, we'd give the player www.streamingsite.com/api?video=1234, which would redirect to an m3u8 hosted on a CDN. The problem was that the player would attempt to play www.streamingsite.com/api/video.m3u8 instead of www.cdn.com/video.m3u8. For various reasons, we don't want to, say, CURL the API URL before feeding it to the player.

#### A solution
With this PR, if the initial response URL doesn't match the source URL, it is assumed (perhaps dangerously?) that we're dealing with a redirect of some kind, and we update the source URL in the HLS Handler.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -337,7 +337,13 @@ export default class PlaylistLoader extends Stream {
 
       loader.state = 'HAVE_MASTER';
 
-      parser.manifest.uri = srcUrl;
+      // if the responseURL doesn't match the url requested initially, we're assuming an api call
+      if (req.responseURL != srcUrl) {
+        parser.manifest.uri = req.responseURL;
+        loader.trigger('mediasourcechange', req.responseURL);
+      } else {
+        parser.manifest.uri = srcUrl;
+      }
 
       // loaded a master playlist
       if (parser.manifest.playlists) {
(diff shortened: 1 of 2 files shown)


gkatsev:
Interesting, I didn't realize that we actually had access to the redirected url. We had an issue about this recently (https://github.com/videojs/videojs-contrib-hls/issues/548) and we thought we couldn't handle it.

dmlap:
[responseURL](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseURL) isn't supported in IE :(

forbesjo:
Could you add a check for responseURL to handle IE?

dmlap:
@forbesjo there is no equivalent mechanism in IE. We could only use responseURL where available but that would make m3u8s with this type of redirect fail in IE and work elsewhere. That's pretty confusing for end-users and developers so I lean towards failing everywhere and encouraging people to not host m3u8s with an initial redirect.

PR #648, opened 2016-04-08
do not do maat on ie11

BrandonOCasey:


dmlap:
Some combinations of Windows10/IE11 have bad MSE behavior that we don't think is workaround-able or even detectable. We decided to note this in the README rather than baking-in Flash-based MSE with UA detection in case other people wanted to try it out for themselves.

BrandonOCasey:
see #651

PR #652, opened 2016-04-11
expose global settings and add unit tests to verify

BrandonOCasey:


test/videojs-contrib-hls.test.js
```diff
@@ -1133,10 +1133,10 @@ QUnit.test('reloads out-of-date live playlists when switching variants', functio
 });
 
 QUnit.test('if withCredentials global option is used, withCredentials is set on the XHR object', function() {
-  let hlsOptions = videojs.options.hls;
+  let hlsOptions = videojs.Hls.options;
 
   this.player.dispose();
-  videojs.options.hls = {
```
dmlap: Does this mean the old options location will be ignored?

BrandonOCasey: It would I think I have a more backwards compatible solution, I will submit it shortly

dmlap:
@BrandonOCasey what's the intended precedence on the options? It would be nice to have a test that enforced that (if we don't already).

dmlap:
Otherwise, LGTM

tenacex:
Looks good

BrandonOCasey:
@dmlap we have tests to verify that mode and withCredentials options will work when passed in with a source, or passed in globally. I am not sure what your first question is asking.

Issue #621, opened 2016-03-29
Adaptive bitrate selection doesn't work in fullscreen (Chrome)


drizzt:
When I go to fullscreen and so my resolution change the stream is not switched to a "better" one

dmlap:
Right now, we don't throw away any buffered content when you move to fullscreen. With our standard buffering settings, that means it can take up to 30 seconds for the player to adapt to the new dimensions.

dmlap:
We could remove some portion of the buffer (currentTime + 5 seconds, say?) on the fullscreen transition to force a quicker upswitch. The downside of that approach is we risk stalling if the network happens to dip in performance right then.

gesinger:
This was since released as part of #659 . While the timing has changed slightly for fastQualityChange_ since that release, this should switch faster than when first reported.

Thank you again for reporting @drizzt .

PR #659, opened 2016-04-20
Switch quality levels quickly when going fullscreen

dmlap:
Detect when the player is taken fullscreen and clear out some of the video buffer to trigger a quicker quality switch. Fix some tests that were using bad mocks that polluted the environment.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -447,6 +447,21 @@ export default class MasterPlaylistController extends videojs.EventTarget {
   }
 
   /**
+   * Re-tune playback quality level for the current player
+   * conditions. This method may perform destructive actions, like
+   * removing already buffered content, to readjust the currently
+   * active playlist quickly.
+   */
+  fastQualityChange_() {
+    let media = this.hlsHandler.selectPlaylist();
+
+    if (media !== this.masterPlaylistLoader_.media()) {
+      this.masterPlaylistLoader_.media(media);
+      this.mainSegmentLoader_.sourceUpdater_.remove(this.currentTimeFunc() + 5, Infinity);
+    }
+  }
+
+  /**
    * Seek to the latest media position if this is a live video and the
    * player and video are loaded and initialized.
    */
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -3,6 +3,7 @@
  * The main file for the HLS project.
  * License: https://github.com/videojs/videojs-contrib-hls/blob/master/LICENSE
  */
+import document from 'global/document';
 import PlaylistLoader from './playlist-loader';
 import Playlist from './playlist';
 import xhr from './xhr';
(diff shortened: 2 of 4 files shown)


dmlap:
Fix for #621

src/master-playlist-controller.js
```diff
@@ -446,6 +446,22 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
   }
 
+
+  /**
+   * Re-tune playback quality level for the current player
+   * conditions. This method may perform destructive actions, like
+   * removing already buffered content, to readjust the currently
+   * active playlist quickly.
+   */
+  fastQualityChange_() {
```
gesinger: If this is being used outside of master-playlist-controller (i.e., in videojs-contrib-hls), should we remove the underscore that identifies it as private?

src/videojs-contrib-hls.js
```diff
@@ -95,6 +95,25 @@ export default class HlsHandler extends Component {
     this.bandwidth = this.options_.bandwidth || 4194304;
     this.bytesReceived = 0;
 
+    // listen for fullscreenchange events for this player so that we
+    // can adjust our quality selection quickly
+    if (document.fullscreenEnabled || document.webkitFullscreenEnabled) {
+      this.on(document, [
```
gkatsev: this is because the `fullscreenchange` bubbles up from the player element to the document and as a way to get around the infinite loop that could be created by listening to it on the player/tech?

src/master-playlist-controller.js
```diff
@@ -446,6 +446,22 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+    let media = this.hlsHandler.selectPlaylist();
+
+    if (media !== this.masterPlaylistLoader_.media()) {
+      this.masterPlaylistLoader_.media(media);
+      this.mainSegmentLoader_.sourceUpdater_.remove(this.currentTimeFunc() + 5, Infinity);
```
gesinger: Should we be using 5 as a set amount, or do we want to do a segment duration if we have that information available?

gesinger: And do we want to add a method to mainSegmentLoader for removing certain time ranges?

test/master-playlist-controller.test.js
```diff
@@ -106,6 +106,45 @@ QUnit.test('creates combined and audio only SegmentLoaders', function() {
...
+    removes.push({ start, end });
+  };
+  this.masterPlaylistController.hlsHandler.selectPlaylist = () => {
+    this.masterPlaylistController.master().playlists[0];
+  };
+
+  this.masterPlaylistController.fastQualityChange_();
+
+  equal(removes.length, 1, 'removed buffered content');
+  equal(removes[0].end, Infinity, 'removed to the end');
```
gesinger: Should we add a check for the start as well as the end?

src/master-playlist-controller.js
dmlap: The underscore means "don't use outside this project" to me. I feel like this is a handy debugging capability but I don't think we advertise it as a first-class part of the API (today).

src/videojs-contrib-hls.js
dmlap: Yeah. I experimented with a *ton* of ways to avoid the infinite event dispatch loop between the player and tech. A number of them technically did the job but they felt grafted-in or added significant capabilities to video.js that I didn't think were warranted. I'm not a huge fan of touching globals but this seems safe & correct to me.

src/master-playlist-controller.js
dmlap: I debated using the segment loader's bandwidth info and the bitrate of the newly selected bitrate to calculate a safe removal range. Seems like it might be a good idea but I thought it could wait for a separate PR. Seem reasonable?

Proxying remove() calls from the segment loader to the source updater increases encapsulation but I don't think it was worthwhile given all the actors are in the same project in this case.

test/master-playlist-controller.test.js
dmlap: I'll do that.

src/master-playlist-controller.js
gesinger: Both seem reasonable

gesinger:
LGTM

PR #661, opened 2016-04-20
Extend Videojs AudioTracks and fix an issue in Firefox

BrandonOCasey:
added an hls-audio-track class to handle hls AudioTrack needs
readme was updated by doctoc for previous commits changes
updated the code to use this new AudioTrack
added a firefox issue workaround with audioinfo changes

contains changes from #641


src/master-playlist-controller.js
```diff
@@ -321,37 +306,79 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+    return mediaGroup;
+  }
+
+  audioTracks() {
+    this.audioTracks_ = this.audioTracks_ || [];
+    return this.audioTracks_;
+  }
+
+  useAudio() {
+    let track = this.audioTracks().find((t) => t.enabled === true);
```
imbcmdth: You can't use Array's `find` that only works on Chrome.

src/master-playlist-controller.js
```diff
@@ -695,11 +706,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
 
   dispose() {
     this.masterPlaylistLoader_.dispose();
-    for (let loader in this.audioPlaylistLoaders_) {
-      if (this.audioPlaylistLoaders_.hasOwnProperty(loader)) {
-        this.audioPlaylistLoaders_[loader].dispose();
-      }
-    }
```
imbcmdth: Don't we need to still dispose of audioPlaylistLoaders? I think the active playlistLoader could have a timer (to refresh the playlist on a live hls stream) or an outstanding XHR request that we aren't cleaning up/aborting.

BrandonOCasey: we do, I will add some code to loop through and dispose them

src/master-playlist-controller.js
BrandonOCasey: good call will replace with something else

gkatsev: Also, audioTracks isn't an array but an AudioTrackList object.

BrandonOCasey: @gkatsev in video.js thats the case, but this is a different list stored on Master Playlist Controller

src/hls-audio-track.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    let trackProps = {
+      autoselect: options.autoselect || false,
+      default: options.default || false,
+      withCredentials: options.withCredentials || false
+    };
+
+    /* eslint-disable no-loop-func */
+    for (let key in trackProps) {
```
dmlap: Given you know all the `trackProps` ahead of time, you probably should use a static defineProperties() call to do this instead of creating properties in a for-loop (and that gets rid of the eslint issue).

dmlap: Also, why do these need custom properties?

src/master-playlist-controller.js
gkatsev: Ah, carry on then.

src/hls-audio-track.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  /**
+   * remove a playlist loader from a track given the mediaGroup
+   *
+   * @param {String} mediaGroup the mediaGroup to remove
+   */
+  removeLoader(mediaGroup) {
+    delete this.mediaGroups[mediaGroup];
```
dmlap: I'm not 100% consistent about this myself but I think it's better to null the field out instead of using `delete`.

src/master-playlist-controller.js
```diff
@@ -321,37 +306,85 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+    this.audioTracks().forEach((t) => {
+      if (!track && t.enabled) {
+        track = t;
+      }
+    });
 
-    if (!master.mediaGroups.AUDIO[mediaGroupName]) {
-      videojs.log.warn('useAudio() was called with a mediaGroup ' + mediaGroupName +
-                       ' that does not exist in the master');
+    // called to early or no track is enabled
```
dmlap: > called __too__ early or no track is enabled

src/hls-audio-track.js
BrandonOCasey: I will implement the properties using Object.defineProperties as that sounds better.

They don't have to be custom properties, but I think that it makes it more consistent since they should never change, and since AudioTrack in video.js uses custom properties for everything already.

src/hls-audio-track.js
BrandonOCasey: nulling the field would not actually remove the key from the object. Unless I am thinking about this wrong.

src/master-playlist-controller.js
```diff
@@ -360,32 +393,16 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       this.audioSegmentLoader_.pause();
     }
 
-    let label = null;
+    // is the enabled track part of the main combined stream
+    // if it has a loader its not
```
dmlap: This comment is a bit hard for me to parse.

src/master-playlist-controller.js
```diff
@@ -321,37 +306,85 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
-    // manifest, not a master.
-    if (!media.attributes || !media.attributes.AUDIO ||
-        !master.mediaGroups || !master.mediaGroups.AUDIO) {
-      return;
-    }
-    let mediaGroupName = media.attributes.AUDIO;
+    return mediaGroup;
+  }
+
+  audioTracks() {
```
dmlap: This property seems like the sort of thing that is project-internal to me. Could we get away with just using `this.audioTracks_` directly?

src/videojs-contrib-hls.js
```diff
@@ -155,49 +178,64 @@ export default class HlsHandler extends Component {
...
+                `switch to ${error} Unfortunately this means we will have to ` +
+                `return you to the main track '${defaultTrack.label}'. Sorry!`;
+        revert = () => {
+          defaultTrack.enabled = true;
+          this.tech_.audioTracks().removeTrack(enabledTrack);
+        };
       }
+
+      // revert the audio info change on MSE
+      revert();
```
dmlap: Why not just run the revert() action inline in your conditional?

src/videojs-contrib-hls.js
```diff
@@ -215,6 +253,14 @@ export default class HlsHandler extends Component {
       this.masterPlaylistController_.mediaSource));
   }
 
+  audioTracks() {
+    return this.masterPlaylistController_.audioTracks();
+  }
+
+  activeAudioGroup() {
```
dmlap: Should these two be public? If so, they need docs. If not, they need `_`s.

src/hls-audio-track.js
dmlap: That's true but do you need the key to be removed from the object?

gkatsev: `delete` should be avoided as much as possible because it de-optimizes the objects associated with it in the engine, so, unless absolutely required, nulling them out would be best.

src/videojs-contrib-hls.js
BrandonOCasey: yeah that makes more sense

src/videojs-contrib-hls.js
BrandonOCasey: made them private

src/hls-audio-track.js
```diff
@@ -0,0 +1,114 @@
...
+    let trackProps = {
+      autoselect: options.autoselect || false,
+      default: options.default || false,
+      withCredentials: options.withCredentials || false
+    };
+
+    // set noop
+    let set = () => {};
+
+    Object.defineProperties(this, {
```
dmlap: I understand your point about consistency but I think it's more important to use the least-powerful tool that gets the job done. Custom properties imply extra functionality to anyone debugging and that would be misleading in this case.

src/videojs-contrib-hls.js
```diff
@@ -327,49 +350,56 @@ export default class HlsHandler extends Component {
...
-      let master = this.masterPlaylistController_.master();
-      let mediaGroups = master.mediaGroups;
-      let attributes = {
-        audio: {main: {default: true}}
-      };
-
-      if (!media.attributes) {
-        // source URL was playlist manifest, not master
-        // no audio tracks to add
+    this.masterPlaylistController_.on('audioinfo', (e) => {
```
dmlap: `audioinfo` will fire when there is no change in the audio settings? Seems like this could be managed at a lower level and remove the need for this objectChanged()

src/videojs-contrib-hls.js
```diff
@@ -394,6 +424,14 @@ export default class HlsHandler extends Component {
       this.masterPlaylistController_.mediaSource));
   }
 
+  audioTracks_() {
```
dmlap: What's the point of this method over accessing the property?

src/videojs-contrib-hls.js
BrandonOCasey: Originally it was at a lower level but it ended up with HLS trying to revert data on MediaSources which was a big jump. So it was determined that we should just fire 'audioinfo' every time we get it, and have the objects that care about the event act on it. In this case HLS.

src/videojs-contrib-hls.js
BrandonOCasey: we can probably just use the property

PR #662, opened 2016-04-21
force a specific tech when mode override is active

BrandonOCasey:
remove code for mode override in src (as its not possible)

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -208,14 +208,14 @@ Hls.isSupported = function() {
 const Component = videojs.getComponent('Component');
 
 export default class HlsHandler extends Component {
-  constructor(tech, options) {
+  constructor(source, tech, options) {
     super(tech);
-    let _player;
 
     // tech.player() is deprecated but setup a reference to HLS for
     // backwards-compatibility
     if (tech.options_ && tech.options_.playerId) {
-      _player = videojs(tech.options_.playerId);
+      let _player = videojs(tech.options_.playerId);
+
       if (!_player.hasOwnProperty('hls')) {
         Object.defineProperty(_player, 'hls', {
           get: () => {
@@ -226,17 +226,9 @@ export default class HlsHandler extends Component {
       }
     }
 
-    // TODO: when source handler options are possible (pull request is accepted
-    // into video.js) then we can add other options entries. Right now only
-    // mode/source will ever be passed as we pass them in
-    // HlsSourceHandler.handleSource()
-    // @link https://github.com/videojs/video.js/pull/3245
-    this.options_ = videojs.mergeOptions(options, videojs.options.hls || {});
-    // we cannot override the mode at this state, we have to use what was passed
-    // to us
-    this.options_.mode = options.mode;
+    this.options_ = videojs.mergeOptions(videojs.options.hls || {}, options.hls);
     this.tech_ = tech;
-    this.source_ = options.source;
+    this.source_ = source;
 
     // start playlist selection at a reasonable bandwidth for
     // broadband internet
@@ -441,12 +433,14 @@ const HlsSourceHandler = function(mode) {
     canHandleSource(srcObj) {
       // this forces video.js to skip this tech/mode if its not the one we have been
       // overriden to use, by returing that we cannot handle the source.
-      if (typeof videojs.options.hls.mode === 'string' && mode !== videojs.options.hls.mode) {
+      if (videojs.options.hls &&
+          videojs.options.hls.mode &&
+          videojs.options.hls.mode !== mode) {
         return false;
       }
       return HlsSourceHandler.canPlayType(srcObj.type);
     },
-    handleSource(source, tech) {
+    handleSource(source, tech, options) {
       if (mode === 'flash') {
         // We need to trigger this asynchronously to give others the chance
         // to bind to the event when a source is set at player creation
@@ -454,10 +448,9 @@ const HlsSourceHandler = function(mode) {
           tech.trigger('loadstart');
         }, 1);
       }
-      tech.hls = new HlsHandler(tech, {
-        source,
-        mode
-      });
+      let settings = videojs.mergeOptions(options, {hls: {mode}});
+
+      tech.hls = new HlsHandler(source, tech, settings);
       tech.hls.src(source.src);
       return tech.hls;
     },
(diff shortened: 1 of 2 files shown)


src/videojs-contrib-hls.js
```diff
@@ -273,6 +276,12 @@ export default class HlsHandler extends Component {
 const HlsSourceHandler = function(mode) {
   return {
     canHandleSource(srcObj) {
+      // this will video.js to to not use this tech/mode if its
+      // not the one we have been overriden to use. It does this by saying that
+      // the tech you don't want to use can't handle this source
```
imbcmdth: This comment needs some love: "this will video.js to to not use" and "overriden".

src/videojs-contrib-hls.js
```diff
@@ -87,6 +87,9 @@ export default class HlsHandler extends Component {
     // HlsSourceHandler.handleSource()
     // @link https://github.com/videojs/video.js/pull/3245
     this.options_ = videojs.mergeOptions(options, videojs.options.hls || {});
+    // we cannot override the mode at this state, we have to use what was passed
+    // to us
+    this.options_.mode = options.mode;
```
imbcmdth: `this.options_` is already built from `options`. I'm not sure what this line is accomplishing.

src/videojs-contrib-hls.js
```diff
@@ -273,6 +276,12 @@ export default class HlsHandler extends Component {
 const HlsSourceHandler = function(mode) {
   return {
     canHandleSource(srcObj) {
+      // this will video.js to to not use this tech/mode if its
+      // not the one we have been overriden to use. It does this by saying that
+      // the tech you don't want to use can't handle this source
+      if (typeof videojs.options.hls.mode === 'string' && mode !== videojs.options.hls.mode) {
```
imbcmdth: So `this.options_` is never used?

src/videojs-contrib-hls.js
BrandonOCasey: Well if we get here with a with `videojs.options.hls.mode = 'flash'` but the tech is currently html and we were passed `mode = 'html'` by the HlsSourceHandler a MediaSource error will be thrown, so we only ever want to use what HlsSourceHandler passes us.

src/videojs-contrib-hls.js
BrandonOCasey: well we don't have `this.options_` at this point and this.options_ is used in the `src()` function

src/videojs-contrib-hls.js
dmlap: It might be a good thing to throw a helpful error in that case. It seems to me this would only occur because of programmer misconfiguration and normally it's good to point that out during development.

src/videojs-contrib-hls.js
```diff
@@ -87,6 +87,9 @@ export default class HlsHandler extends Component {
     // HlsSourceHandler.handleSource()
     // @link https://github.com/videojs/video.js/pull/3245
     this.options_ = videojs.mergeOptions(options, videojs.options.hls || {});
```
dmlap: Looking over this, the order seems wrong. Object-level options should trump page-level and this looks like it does the opposite. Could that be what's making the code below hard?

src/videojs-contrib-hls.js
```diff
@@ -273,6 +276,11 @@ export default class HlsHandler extends Component {
 const HlsSourceHandler = function(mode) {
   return {
     canHandleSource(srcObj) {
+      // this forces video.js to skip this tech/mode if its not the one we have been
+      // overriden to use, by returing that we cannot handle the source.
+      if (typeof videojs.options.hls.mode === 'string' && mode !== videojs.options.hls.mode) {
```
dmlap: `videojs.options.hls` may not be defined.

dmlap: I think we should try to avoid registering the HlsSourceHandler if the mode setting prevents it from being useful.

test/videojs-contrib-hls.test.js
```diff
@@ -1201,36 +1201,6 @@ QUnit.test('if mode global option is used, mode is set to global option', functi
   videojs.options.hls = hlsOptions;
 });
 
-QUnit.test('if source option used, mode is set to the source option', function() {
```
dmlap: Not sure why these need to be deleted. Doesn't seem right to me.

src/videojs-contrib-hls.js
BrandonOCasey: well object level options don't really exist atm, unless that PR I mentioned was recently merged.

test/videojs-contrib-hls.test.js
BrandonOCasey: you cannot override mode as a source option, it won't work because the tech that is being used has to be the mode we are in. So if you are using HTML tech and you do `player.src({mode: 'flash'})` things would break because we can't use Flash mode with HTML tech. maybe there could be a way to do it in the future but I don't see it right now.

src/videojs-contrib-hls.js
BrandonOCasey: yeah thats probably a better option

dmlap: > This would restrict when the mode can be overridden.

I don't understand what "this" is. My point was you may be dereferencing `undefined` in this line.

BrandonOCasey: My comment was really talking about your second comment, but it was admittedly badly worded. Your first comment is correct and something that needs to be checked for.

Your second comment, where we avoid registering the HlsSourceHandler if the mode settings specifies not to, would force anyone who uses the project to declare mode overrides before we load the HLS script. This is due to the fact that we register HlsSourceHandlers in the main body of the script

src/videojs-contrib-hls.js
```diff
@@ -436,20 +431,26 @@ export default class HlsHandler extends Component {
 const HlsSourceHandler = function(mode) {
   return {
     canHandleSource(srcObj) {
```
dmlap: @gkatsev it seems like the SourceHandler options should be passed to this method as well. As it stands, we can't figure out a way to force HLS to use Flash (or MSE) at the player level. Instead we have to use the kind of ugly global method you can see below.

dmlap:
LGTM

src/videojs-contrib-hls.js
gkatsev: Should we just pass the options to each of the methods?

BrandonOCasey: I think we should yeah

PR #666, opened 2016-04-22
Remove two-way dependency between HlsHandler and MPC

dmlap:
Try to further untangle the relationship between the HlsHandler and the MasterPlaylistController by removing all references to the handler in the MPC. We're still passing the tech down to the MPC which is a backdoor to sneak a lot of shared state around but this clarifies the ownership relationship a bit.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -5,154 +5,7 @@ import videojs from 'video.js';
 
 // 5 minute blacklist
 const BLACKLIST_DURATION = 5 * 60 * 1000;
-
-// A fudge factor to apply to advertised playlist bitrates to account for
-// temporary flucations in client bandwidth
-const BANDWIDTH_VARIANCE = 1.2;
-
 let Hls;
-
-/**
- * Returns the CSS value for the specified property on an element
- * using `getComputedStyle`. Firefox has a long-standing issue where
- * getComputedStyle() may return null when running in an iframe with
- * `display: none`.
- * @see https://bugzilla.mozilla.org/show_bug.cgi?id=548397
- */
-const safeGetComputedStyle = function(el, property) {
-  let result;
-
-  if (!el) {
-    return '';
-  }
-
-  result = getComputedStyle(el);
-  if (!result) {
-    return '';
-  }
-
-  return result[property];
-};
-
-/**
- * Chooses the appropriate media playlist based on the current
- * bandwidth estimate and the player size.
- * @return the highest bitrate playlist less than the currently detected
- * bandwidth, accounting for some amount of bandwidth variance
- */
-const selectPlaylist = function() {
-  let effectiveBitrate;
-  let sortedPlaylists = this.playlists.master.playlists.slice();
-  let bandwidthPlaylists = [];
-  let now = +new Date();
-  let i;
-  let variant;
-  let bandwidthBestVariant;
-  let resolutionPlusOne;
-  let resolutionPlusOneAttribute;
-  let resolutionBestVariant;
-  let width;
-  let height;
-
-  sortedPlaylists.sort(Hls.comparePlaylistBandwidth);
-
-  // filter out any playlists that have been excluded due to
-  // incompatible configurations or playback errors
-  sortedPlaylists = sortedPlaylists.filter((localVariant) => {
-    if (typeof localVariant.excludeUntil !== 'undefined') {
-      return now >= localVariant.excludeUntil;
-    }
-    return true;
-  });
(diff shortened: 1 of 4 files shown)


src/master-playlist-controller.js
```diff
@@ -176,40 +56,45 @@ const parseCodecs = function(codecs) {
...
 export default class MasterPlaylistController extends videojs.EventTarget {
-  constructor({url, withCredentials, currentTimeFunc, mediaSourceMode, hlsHandler,
-    externHls}) {
+  constructor({
+    url,
+    withCredentials,
+    mode,
+    tech,
+    bandwidth,
+
```
gesinger: Is this extra line on purpose?

src/master-playlist-controller.js
```diff
@@ -363,9 +247,9 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     let label = null;
 
     // if no label was passed in we are switching to the currently enabled audio
-    for (let i = 0; i < this.hlsHandler.tech_.audioTracks().length; i++) {
-      if (this.hlsHandler.tech_.audioTracks()[i].enabled) {
-        label = this.hlsHandler.tech_.audioTracks()[i].label;
+    for (let i = 0; i < this.tech_.audioTracks().length; i++) {
```
gesinger: I know it's not really in scope, but it would be so much prettier with with find() and a polyfill for IE.

Otherwise, again out of scope, but it may read easier if we create a variable for this.tech_.audioTracks()

src/videojs-contrib-hls.js
```diff
@@ -139,13 +139,37 @@ export default class HlsHandler extends Component {
...
-      withCredentials: this.options_.withCredentials,
-      currentTimeFunc: this.tech_.currentTime.bind(this.tech_),
-      mediaSourceMode: this.options_.mode,
-      hlsHandler: this,
-      externHls: Hls
+    this.options_.url = this.source_.src;
+    this.options_.tech = this.tech_;
+    this.options_.externHls = Hls;
+    this.options_.bandwidth = this.bandwidth;
+    this.masterPlaylistController_ = new MasterPlaylistController(this.options_);
```
gesinger: It's minor, but passing the whole this.options_ object when creating MasterPlaylistController feels a bit strange to me. Maybe it all is needed right now for MasterPlaylistController, and in the future, if we're ever leaking more information than we should, we can just specify what we need to send then. But it just feels a bit odd.

src/videojs-contrib-hls.js
```diff
@@ -205,6 +229,12 @@ export default class HlsHandler extends Component {
       this.tech_.trigger('loadedmetadata');
     });
 
+    // the bandwidth of the primary segment loader is our best
+    // estimate of overall bandwidth
+    this.masterPlaylistController_.on('progress', function() {
```
gesinger: I think we're only triggering progress on the tech. That said, MPC should probably trigger it on itself too (or only trigger it on itself, and let this guy trigger it on the tech).

src/videojs-contrib-hls.js
```diff
@@ -139,13 +139,37 @@ export default class HlsHandler extends Component {
...
+    // controller. Using a custom property for backwards compatibility
+    // with < v2
+    Object.defineProperties(this, {
+      selectPlaylist: {
+        get() {
+          return this.masterPlaylistController_.selectPlaylist
+            .bind(this.masterPlaylistController_);
+        },
+        set(selectPlaylist) {
+          this.masterPlaylistController_.selectPlaylist = selectPlaylist;
```
imbcmdth: Any function passed in to act as `selectPlaylist` expects `this` to be equal to the HlsHandler object and have access to it's non-underlined functions specifically.

src/master-playlist-controller.js
BrandonOCasey: We probably don't want to do it as it will be changed in #661

src/master-playlist-controller.js
dmlap: Not really. I was using it as a separator for "dependencies that need to be removed" :)

src/videojs-contrib-hls.js
dmlap: I can understand that feeling. I was thinking the MPC is a "project private" sort of thing so it didn't need its own set of configuration options.

src/videojs-contrib-hls.js
dmlap: Good catch!

gesinger:
LGTM

PR #668, opened 2016-04-25
Spoof an unofficially supported codec

tenacex:
Streams were failing for numerous reasons, one of which being the video codecs not being supported by chrome. This spoofs the codec to enable playback.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -735,7 +735,13 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     // if the codecs were explicitly specified, pass them along to the
     // source buffer
     if (media.attributes && media.attributes.CODECS) {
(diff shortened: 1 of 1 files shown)


src/master-playlist-controller.js
```diff
@@ -735,7 +735,13 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     // if the codecs were explicitly specified, pass them along to the
     // source buffer
     if (media.attributes && media.attributes.CODECS) {
-      mimeType += '; codecs="' + media.attributes.CODECS + '"';
+
+      if (media.attributes.CODECS === 'avc1.77.30, mp4a.40.2') {
```
dmlap: This is a very (too) specific spoof. Why is `avc1.77.30` unique in triggering this? Also, this kind of browser/MSE compatibility issue should probably be fixed in videojs-contrib-media-sources.

PR #655, opened 2016-04-13
Added tests for Jons retry logic additions

tenacex:
Added some tests so that Jon's changes can get merged.

PR #670, opened 2016-04-26
Added tests for Jons retry logic additions

tenacex:
The test fails without the retry additions.


dmlap:
Re. #655: sorry, didn't see that was a merge into the blacklist-manifest-stuck branch. Looks good to merge into blacklist-manifest-stuck to me. @imbcmdth?

PR #672, opened 2016-04-26
HLS no longer fires its own loadedmetadata event

BrandonOCasey:
simplified SegmentLoader options in MasterPlaylistController
added bandwidth optional to SegmentLoader options
updated unit tests to reflect havePlayed_ function
verify that we don't download more than 1 segment if we have not played yet
added a unit test to verify metadata preload

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -64,27 +64,21 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     // load the media source into the player
     this.mediaSource.addEventListener('sourceopen', this.handleSourceOpen_.bind(this));
 
-    // combined audio/video or just video when alternate audio track is selected
-    this.mainSegmentLoader_ = new SegmentLoader({
+    let segmentLoaderOptions = {
       mediaSource: this.mediaSource,
       currentTime: this.tech_.currentTime.bind(this.tech_),
       withCredentials: this.withCredentials,
       seekable: () => this.seekable(),
       seeking: () => this.tech_.seeking(),
-      setCurrentTime: (a) => this.setCurrentTime(a)
-    });
-    // pass along the starting bandwidth estimate
-    this.mainSegmentLoader_.bandwidth = bandwidth;
+      setCurrentTime: (a) => this.setCurrentTime(a),
+      hasPlayed: () => this.tech_.played().length !== 0,
+      bandwidth
+    };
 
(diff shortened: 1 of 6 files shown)


src/master-playlist-controller.js
```diff
@@ -85,16 +79,15 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+      if (media.endList && this.tech_.preload() !== 'none') {
         this.mainSegmentLoader_.playlist(media);
         this.mainSegmentLoader_.expired(this.masterPlaylistLoader_.expired_);
         this.mainSegmentLoader_.load();
       }
 
       this.setupSourceBuffer_();
       this.setupFirstPlay();
+      this.useAudio();
       this.trigger('loadedmetadata');
```
imbcmdth: Is there any reason to still emit this? I see you removed the `loadedmetadata` handler from `videojs-contrib-hls.js` below.

test/videojs-contrib-hls.test.js
```diff
@@ -952,7 +952,7 @@ QUnit.test('fire loadedmetadata once we successfully load a playlist', function(
   });
   openMediaSource(this.player, this.clock);
   this.player.tech_.hls.bandwidth = 20000;
-  this.player.on('loadedmetadata', function() {
+  this.player.tech_.hls.masterPlaylistController_.on('loadedmetadata', function() {
```
imbcmdth: Hmm, this test might need to be rethought completely.. It really isn't testing anything now but I'm not sure if it is possible to test that `loadedmetadata` is being fired by the video element without appending some real data.

BrandonOCasey: Yeah I don't think its possible to test that loaded metadata fires without using real data, and I am not sure that a unit test should cover that scenario.

src/master-playlist-controller.js
BrandonOCasey: Not a good reason, but I kept it around so that I could listen to it for one of the tests in `videojs-contrib-hls.test.js`. We can probably remove it

imbcmdth: I would remove this then and the test below then. They seem superfluous now anyway.

BrandonOCasey:
@imbcmdth changes made, ready for a re-review

imbcmdth:
LGTM

PR #673, opened 2016-04-27
a fix so that SegmentLoader and HLS share GOAL_BUFFER_LENGTH

BrandonOCasey:
updated SegmentLoader test to use this value
added tests to verify that GOAL_BUFFER_LENGTH is shared

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -15,7 +15,7 @@ import {Decrypter} from './decrypter';
 const CHECK_BUFFER_DELAY = 500;
 
 // the desired length of video to maintain in the buffer, in seconds
-export const GOAL_BUFFER_LENGTH = 30;
+let GOAL_BUFFER_LENGTH = 30;
 
 /**
  * Updates segment with information about its end-point in time and, optionally,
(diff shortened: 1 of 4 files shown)


src/videojs-contrib-hls.js
```diff
@@ -48,7 +49,10 @@ const Hls = {
 };
 
 // the desired length of video to maintain in the buffer, in seconds
-Hls.GOAL_BUFFER_LENGTH = 30;
+Object.defineProperty(Hls, 'GOAL_BUFFER_LENGTH', {
+  get: () => SegmentLoader.GOAL_BUFFER_LENGTH,
+  set: (v) => SegmentLoader.GOAL_BUFFER_LENGTH = v
+});
```
imbcmdth: A getter calling a getter? OMG!

gkatsev: We heard you like getters so we put a getter inside your getter so you can get while you get

src/segment-loader.js
```diff
@@ -155,6 +155,14 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
   }
 
+  static get GOAL_BUFFER_LENGTH() {
+    return GOAL_BUFFER_LENGTH;
+  }
+
+  static set GOAL_BUFFER_LENGTH(v) {
+    GOAL_BUFFER_LENGTH = v;
+  }
```
imbcmdth: There is really little reason to make this a getter/setter combo. We tend to avoid them except where required for compatibility reasons.

src/videojs-contrib-hls.js
imbcmdth: I am thinking that we should deprecate this property (log a warning) and prefer a GOAL_BUFFER_LENGTH being set via options.

BrandonOCasey: sounds like a better solution to me

BrandonOCasey:
I am going to close this PR, as I think we are going to go another direction with this

PR #674, opened 2016-04-27
added doc comments to the whole project

BrandonOCasey:
fixed incorrect doc comments
fixed some bad formatting
fixed some weird code


imbcmdth:
NICE. 🐶

src/decrypter/decrypter.js
```diff
@@ -135,6 +142,20 @@ export class Decrypter {
...
   }
+
+  /**
+   * a getter for step the maximum number of bytes to process at one time
+   *
+   * @return {Number} the value of step 32000
+   */
+  static get STEP() {
+    // 4 * 8000;
+    return 32000;
```
dmlap: Why is this a getter?

src/master-playlist-controller.js
```diff
@@ -29,6 +32,15 @@ const parseCodecs = function(codecs) {
   return result;
 };
 
+/**
+ * the master playlist controller controller all interactons
```
dmlap: > controller controller

Not a blocker.

src/decrypter/decrypter.js
BrandonOCasey: I don't think it should be, I am going to change it back to what it was before

PR #681, opened 2016-05-06
Multiple audio track support

imbcmdth:
* Separated segment loading logic into a reusable piece of functionality so that we can have more than one segment loader and they can manage the segment fetch behavior unique to each playlist they are loading from
* Introduced the MasterPlaylistController to coordinate the loading of the master playlist and separate the behavior of the player from videojs-contrib-hls.js leaving the latter to be the glue between HLS and the video element's events
* Added the SourceUpdater to manage the asynchronous bahavior of SourceBuffers and MediaSource objects so that we can treat it as a non-blocking work queue
* Added parsing for MediaGroups in master playlists
* Added support for AudioTrackList objects and events


dmlap:
I'm not re-reviewing this sucker :) LGTM!

PR #682, opened 2016-05-06
Flash live

imbcmdth:
Remove a work-around that was in-place for the broken implementation of MSE events in Firefox < 45 and fix a bug that was keeping Flash from seeking to the live-point in a live HLS stream.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -581,28 +581,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
         return audioSeekable;
       }
     }
-/*
-    // if the seekable start is zero, it may be because the player has
-    // been paused for a long time and stopped buffering. in that case,
-    // fall back to the playlist loader's running estimate of expired
-    // time
-    if (mainSeekable.start(0) === 0) {
-      mainSeekable = videojs.createTimeRanges([[this.masterPlaylistLoader_.expired_,
-                                                this.masterPlaylistLoader_.expired_ +
-                                                  mainSeekable.end(0)]]);
-    }
-    if (!audioSeekable) {
-      // seekable has been calculated based on buffering video data so it
-      // can be returned directly
-      return mainSeekable;
-    }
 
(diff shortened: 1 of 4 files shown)


src/master-playlist-controller.js
```diff
@@ -388,33 +389,32 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+        // 3) the player is not paused
+        !this.tech_.paused() &&
 
-      this.load();
+        // 4) the player has not started playing
+        !this.hasPlayed_) {
 
-      // 4) the video element or flash player is in a readyState of
-      // at least HAVE_FUTURE_DATA
-      if (this.tech_.readyState() >= 1) {
```
gesinger: I think I remember that the reason we had the readyState check was because firefox expected a segment to be loaded before a certain action took place. Was it loadedmetadata firing (which I know we handle ourselves now), or was it the ability to do a seek?

src/playlist.js
```diff
@@ -100,13 +100,15 @@ const forwardDuration = function(playlist, endSequence) {
   * playlist. The duration of a subinterval of the available segments
   * may be calculated by specifying an end index.
   *
-  * @param {Object} playlist a media playlist object
-  * @param {Number=} endSequence an exclusive upper boundary
-  * for the playlist. Defaults to playlist length.
-  * @return {Number} the duration between the first available segment
+  * @param playlist {object} a media playlist object
```
BrandonOCasey: doc comments appear to be backwards (`endSequence {number}` should be `{number} endSequence`) also you can signal optional by putting and equal sign after the type decleration ie `{Number=}`

src/master-playlist-controller.js
imbcmdth: That's exactly the point of this removal. It was an older version of FF with this limitation.

src/playlist.js
imbcmdth: Dammit!

PR #684, opened 2016-05-06
Add support for every HLS mime type possible

imbcmdth:
Include every possible permutation of HLS mime-types in both supports NativeHls and canPlayType. Also canPlayType now returns a proper string value to the browser instead of a boolean.

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -225,21 +225,21 @@ Hls.supportsNativeHls = (function() {
   // HLS manifests can go by many mime-types
   let canPlay = [
     // Apple santioned
-    video.canPlayType('application/vnd.apple.mpegURL'),
+    'application/vnd.apple.mpegurl',
     // Apple sanctioned for backwards compatibility
-    video.canPlayType('audio/mpegURL'),
+    'audio/mpegurl',
     // Very common
-    video.canPlayType('audio/x-mpegURL'),
+    'audio/x-mpegurl',
     // Very common
-    video.canPlayType('application/x-mpegURL'),
+    'application/x-mpegurl',
     // Included for completeness
-    video.canPlayType('video/x-mpegURL'),
-    video.canPlayType('video/mpegURL'),
-    video.canPlayType('application/mpegURL')
+    'video/x-mpegurl',
+    'video/mpegurl',
+    'application/mpegurl'
   ];
 
   return canPlay.some(function(canItPlay) {
-    return (/maybe|probably/i).test(canItPlay);
+    return (/maybe|probably/i).test(video.canPlayType(canItPlay));
   });
 }());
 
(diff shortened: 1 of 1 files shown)


src/videojs-contrib-hls.js
```diff
@@ -216,18 +216,31 @@ Hls.canPlaySource = function() {
...
+    // Very common
+    video.canPlayType('application/x-mpegURL'),
+    // Included for completeness
+    video.canPlayType('video/x-mpegURL'),
+    video.canPlayType('video/mpegURL'),
+    video.canPlayType('application/mpegURL')
+  ];
+
+  return canPlay.some(function(canItPlay) {
+    return (/maybe|probably/i).test(canItPlay);
```
gesinger: Since we just want to test that they are not an empty string, would it be simpler to remove the array and the some call and just do a return with ors between each?

return video.canPlayType... || video.canPlayType... || ...

Or are we protecting against other potential values of canPlayType?

gesinger:
Maybe it isn't worth it for this, but we could have some tests for those formats.

src/videojs-contrib-hls.js
```diff
@@ -604,11 +620,11 @@ Hls.comparePlaylistResolution = function(left, right) {
...
 HlsSourceHandler.canPlayType = function(type) {
-  let mpegurlRE = /^application\/(?:x-|vnd\.apple\.)mpegurl/i;
+  let mpegurlRE = /^(audio|video|application)\/(x-|vnd\.apple\.)?mpegurl/i;
 
   // favor native HLS support if it's available
   if (Hls.supportsNativeHls) {
-    return false;
+    return '';
   }
   return mpegurlRE.test(type);
```
BrandonOCasey: do we want to return maybe/probably here rather than a boolean?

BrandonOCasey: Maybe to prevent confusion we should still return false in this function, since empty string is going to be used as false in this case, and since we return a boolean below anyway.

src/videojs-contrib-hls.js
BrandonOCasey: maybe to cut down on typing put video.canPlayType in here? doesn't really matter though

gkatsev: or `return canPlay.some(Boolean)` since `maybe` and `probably` would be true and an empty string would be `false`

src/videojs-contrib-hls.js
```diff
@@ -216,18 +216,31 @@ Hls.canPlaySource = function() {
...
   if (!videojs.getComponent('Html5').isSupported()) {
     return false;
   }
 
-  xMpegUrl = video.canPlayType('application/x-mpegURL');
-  vndMpeg = video.canPlayType('application/vnd.apple.mpegURL');
-  return (/probably|maybe/).test(xMpegUrl) ||
-    (/probably|maybe/).test(vndMpeg);
+  // HLS manifests can go by many mime-types
+  let canPlay = [
```
gkatsev: why not just have this be a list of strings and then in the return statement do a `map` before the `some`?

BrandonOCasey:
LGTM

src/videojs-contrib-hls.js
imbcmdth: I kind of, sort of did that in a later commit.

PR #685, opened 2016-05-09
Added bandwidth stats to HLS

BrandonOCasey:
Added bandwidth stat assertions to existing tests

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -86,6 +86,9 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       throw new Error('A non-empty playlist URL is required');
     }
 
+    this.mainSegmentLoader_.on('stat', (e) => this.trigger(e));
+    this.audioSegmentLoader_.on('stat', (e) => this.trigger(e));
+
     this.masterPlaylistLoader_ = new PlaylistLoader(url, this.hls_, this.withCredentials);
 
     this.masterPlaylistLoader_.on('loadedmetadata', () => {
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -86,9 +86,6 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       throw new Error('A non-empty playlist URL is required');
     }
 
-    this.mainSegmentLoader_.on('stat', (e) => this.trigger(e));
-    this.audioSegmentLoader_.on('stat', (e) => this.trigger(e));
-
     this.masterPlaylistLoader_ = new PlaylistLoader(url, this.hls_, this.withCredentials);
 
     this.masterPlaylistLoader_.on('loadedmetadata', () => {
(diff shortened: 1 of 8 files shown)


src/videojs-contrib-hls.js
```diff
@@ -369,10 +369,12 @@ class HlsHandler extends Component {
       },
       bandwidth: {
         get() {
-          return this.masterPlaylistController_.mainSegmentLoader_.bandwidth;
+          Stats.bandwidth = this.masterPlaylistController_.mainSegmentLoader_.bandwidth;
+          return Stats.bandwidth;
```
tenacex: Does the getter really need to set this? Seems a bit odd to me.

tenacex: Besides that it looks good

src/videojs-contrib-hls.js
```diff
@@ -14,6 +14,7 @@ import {MediaSource, URL} from 'videojs-contrib-media-sources';
 import m3u8 from './m3u8';
 import videojs from 'video.js';
 import MasterPlaylistController from './master-playlist-controller';
+import Stats from './stats';
```
imbcmdth: I believe this will import the same object every time so that it is shared between videos. Meaning that the stats will increment forever even when we change sources.

src/segment-loader.js
```diff
@@ -635,6 +635,31 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          amount: request.bytesReceived || 0
+        }
+      });
+
+      // if we get this far there are no errors,
+      // increment the stat for media requests
+      this.trigger({
+        type: 'stat',
+        data: {
+          name: 'numberOfMediaRequests',
```
dmlap: Our standard pattern in video.js has been to fire "simple events" when interesting state changes and provide accessors to query state. For consistency, I think we should continue to follow that pattern here.

dmlap: We could also look at how [MSE exposed video quality](https://w3c.github.io/media-source/#htmlvideoelement-extensions) for inspiration on how to expose this info in a way that fits with the rest of the platform.

BrandonOCasey:
@dmlap let me see if I can explain how I interpreted what you said. You tell me if it is inline with what we want to do. 

We want the stat events that I currently have implemented to keep being fired. Rather than having the stat event carry data we store the data on SegmentLoader. The object that is listening for the stat event on SegmentLoader will grab the stat from the instance of segment loader that fired the stat event.

or

Get rid of all events, store the stats on each segment loader. add a method on MasterPlaylistController to get all of the combined stats for SegmentLoaders. Call that function on HLS when we need to get stats for analytics.

src/master-playlist-controller.js
```diff
@@ -172,6 +172,20 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     this.masterPlaylistLoader_.load();
   }
 
+
+  getMediaRequests_() {
```
dmlap: Naming nitpick (sorry!): video.js doesn't typically use the `get` prefix for getters. That is, `s/getMediaRequests/mediaRequests/g`.

BrandonOCasey:
@dmlap I rebased and removed the get prefix from those functions

test/master-playlist-controller.test.js
```diff
@@ -163,6 +177,13 @@ QUnit.test('if buffered, will request second segment byte range', function() {
   this.masterPlaylistController.mediaSource.sourceBuffers[0].trigger('updateend');
   this.clock.tick(10 * 1000);
   QUnit.equal(this.requests[2].headers.Range, 'bytes=1823412-2299991');
+
+  // verify stats
+  QUnit.equal(this.player.tech_.hls.stats.bandwidth, Infinity, 'Live stream');
```
dmlap: *Bandwidth* is Infinity?

dmlap:
LGTM

PR #676, opened 2016-04-29
add tests for all options in HLS

BrandonOCasey:
add configuration heirarchy testing
deprecate HLS.GOAL_BUFFER_LENGTH

this PR relies on an as-of-yet unreleased version of video.js (since sourceHandler options are not yet released, but they are in master)

PR #686, opened 2016-05-09
Add unit tests for HLS configurations and fix GOAL_BUFFER_LENGTH

BrandonOCasey:
add tests for all options in HLS
add configuration heirarchy testing
deprecate HLS.GOAL_BUFFER_LENGTH
bump the video.js version to one that supports source handler options
Hls.GOAL_BUFFER_LENGTH is settable to a number greater than 0 with a warning
use Hls.GOAL_BUFFER_LENGTH if it exists
added a file to keep track of GOAL_BUFFER_LENGTH across code

this PR was created because #676 is against development which is no longer in use

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -6,12 +6,10 @@ import {getMediaIndexForTime_ as getMediaIndexForTime, duration} from './playlis
 import videojs from 'video.js';
 import SourceUpdater from './source-updater';
 import {Decrypter} from './decrypter';
+import Config from './config';
 
 // in ms
 const CHECK_BUFFER_DELAY = 500;
-
-// the desired length of video to maintain in the buffer, in seconds
-export const GOAL_BUFFER_LENGTH = 30;
 
 /**
  * Updates segment with information about its end-point in time and, optionally,
@@ -78,8 +76,9 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex, currentB
     (appendedLastSegment || bufferedToEnd);
 };
 
-/*  Turns segment byterange into a string suitable for use in
- *  HTTP Range requests
+/**
+ * Turns segment byterange into a string suitable for use in
+ * HTTP Range requests
  */
 const byterangeStr = function(byterange) {
   let byterangeStart;
@@ -92,7 +91,8 @@ const byterangeStr = function(byterange) {
   return 'bytes=' + byterangeStart + '-' + byterangeEnd;
 };
 
-/*  Defines headers for use in the xhr request for a particular segment.
+/**
+ * Defines headers for use in the xhr request for a particular segment.
  */
 const segmentXhrHeaders = function(segment) {
   let headers = {};
@@ -385,7 +385,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
       // if there is plenty of content buffered, and the video has
       // been played before relax for awhile
-      if (this.hasPlayed_() && bufferedTime >= GOAL_BUFFER_LENGTH) {
+      if (this.hasPlayed_() && bufferedTime >= Config.GOAL_BUFFER_LENGTH) {
         return null;
       }
       mediaIndex = getMediaIndexForTime(playlist,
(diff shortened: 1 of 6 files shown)


src/videojs-contrib-hls.js
```diff
@@ -52,8 +53,21 @@ const Hls = {
...
-// the desired length of video to maintain in the buffer, in seconds
-Hls.GOAL_BUFFER_LENGTH = 30;
+Object.defineProperty(Hls, 'GOAL_BUFFER_LENGTH', {
+  get() {
+    videojs.log.warn('Hls.GOAL_BUFFER_LENGTH is deprecated and should not be used');
+    return Config.GOAL_BUFFER_LENGTH;
+  },
+  set(v) {
+    videojs.log.warn('Hls.GOAL_BUFFER_LENGTH is deprecated and should not be used');
+    if (typeof v !== 'number' && v > 0) {
```
tenacex: Wouldn't this be 
```javascript
if (typeof v !== 'number' || v <= 0)
```
?

BrandonOCasey: good catch

tenacex: Perhaps a test that would have caught this.

tenacex: Besides that LGTM

src/videojs-contrib-hls.js
```diff
@@ -52,8 +53,21 @@ const Hls = {
   xhr: xhrFactory()
 };
 
-// the desired length of video to maintain in the buffer, in seconds
-Hls.GOAL_BUFFER_LENGTH = 30;
+Object.defineProperty(Hls, 'GOAL_BUFFER_LENGTH', {
+  get() {
+    videojs.log.warn('Hls.GOAL_BUFFER_LENGTH is deprecated and should not be used');
```
imbcmdth: I don't think it should be deprecated if we don't provide any other way of setting. It's UNSAFE but not necessarily deprecated!

src/videojs-contrib-hls.js
```diff
@@ -52,8 +53,21 @@ const Hls = {
...
+    return Config.GOAL_BUFFER_LENGTH;
+  },
+  set(v) {
+    videojs.log.warn('Hls.GOAL_BUFFER_LENGTH is deprecated and should not be used');
```
imbcmdth: I don't think it should be deprecated if we don't provide any other way of setting. It's UNSAFE but not necessarily deprecated!

src/videojs-contrib-hls.js
```diff
@@ -325,6 +338,24 @@ class HlsHandler extends Component {
     this.on(this.tech_, 'play', this.play);
   }
 
+  setOptions_() {
+    // defaults
+    this.options_.withCredentials = this.options_.withCredentials || false;
+
+    // start playlist selection at a reasonable bandwidth for
+    // broadband internet
+    // 0.5 Mbps
```
imbcmdth: This has always bothered me.. It's 0.5**MB/s** not **mbps** - I know you didn't write this comment but I would be very appreciative if you could fix it!

src/segment-loader.js
```diff
@@ -6,13 +6,11 @@ import {getMediaIndexForTime_ as getMediaIndexForTime, duration} from './playlis
 import videojs from 'video.js';
 import SourceUpdater from './source-updater';
 import {Decrypter} from './decrypter';
+import Config from './config';
```
imbcmdth: `Config` will be the same object every time so that each source will start off with the config of it's predecessor instead of the default config.

BrandonOCasey: the same is true for any other global configuration though, I could see setting GOAL_BUFFER_LENGTH back to default on HLS dispose but on src change I think that global options should remain at what it was set to. We basically need HLS and SegmentLoader to share the value somehow so I went with this as it did not seem like we wanted to use crazy getters and setters.

BrandonOCasey:
the travis build is stalling after unit tests pass???

PR #687, opened 2016-05-09
run browserify-shim transform on watch:js:browserify

BrandonOCasey:


gkatsev:
Must've missed that one in https://github.com/videojs/videojs-contrib-hls/pull/675
can you double check and make sure that's the only one left over?

BrandonOCasey:
will do

BrandonOCasey:
@gkatsev seems like the only one that was missed was the one for `watch:js:browserify`

gkatsev:
LGTM

PR #689, opened 2016-05-09
updated playlist-loader to have better es6 syntax

BrandonOCasey:
use this rather than loader
convert some previously private variables to 'private' instance variables
#557 but against master and rebased

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -11,20 +11,20 @@ import Stream from './stream';
 import m3u8 from './m3u8';
 
 /**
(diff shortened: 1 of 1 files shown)


gesinger:
Thanks @BrandonOCasey , and we'll definitely be trying to move towards a cleaner playlist-loader, however, with the changes made to playlist-loader since this PR, it may be easier to to start anew than to rebase, and if we do move towards better syntax, we'll probably be better off doing it in pieces and trying to make things more functional.

PR #697, opened 2016-05-12
Intelligent seeking over un-closeable buffer gaps

tenacex:
Todo:

Adaptive seeking currently only works if we play into an unbuffered region. Ultimately, I would like to enable adaptive seeking when you seek into an unbuffered region as well. Due to Firefox and Chrome having different events for similar situations, I haven't figured out how to do this yet.


src/GapSkipper.js
```diff
@@ -0,0 +1,85 @@
+import videojs from 'video.js';
```
BrandonOCasey: I would rename this file gap-skipper to be more in line with the other files in the src folder.

BrandonOCasey: I also think we wanted to use this more like a plugin for video.js https://github.com/videojs/video.js/blob/master/docs/guides/plugins.md

tenacex: I'll change it right now

src/videojs-contrib-hls.js
```diff
@@ -272,7 +272,8 @@ class HlsHandler extends Component {
     // backwards-compatibility
     if (tech.options_ && tech.options_.playerId) {
       let _player = videojs(tech.options_.playerId);
-
+      videojs.plugin('gapSkipper', gapSkipper);
```
BrandonOCasey: I would register the plugin below, where we do videojs.registerComponent

src/videojs-contrib-hls.js
```diff
@@ -272,7 +272,8 @@ class HlsHandler extends Component {
...
+      _player.gapSkipper({tech:tech});
```
BrandonOCasey: gapskipper should be an instance to player so you can probably just do `this.tech_` inside of it

src/videojs-contrib-hls.js
```diff
@@ -286,6 +287,8 @@ class HlsHandler extends Component {
     this.options_ = videojs.mergeOptions(videojs.options.hls || {}, options.hls);
     this.tech_ = tech;
     this.source_ = source;
+    // register gapSKipper plugin
```
BrandonOCasey: comment is no longer needed

tenacex: removed

src/videojs-contrib-hls.js
tenacex: I tried that originally, it didn't work. Not sure why its undefined

src/GapSkipper.js
```diff
@@ -0,0 +1,85 @@
...
+  // Listen for waiting when finished seeking
+  player.on('seeked', function() {
+    seeking = false;
+  });
+
+  tech.on('playing', function() {
+    tech.on('waiting', adaptiveSeeking);
+  });
+
+  tech.on('adaptive-seeking', adaptiveSeeking);
```
BrandonOCasey: where does the 'adaptive-seeking` event come from?

src/GapSkipper.js
```diff
@@ -0,0 +1,85 @@
...
+
+  tech.on('error', function() {
```
BrandonOCasey: player should emit all tech events so you probably don't even need tech

BrandonOCasey:
`GapSkipper.js` should probably be `gap-skipper.js` to fit in with the common file names of the files in this project

src/videojs-contrib-hls.js
```diff
@@ -273,6 +273,8 @@ class HlsHandler extends Component {
     if (tech.options_ && tech.options_.playerId) {
       let _player = videojs(tech.options_.playerId);
 
+      videojs.plugin('gapSkipper', gapSkipper);
```
BrandonOCasey: You should be able to register this at the bottom where we do videojs.registerComponent

src/GapSkipper.js
tenacex: It comes from the custom chrome waiting event

src/videojs-contrib-hls.js
tenacex: changed.

src/gap-skipper.js
```diff
@@ -0,0 +1,85 @@
...
+let timer = null;
+let playerState;
+let lastRecordedTime;
+let adaptiveSeeking;
+let player;
+let tech;
+let gapSkipper;
+
+gapSkipper = function(options) {
+  player = this;
```
BrandonOCasey: we should probably make this a class/object rather than using module local variables

src/gap-skipper.js
```diff
@@ -0,0 +1,85 @@
...
+      return;
+    }
+    let currentTime = player.currentTime();
+
+    if (consecutiveUpdates === 5 && currentTime === lastRecordedTime) {
+      // trigger waiting
+      if (playerState !== 'waiting') {
+        consecutiveUpdates = 0;
+        playerState = 'waiting';
+        tech.trigger('adaptive-seeking');
```
BrandonOCasey: we should probably trigger this on the new gap-skipper object rather than on the tech directly

src/gap-skipper.js
```diff
@@ -0,0 +1,85 @@
...
+  player.on('seeking', function() {
+    seeking = true;
+  });
+
+  // Listen for waiting when finished seeking
+  player.on('seeked', function() {
+    seeking = false;
+  });
+
+  tech.on('playing', function() {
```
BrandonOCasey: player should fire this event as well we probably don't even need to use tech in gap-skipper

src/gap-skipper.js
```diff
@@ -0,0 +1,85 @@
...
+    tech.on('waiting', adaptiveSeeking);
+  });
+
+  tech.on('adaptive-seeking', adaptiveSeeking);
+
+  tech.on('error', function() {
```
BrandonOCasey: player should fire this event as well so that we can remove tech usage in gap-skipper

src/videojs-contrib-hls.js
```diff
@@ -273,6 +273,7 @@ class HlsHandler extends Component {
     if (tech.options_ && tech.options_.playerId) {
       let _player = videojs(tech.options_.playerId);
 
+      _player.gapSkipper({tech});
```
BrandonOCasey: we should move this line outside of this if check, maybe even into the src function of HlsHandler

test/videojs-contrib-hls.test.js
```diff
@@ -2121,6 +2121,58 @@ QUnit.test('Allows overriding the global beforeRequest function', function() {
...
+  this.player.src({
+    src: 'master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+
+  this.player.tech_.triggerReady();
+  this.clock.tick(1);
+  standardXHRResponse(this.requests.shift());
+  openMediaSource(this.player, this.clock);
+  this.player.tech_.trigger('play');
```
BrandonOCasey: it would probably be good to have a file just for gap skipper

src/videojs-contrib-hls.js
tenacex: changed.

src/GapSkipper.js
tenacex: I changed it to only listen to player events but it didn't work, going to leave it the way it is now, but I will investigate why it isn't working

src/gap-skipper.js
tenacex: aren't plugins always functions?

src/videojs-contrib-hls.js
tenacex: It is the only place in the project where we have a guaranteed reference to player, and if I moved it to the src function, I wouldn't have access to the tech.

src/gap-skipper.js
tenacex: Still looking into why it doesn't re-emit these events

dmlap:
@tenacex @imbcmdth The term "adaptive seeking" is jargon-- it's not descriptive enough to be clear to someone who hasn't been intimately involved in this project. Jargon makes it harder for new people to figure out what's going on and help out. It's fine if you'd like to refer to this technique as adaptive seeking among yourselves but we should use more accessible terminology in the code, even if it means we have to be more verbose.

src/gap-skipper.js
BrandonOCasey: make the plugin a function that uses a gap-skipper object

tenacex: Done

src/gap-skipper.js
```diff
@@ -0,0 +1,88 @@
...
+
+    seekingObject.player.on('error', function() {
+      if (seekingObject.timer) {
+        clearTimeout(seekingObject.timer);
+      }
+    });
+  });
+};
+
+adaptiveSeeking = function() {
```
BrandonOCasey: you may just want to make this a method of seeking object so that you don't have to use a global reference

src/gap-skipper.js
```diff
@@ -0,0 +1,88 @@
+import Ranges from './ranges';
+
+let adaptiveSeeking;
+let gapSkipper;
+let seekingObject;
+
+gapSkipper = function(options) {
```
BrandonOCasey: this should probably be a const rather than being defined above, also I don't think it takes any options anymore. 
ex: `const gapSkipper = function() {`. The same goes for `adaptiveSeeking`

src/videojs-contrib-hls.js
```diff
@@ -273,6 +273,7 @@ class HlsHandler extends Component {
     if (tech.options_ && tech.options_.playerId) {
       let _player = videojs(tech.options_.playerId);
 
+      _player.gapSkipper({});
```
BrandonOCasey: I would add this to the top or bottom of the constructor rather than in this if statement, as it no longer needs to be here. You also don't need to pass it an empty object.

test/gap-skipper.test.js
```diff
@@ -0,0 +1,157 @@
+/* eslint-disable max-len */
```
BrandonOCasey: this line needs to be removed

src/gap-skipper.js
```diff
@@ -0,0 +1,88 @@
...
+  this.ready(() => {
+    seekingObject = { player: this,
```
BrandonOCasey: I wouldn't store a reference to player or tech here I would just use them directly.

src/gap-skipper.js
```diff
@@ -0,0 +1,88 @@
...
+    // Listen for waiting when finished seeking
+    seekingObject.player.on('seeked', function() {
+      seekingObject.seeking = false;
+    });
+
+    seekingObject.player.on('playing', function() {
+      seekingObject.player.on('waiting', adaptiveSeeking);
+    });
+
+    seekingObject.tech.on('adaptive-seeking', adaptiveSeeking);
```
BrandonOCasey: I think that you should make `seekingObject` extend `videojs.EventTarget` so that it can fire this event itself rather than on the tech

BrandonOCasey:
@dmlap any suggestions on a name other than adaptive seeking? 

My idea is to turn the `adaptive-seeking` event into a `playbackstalled` event. Then maybe just call the `adaptiveSeeking` function `gapFound` or `skipTheGap` or something like that

dmlap:
@BrandonOCasey I like that idea a lot. I'm digging through the HTML spec to see if there's a semantic equivalent for `playbackstalled` just to make sure we don't create a new name if one already exists. `skipTheGap()` or similar sounds like a good name-- I like that it's very clear about the side-effect of calling it.

test/gap-skipper.test.js
```diff
@@ -0,0 +1,157 @@
...
+QUnit.module('HLS', {
+  beforeEach() {
+    this.env = useFakeEnvironment();
+    this.requests = this.env.requests;
+    this.mse = useFakeMediaSource();
+    this.clock = this.env.clock;
+    this.old = {};
+
+    // mock out Flash features for phantomjs
+    this.old.Flash = videojs.mergeOptions({}, Flash);
```
dmlap: Use test-helpers, do not copy-paste this setup code from elsewhere.

BrandonOCasey:
@tenacex any more progress on this PR?

tenacex:
I'll do those changes now. @dmlap @BrandonOCasey

tenacex:
@BrandonOCasey just got time to address these, I'll let you know when I push.

src/gap-skipper.js
tenacex: Done. it is called `skipTheGap`

src/gap-skipper.js
tenacex: Changed.

src/videojs-contrib-hls.js
tenacex: Only reason I have kept it here is because it is the only place we have a reference to the player.

tenacex: Removed empty object.

test/gap-skipper.test.js
tenacex: Done.

src/gap-skipper.js
```diff
@@ -0,0 +1,84 @@
...
+
+const gapSkipper = function() {
+  this.ready(() => {
+    seekingObject = { player: this,
+                      tech: this.tech_,
+                      seeking: false,
+                      consecutiveUpdates: 0,
+                      timer: null,
+                      playerState: null,
+                      lastRecordedTime: null,
```
BrandonOCasey: Remove `lastRecordedTime` and `adativeSeeking` as they don't seem to be removed.

src/gap-skipper.js
```diff
@@ -0,0 +1,84 @@
+import Ranges from './ranges';
+
+let seekingObject;
+
+const gapSkipper = function() {
+  this.ready(() => {
+    seekingObject = { player: this,
```
BrandonOCasey: declare the object here rather than in the module scope with let aka

``` JavaScript
let seekingObject = { player: this
...
```

src/videojs-contrib-hls.js
```diff
@@ -12,10 +12,11 @@ import {Decrypter, AsyncStream, decrypt} from './decrypter';
 import utils from './bin-utils';
 import {MediaSource, URL} from 'videojs-contrib-media-sources';
 import m3u8 from './m3u8';
-import videojs from 'video.js';
 import MasterPlaylistController from './master-playlist-controller';
 import Config from './config';
-
+import gapSkipper from './GapSkipper';
```
BrandonOCasey: we don't need this first import

src/videojs-contrib-hls.js
```diff
@@ -289,6 +290,7 @@ class HlsHandler extends Component {
     if (tech.options_ && tech.options_.playerId) {
       let _player = videojs(tech.options_.playerId);
 
+      _player.gapSkipper();
```
BrandonOCasey: does videojs.gapSkipper also call this function? if so we should call gapSkipper somewhere else. Preferably towards the bottom of this constructor.

src/videojs-contrib-hls.js
```diff
@@ -677,6 +679,8 @@ videojs.HlsSourceHandler = HlsSourceHandler;
 videojs.Hls = Hls;
 videojs.m3u8 = m3u8;
 videojs.registerComponent('Hls', Hls);
+videojs.plugin('gapSkipper', gapSkipper);
```
gkatsev: why does this need to be a plugin?

src/gap-skipper.js
tenacex: What do you mean they don't seem to be removed? I still use `lastRecordedTime`, but I can remove `adaptiveSeeking`

src/gap-skipper.js
tenacex: Changed.

src/videojs-contrib-hls.js
BrandonOCasey: it is being designed so that we can pull it out of HLS and have it stand on it's own if needed

tenacex: Was told to build it like that.

src/videojs-contrib-hls.js
tenacex: Removed

src/gap-skipper.js
```diff
@@ -0,0 +1,81 @@
+import Ranges from './ranges';
+
+const gapSkipper = function() {
+  this.ready(() => {
+    let seekingObject = { player: this,
+                      tech: this.tech_,
```
BrandonOCasey: these should line up with the player variable above

src/videojs-contrib-hls.js
BrandonOCasey: any status on this?

test/gap-skipper.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  this.player.autoplay(true);
+  this.player.buffered = function() {
+    return videojs.createTimeRanges([[0, 10], [20, 30]]);
+  };
+  this.player.src({
+    src: 'master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  this.player.tech_.triggerReady();
```
BrandonOCasey: Can we get a test for seeking into a gap?

tenacex: seeking into a gap doesn't work.

BrandonOCasey:
LGTM

src/gap-skipper.js
```diff
@@ -0,0 +1,81 @@
+import Ranges from './ranges';
+
+const gapSkipper = function() {
+  this.ready(() => {
+    let seekingObject = { player: this,
```
forbesjo: Why is this an object?

src/videojs-contrib-hls.js
forbesjo: It's been decided that this should not be exposed as a plugin since it is not going to be used by any other projects soon.

src/videojs-contrib-hls.js
tenacex: Sorry @BrandonOCasey making the changes now.

tenacex: I don't think videojs calls this function, isn't this how you instantiate a plugin? Either way this is all subject to change because I am turning it into a module, so stay tuned.

src/gap-skipper.js
```diff
@@ -0,0 +1,86 @@
+import Ranges from './ranges';
+import videojs from 'video.js';
+
+export default class gapSkipper extends videojs.EventTarget {
+  constructor(options) {
+    super();
+    if (options.tech.options_.playerId) {
+      let self = this;
```
BrandonOCasey: use arrow functions () => rather than defining self again

src/gap-skipper.js
```diff
@@ -0,0 +1,86 @@
+import Ranges from './ranges';
+import videojs from 'video.js';
+
+export default class gapSkipper extends videojs.EventTarget {
```
BrandonOCasey: we don't have to extend EventTarget as we are not emitting any of our own events

src/gap-skipper.js
```diff
@@ -0,0 +1,86 @@
...
+    super();
+    if (options.tech.options_.playerId) {
+      let self = this;
+
+      this.player = videojs(options.tech.options_.playerId);
+      this.tech_ = options.tech;
+      this.seeking = false;
+      this.consecutiveUpdates = 0;
+      this.timer = null;
+      this.playerState = null;
```
BrandonOCasey: this file could probably use doc comments similar to the rest of the project

src/gap-skipper.js
```diff
@@ -0,0 +1,86 @@
...
+          clearTimeout(self.timer);
+        }
+      });
+    }
+  }
+
+  skipTheGap() {
+    let self = this;
+
+    if (!this.seeking) {
```
BrandonOCasey: I would return from the function if we are seeking so we don't have to wrap so many this code in so many if statements. Example:
```javascript
if (this.seeking) {
  return
}

File: test/gap-skipper.test.js [hunk 33]
```diff
@@ -0,0 +1,81 @@
...
+    this.player = createPlayer();
+  },
+
+  afterEach() {
+    this.env.restore();
+    this.mse.restore();
+    this.player.dispose();
+  }
+});
+QUnit.test('Adaptive seeking skips over gap in firefox with waiting event', function() {
```
BrandonOCasey: remove `Adaptive seeking` from the test name since we are not using that name anymore

test/gap-skipper.test.js
```diff
@@ -0,0 +1,81 @@
...
+import QUnit from 'qunit';
+import {
+  useFakeEnvironment,
+  useFakeMediaSource,
+  createPlayer,
+  openMediaSource,
+  standardXHRResponse
+} from './test-helpers.js';
+
+QUnit.module('Adaptive Seeking', {
```
BrandonOCasey: Call this module GapSkipper

test/gap-skipper.test.js
```diff
@@ -0,0 +1,81 @@
...
+  this.player.tech_.trigger('playing');
+  this.clock.tick(1);
+  this.player.currentTime(10);
+  this.player.trigger('waiting');
+  this.clock.tick(12000);
+  QUnit.equal(Math.round(this.player.currentTime()),
+    20, 'Player seeked over gap after timer');
+});
+
+QUnit.test('Adaptive seeking skips over gap in chrome without waiting event', function() {
```
BrandonOCasey: remove `Adaptive seeking` from the test name since we are not using that name anymore and it will be covered by the GapSkipper module name above

src/gap-skipper.js
```diff
@@ -0,0 +1,86 @@
...
+      let buffered = this.player.buffered();
+      let currentTime = this.player.currentTime();
+
+      if (buffered.length > 0) {
```
BrandonOCasey: you can probably do the same with this if statement and the next one as well

src/gap-skipper.js
tenacex: Changed.

src/gap-skipper.js
tenacex: Good call.

src/gap-skipper.js
tenacex: Isn't this the same if statement?

tenacex: I changed the one that check if we had a playerID though

test/gap-skipper.test.js
tenacex: Changed

test/gap-skipper.test.js
tenacex: K

src/gap-skipper.js
tenacex: Thanks. I needed that advice badly

test/gap-skipper.test.js
```diff
@@ -0,0 +1,81 @@
...
+    this.player = createPlayer();
+  },
+
+  afterEach() {
+    this.env.restore();
+    this.mse.restore();
+    this.player.dispose();
+  }
+});
+QUnit.test('skips over gap in firefox with waiting event', function() {
```
forbesjo: Could you add some new lines to these tests? It's difficult to read.

src/videojs-contrib-hls.js
```diff
@@ -677,6 +678,7 @@ videojs.HlsSourceHandler = HlsSourceHandler;
 videojs.Hls = Hls;
 videojs.m3u8 = m3u8;
 videojs.registerComponent('Hls', Hls);
+
```
forbesjo: Unnecessary new line.

src/videojs-contrib-hls.js
```diff
@@ -12,10 +12,10 @@ import {Decrypter, AsyncStream, decrypt} from './decrypter';
 import utils from './bin-utils';
 import {MediaSource, URL} from 'videojs-contrib-media-sources';
 import m3u8 from './m3u8';
-import videojs from 'video.js';
 import MasterPlaylistController from './master-playlist-controller';
 import Config from './config';
-
+import GapSkipper from './gap-skipper';
+import videojs from 'video.js';
```
forbesjo: No need to move this.

src/gap-skipper.js
```diff
@@ -0,0 +1,85 @@
...
+
+  skipTheGap() {
+
+    if (this.seeking) {
+      return;
+    }
+    let buffered = this.player.buffered();
+    let currentTime = this.player.currentTime();
+
+    if (buffered.length > 0) {
```
BrandonOCasey: for the next two if's you can check the opposite of what they are checking and return right away

test/gap-skipper.test.js
tenacex: Sure.

src/videojs-contrib-hls.js
tenacex: Removed.

src/videojs-contrib-hls.js
tenacex: Fixed it up.

src/gap-skipper.js
tenacex: Changed.

src/gap-skipper.js
```diff
@@ -0,0 +1,87 @@
...
+
+  skipTheGap() {
+
+    if (this.seeking) {
+      return;
+    }
+    let buffered = this.player.buffered();
+    let currentTime = this.player.currentTime();
+
+    if (buffered.length <= 0) {
```
BrandonOCasey: you may even just want to declare all the variables at the top and combine all of the if statements

BrandonOCasey:
just need doc comments and that that thing about combining the if statements and I think your good to go. [More info on doc comments](http://usejsdoc.org/about-getting-started.html) or you can just look at a file that has a class in it and copy the formatting.

src/gap-skipper.js
tenacex: The reason i chain them together is because I don't want to call the functions to create the variable values without first checking if the input arguments are valid. I think it might be best to keep it this way.

gkatsev: `Ranges.findNextRange` handles the case where the time ranges are of length zero, so, you could just remove this this if statement.

src/gap-skipper.js
```diff
@@ -0,0 +1,87 @@
...
+      return;
+    }
+    let nextRange = Ranges.findNextRange(buffered, currentTime);
+
+    if (nextRange.length <= 0) {
```
gkatsev: you can't have an array of length less than 0.

src/gap-skipper.js
tenacex: great @gkatsev

src/gap-skipper.js
```diff
@@ -0,0 +1,106 @@
...
+      }
+    });
+  }
+
+  /**
+  * Set a timer to skip the unbuffered region.
+  *
+  * @private
+  */
+  skipTheGap() {
```
BrandonOCasey: if this is private you should add an _ at the end of the name

src/gap-skipper.js
```diff
@@ -0,0 +1,106 @@
...
+ * region and there is a buffered region ahead.
+ * It then handles the skipping behavior.
+ *
+ * @class GapSkipper
+ */
+export default class GapSkipper {
+
+  /**
+  * Represents a GapSKipper object.
+  * @constructor
```
BrandonOCasey: need to add an `@param` for options

src/gap-skipper.js
tenacex: Changed.

src/gap-skipper.js
tenacex: Added

src/gap-skipper.js
```diff
@@ -17,6 +17,7 @@ export default class GapSkipper {
   /**
   * Represents a GapSKipper object.
   * @constructor
+  * @param {object} the options object that includes the tech
```
BrandonOCasey: Should be `@param {object} options object that includes the tech and settings`

Specifically the parameter name comes after the type

BrandonOCasey:
Once that doc comment is fixed LGTM

BrandonOCasey:
LGTM

BrandonOCasey:
see #735

PR #702, opened 2016-05-16
Use a single default audio track when not in html5 mode

forbesjo:
This PR is to make sure there is only one audio track when in Flash mode. Previously if a video had audio media groups in flash mode the track list would have a default track and all the audio groups (ex. "default", "track A", "track B").


BrandonOCasey:
LGTM

mrocajr:
This has been tested and is working as expected.

PR #704, opened 2016-05-16
Fix fastQualityChange_ to stop calling a function that no longer exists

imbcmdth:
The `currentTimeFunc` variable was somehow lost during all the rebases and merges. The function `fastQualityChange_` calls that property of `this` even though it no longer exists and throws an error. This PR simply restores the property.


dmlap:
It seems like the idea was to call the tech when currentTime() was required. Why not update `fastQualityChange_` to call the tech?

PR #706, opened 2016-05-18
Fix time correction

imbcmdth:
* Always apply timeCorrection_ when we haven't learned anything even if playlists have changed
* Apply timeCorrection_ directly to the currentTime when calling getMediaIndexForTime so that we can work around bad segment.end metadata in the playlist
* Emit an error if we have been stuck in a "timeCorrection" loop for more than 5 fetch attempts

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -407,22 +407,16 @@ export default class SegmentLoader extends videojs.EventTarget {
                                                           buffered);
 
     if (percentBuffered >= 90) {
-      // Retry the buffered calculation with the next segment if there is another
-      // segment after the currently selected segment
-      if (mediaIndex + 1 < playlist.segments.length) {
-        percentBuffered = this.getSegmentBufferedPercent_(playlist,
-                                                          mediaIndex + 1,
-                                                          currentTime,
-                                                          buffered);
+      // Increment the timeCorrection_ variable to push the fetcher forward
+      // in time and hopefully skip any gaps or flaws in our understanding
+      // of the media
+      this.incrementTimeCorrection_(playlist.targetDuration);
+
+      if (!this.paused()) {
+        this.fillBuffer_();
       }
 
-      // If both checks failed return and don't load anything
-      if (percentBuffered >= 90) {
-        return;
-      }
-
-      // Otherwise, continue with the next segment
-      mediaIndex += 1;
+      return;
     }
 
     segment = playlist.segments[mediaIndex];
@@ -850,18 +844,33 @@ export default class SegmentLoader extends videojs.EventTarget {
     // to the buffered time ranges and improves subsequent media
     // index calculations.
     if (!timelineUpdated) {
-      // appends haven't produced any new information for at least 5
-      // consecutive segments loads it is time to signal an error
-      // and stop
-      if (this.timeCorrection_ > this.playlist_.targetDuration * 5) {
-        this.timeCorrection_ = 0;
-        this.pause();
-        return this.trigger('error');
-      }
-
-      this.timeCorrection_ += segmentLength;
+      this.incrementTimeCorrection_(segmentLength);
     } else {
       this.timeCorrection_ = 0;
     }
+  }
+
+  /**
+   * add a number of seconds to the currentTime when determining which
+   * segment to fetch in order to force the fetcher to advance in cases
+   * where it may get stuck on the same segment due to buffer gaps or
+   * missing segment annotation after a rendition switch (especially
+   * during a live stream)
+   *
+   * @private
+   * @param {Number} secondsToIncrement number of seconds to add to the
+   * timeCorrection_ variable
+   */
+  incrementTimeCorrection_(secondsToIncrement) {
+    // If we have already incremented timeCorrection_ beyond the limit,
+    // then stop trying to find a segment, pause fetching, and emit an
+    // error event
+    if (this.timeCorrection_ >= this.playlist_.targetDuration * 5) {
+      this.timeCorrection_ = 0;
+      this.pause();
+      return this.trigger('error');
+    }
+
+    this.timeCorrection_ += secondsToIncrement;
   }
 }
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -364,8 +369,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (currentBuffered.length === 0) {
       // find the segment containing currentTime
       mediaIndex = getMediaIndexForTime(playlist,
-                                        currentTime,
-                                        this.expired_ + this.timeCorrection_);
+                                        currentTime + this.timeCorrection_,
+                                        this.expired_);
```
imbcmdth: The idea here is that expired will only ever be used when we do *not* have accurate timing data in the manifest. In most cases we do so the original code was not advancing the fetcher. Now we are directly fudging the `currentTime` so we force the fetcher to walk forward even if we have some (possibly incorrect) information recorded into the segment annotation.

src/segment-loader.js
```diff
@@ -303,6 +303,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (this.state === 'READY') {
       this.fillBuffer_();
     }
+
+    if (this.checkBufferTimeout_) {
+      window.clearTimeout(this.checkBufferTimeout_);
+    }
```
imbcmdth: This is a separate fix. I noticed that after seeking a lot we ended up with `monitorBuffer_ ` being called ever few milliseconds instead of twice a second. Something was causing the timer to be set repeatedly.

There is no good reason not to unset the timer before creating a new one just to guard against some other code calling `monitorBuffer_`.

src/segment-loader.js
```diff
@@ -757,12 +762,11 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-    // same playlist
-    if (segmentInfo && segmentInfo.playlist.uri === this.playlist_.uri) {
-      this.updateTimeline_(segmentInfo);
-      this.trigger('progress');
-    }
+
+    // add segment metadata if it we have gained information during the
+    // last append
+    this.updateTimeline_(segmentInfo);
+    this.trigger('progress');
```
imbcmdth: The condition here was moved into the `updateTimeline_` function.

src/segment-loader.js
```diff
@@ -819,32 +823,43 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    let segmentLength = this.playlist_.targetDuration;
     let playlist = segmentInfo.playlist;
     let currentMediaIndex = segmentInfo.mediaIndex;
 
     currentMediaIndex += playlist.mediaSequence - this.playlist_.mediaSequence;
     segment = playlist.segments[currentMediaIndex];
 
-    if (!segment) {
-      return;
-    }
```
imbcmdth: Don't exit early so that we have a chance to set `timeCorrection_`.

src/segment-loader.js
gesinger: Is it OK that we are now triggering progress even if we have switched playlists, whereas we weren't before?

For instance, if we go to appendBuffer, playlist is switched, and we just go ahead and append the buffer after appendBuffer finishes, we have progress for the old rendition, but not the current one. Should that count?

src/segment-loader.js
```diff
@@ -819,32 +823,43 @@ export default class SegmentLoader extends videojs.EventTarget {
...
     // index calculations.
     if (!timelineUpdated) {
-      this.timeCorrection_ -= segment.duration;
+      // appends haven't produced any new information for at least 5
+      // consecutive segments loads it is time to signal an error
+      // and stop
+      if (this.timeCorrection_ > this.playlist_.targetDuration * 5) {
+        this.timeCorrection_ = 0;
+        this.pause();
+        return this.trigger('error');
```
gesinger: Should we set more information about the error?

src/segment-loader.js
dmlap: Should we verify this in a test?

src/segment-loader.js
dmlap: `progress` in this file was written in the spirit of the HTML5 `progress` event, which is supposed to fire every ~350ms when media data is being loaded. I think it's fair to count segment loads against an old rendition as progress based on that.

src/segment-loader.js
dmlap: We should definitely add tests that verify the error is triggered and loading paused when this condition is reached.

src/segment-loader.js
dmlap: Seems like this should have been caught in a test. Sorry to be a broken record in this PR but... mind adding one?

src/segment-loader.js
imbcmdth: Done.

src/segment-loader.js
imbcmdth: Done.

src/segment-loader.js
```diff
@@ -402,22 +407,16 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-      if (mediaIndex + 1 < playlist.segments.length) {
-        percentBuffered = this.getSegmentBufferedPercent_(playlist,
-                                                          mediaIndex + 1,
-                                                          currentTime,
-                                                          buffered);
-      }
+      // Increment the timeCorrection_ variable to push the fetcher forward
+      // in time and hopefully skip any gaps or flaws in our understanding
+      // of the media
+      this.incrementTimeCorrection_(playlist.targetDuration);
```
imbcmdth: Now we use timeCorrection_ to walk forward if we have a >90% buffered segment.

dmlap: This function was intended to be non-side-effecting. Is there a way to reorganize the responsibilities here so that the time correction can be performed in one of the explicitly side-effecting portions of the code?

dmlap: Thought: would it be clearer if we made checkBuffer_() a utility function instead of a method?

dmlap:
LGTM

PR #710, opened 2016-05-24
Use browserfiy-shim globally.

gkatsev:
This is because media-sources requires video.js but doesn't require
browserify-shim itself, so, instead we need to make sure that we
browserify-shim all the dependencies.


imbcmdth:
LGTM

package.json
```diff
@@ -16,7 +16,7 @@
     "build:js": "npm-run-all build:js:babel build:js:browserify build:js:bannerize build:js:uglify",
     "build:js:babel": "babel src -d es5",
     "build:js:bannerize": "bannerize dist/videojs-contrib-hls.js --banner=scripts/banner.ejs",
-    "build:js:browserify": "browserify . -s videojs-contrib-hls -t browserify-shim -o dist/videojs-contrib-hls.js",
+    "build:js:browserify": "browserify . -s videojs-contrib-hls -g browserify-shim -o dist/videojs-contrib-hls.js",
```
dmlap: This is crazy.

PR #711, opened 2016-05-25
fix: SegmentLoader.handleUpdateEnd_ run after player was disposed

andrewmnlv:


dmlap:
Thanks for the PR! It seems to me though that we should have unregistered the handleUpdateEnd_ listener after dispose() so this shouldn't have been invoked in the first place. Would you be up for modifying this PR to properly clean up the listener on dispose instead?

andrewmnlv:
@dmlap np, on weekend

andrewmnlv:
@dmlap up

dmlap:
Perfect, thanks!

PR #712, opened 2016-05-25
Fix time correction.. some more

imbcmdth:
Host of small changes.

* Segments reporting a zero duration (due to rounding of their durations in the manifest) are now given a very short duration so that they will not be ignored during fetch calculations
* Added tests for and fixed getSegmentBufferedPercent_ calculations
  * Now returns the percent buffered of the entire segment duration instead of the "adjusted" duration
  * Handles segments reported as having a zero-duration
* Reduced the number of segments that we will attempt to "timeCorrect" when the segment chosen by `checkBuffer_` is already more than 90% buffered to 1
* No longer trigger errors from `timeCorrection_` handling, returning to the previous behavior
* Use the tech's setCurrentTime function (instead of masterPlaylistController's) in SegmentLoaders
  * Live-seeking correction in updateEndHandler_ should work again


src/m3u8/parse-stream.js
```diff
@@ -130,7 +130,7 @@ export default class ParseStream extends Stream {
         tagType: 'inf'
       };
       if (match[1]) {
-        event.duration = parseFloat(match[1]);
+        event.duration = Math.max(0.01, parseFloat(match[1]));
```
dmlap: When we've defaulted m3u8 attributes elsewhere, we've done it in the [parser and emitted an info event](https://github.com/videojs/videojs-contrib-hls/blob/23f4b359fd3df68d6d13648c0e2790aeb73b0453/src/m3u8/parser.js#L70)

imbcmdth: On it.

src/ranges.js
```diff
@@ -188,17 +188,16 @@ const bufferIntersection = function(bufferA, bufferB) {
  * @param {TimeRanges} buffered - the currently buffered time ranges
```
dmlap: This comment needs to be updated to account for the new parameter.

src/segment-loader.js
```diff
@@ -327,12 +331,42 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+
+    // The entire time range of the segment
+    let originalSegmentRange = videojs.createTimeRanges([[
+      startOfSegment,
+      endOfSegment
+    ]]);
+
+    // The adjusted segment time range that is setup such that it starts
+    // no earlier than currentTime
+    // Flash has no notion of a back-buffer so adjustedSegmentRange adjusts
```
dmlap: We shouldn't be apply Flash-specific fixes in this project. Can we move this into the Flash tech or media sources?

dmlap: I spoke too soon, ignore that comment.

src/segment-loader.js
```diff
@@ -397,6 +431,11 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
+    if (mediaIndex === playlist.segments.length - 1 &&
+        this.mediaSource_.readyState === 'ended') {
```
dmlap: I've been trying to keep this free of `this` references so the state it depends on is explicit in the parameters. Is it possible to move this out to the caller or add additional info to the calling arguments?

src/segment-loader.js
```diff
@@ -863,16 +909,15 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @param {Number} secondsToIncrement number of seconds to add to the
```
dmlap: The new param should be added to the doc comment.

test/segment-loader.test.js
```diff
@@ -451,18 +451,16 @@ QUnit.test('adjusts the playlist offset if no buffering progress is made after'
...
     QUnit.equal(this.requests[0].url, (i + '.ts'), 'requested the next segment');
     this.clock.tick(1);
     this.requests[0].response = new Uint8Array(10).buffer;
     this.requests.shift().respond(200, null, '');
     sourceBuffer.trigger('updateend');
   }
-
-  // so the loader should try the next segment
-  QUnit.equal(errors, 1, 'emitted error');
-  QUnit.ok(loader.paused(), 'loader is paused');
```
dmlap: It's a little weird that the loader is not paused but not fetching segments. I understand we don't want to trigger blacklisting in this particular case but maybe we should trigger a non-error event and pause the loader?

test/segment-loader.test.js
```diff
@@ -1090,3 +1088,149 @@ QUnit.test('doesn\'t allow more than one monitor buffer timer to be set', functi
...
+});
+
+QUnit.test('calculates the percent buffered for segments', function() {
+  let playlist = playlistWithDuration(30);
+
+  loader.mimeType(this.mimeType);
+  loader.playlist(playlist);
+
+  let buffered = videojs.createTimeRanges([[15, 19]]);
+  let percentBuffered = loader.getSegmentBufferedPercent_(playlist, 1, 0, buffered);
```
dmlap: Thought: should we move this utility into the Ranges module? It doesn't have any loader dependency and might keep us from accidentally tying the two together in the future. (Not a blocker for this PR or anything)

test/segment-loader.test.js
```diff
@@ -1090,3 +1088,149 @@ QUnit.test('doesn\'t allow more than one monitor buffer timer to be set', functi
 
   QUnit.equal(this.clock.methods.length, timeoutCount, 'timeout count remains the same');
 });
+
+QUnit.module('Segment Percent Buffered Calculations', {
```
dmlap: Great tests.

src/master-playlist-controller.js
```diff
@@ -72,7 +72,7 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       withCredentials: this.withCredentials,
       seekable: () => this.seekable(),
       seeking: () => this.tech_.seeking(),
-      setCurrentTime: (a) => this.setCurrentTime(a),
```
BrandonOCasey: we may just want to pass the tech to segment loaders at some point

src/segment-loader.js
```diff
@@ -13,6 +13,10 @@ const CHECK_BUFFER_DELAY = 500;
 // the desired length of video to maintain in the buffer, in seconds
 export const GOAL_BUFFER_LENGTH = 30;
 
+const clamp = function(num, [start, end]) {
```
BrandonOCasey: Needs doc comments

imbcmdth:
```
/**
 * It clamps
 */
```

Just kidding..

src/segment-loader.js
imbcmdth: There is no reason it couldn't be placed into `fillBuffer_`

test/segment-loader.test.js
imbcmdth: What would `unpause` the loader?

Also it isn't quite not fetching segments anymore - it's just given up on the walk forward. It'll still get the event in 500ms and try to make some progress (maybe, the buffer might be full enough.) The whole thing is ugly but I think it is the best we can do without introducing some more state (like a history of requests).

test/segment-loader.test.js
imbcmdth: That is a good idea. It evolved into something pretty general-purpose.

src/master-playlist-controller.js
dmlap: The tech is a big ball of shared state so I wish we could minimize the parts of the system that touch it.

src/segment-loader.js
dmlap: That sounds like a good idea.

test/segment-loader.test.js
dmlap: Interesting. Something to keep thinking on.

dmlap:
LGTM after a rebase.

BrandonOCasey:
LGTM

PR #714, opened 2016-05-26
Need to increment less than targetDuration

imbcmdth:
Jumping `targetDuration` into the future will often miss a segment since segment durations are defined in the spec as being **less than or equal to** target-duration.


dmlap:
LGTM though we might want to add some tests around this at some point.

PR #715, opened 2016-05-27
Percent buffered calculation now works correctly

imbcmdth:
## Description
Fixes a bug in the way we handle percentage-buffered calculations. Percent buffered calculation now works correctly when the segment is not buffered and currentTime is within the segment.

## Specific Changes proposed
Consider all the situations where currentTime is within the segment. Added tests for missing cases.

## Requirements Checklist


BrandonOCasey:
LGTM

PR #723, opened 2016-06-01
Manual rendition selection API

imbcmdth:
## Description
An implementation of a "representations" API that allows you to enable/disable media playlists from consideration by the adaptive bitrate selection algorithm.

## Specific Changes proposed
* Add a single function to HLS that returns a list of objects, one per media playlist, that contains the media's attributes (width, height, and bandwidth) and a function to set the enabled-state of the playlist.
* Leverage the `excludeUntil` property of playlists to control their enabled-state

## Requirements Checklist
- [x] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Play back HLS with video.js, even where it's not natively supported.
     - [hls.bytesReceived](#hlsbytesreceived)
     - [hls.selectPlaylist](#hlsselectplaylist)
     - [hls.xhr](#hlsxhr)
+    - [hls.representations](#hlsrepresentations)
   - [Events](#events)
     - [loadedmetadata](#loadedmetadata)
     - [loadedplaylist](#loadedplaylist)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -221,6 +222,34 @@ segment is downloaded. You can override this function to provide your
...
+#### hls.representations
+Type: `function`
+
+To get all of the available representations, call the `representations()` method on `player.hls`. This will return a list of plain objects, each with `width`, `height`, `bandwidth`, and `id` properties, and an `enabled()` method.
+
+```javascript
+player.hls.representations();
+```
+
+To see whether the representation is enabled or disabled, call its `enabled()` method with no arguments. To set whether it is enabled/disabled, call its `enabled()` method and pass in a boolean value.
```
gkatsev: can we call out explicitly that `enabled(true)` enables the representation and that `enabled(false)` disabled it?

src/rendition-mixin.js
```diff
@@ -0,0 +1,94 @@
...
+      this.bandwidth = attributes.BANDWIDTH;
+    }
+
+    // The id is simply the ordinality of the media playlist
+    // within the master playlist
+    this.id = id;
+
+    // Partially-apply the enableFunction to create a playlist-
+    // specific variant
+    this.enabled = enableFunction.bind(this, playlist, fastChangeFunction);
```
gkatsev: why not make it a prototype method?

src/rendition-mixin.js
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * The representation object encapsulates the publicly visible information
+ * in a media playlist along with a setter/getter-type function (enabled)
+ * for changing the enabled-state of a particular playlist entry
+ *
+ * @class Representation
+ */
+class Representation {
+  constructor(hlsHandler, playlist, id) {
```
gkatsev: should we default `id` to something?

src/rendition-mixin.js
imbcmdth: Each instance would still need a partially-applied version with the playlist and fastChangeFunction (the second one isn't explicitly required but I like the decoupling) which would make it pointless to have it on the prototype.

src/rendition-mixin.js
imbcmdth: Why?

gkatsev: Not sure, is it possible for it to be called without an id?

imbcmdth: No, it's called by us only (it is not exposed) and it gets an `id` from the `index` parameter of an array `map` which is always defined.

README.md
imbcmdth: Sure, makes sense.

src/rendition-mixin.js
gkatsev: you could store the playlist on the instance.

imbcmdth: I don't want to leak anything more than what is exposed. Putting the playlist on the representation encourages the user of the API to do HLS-specific things in their renditions selection functions.

As we will be using the API to build a general-purpose implementation on the `VideoTrack` object, I don't want to include any information that won't available for every ABR video source.

gkatsev: 👌 

Also, what about using `videojs.bind` instead if native bind?

PR #726, opened 2016-06-02
update to the newest version of the generator

BrandonOCasey:
## Description

Update HLS to the newest version of generator-videojs-plugin so that we can support budo (a faster dev server) and bower.
## Requirements Checklist


gesinger:
Closing, as much of the generator has changed since this PR, and with the changes in https://github.com/videojs/videojs-contrib-hls/pull/1151 , if we do want to update to the latest version of the generator, we'll be better off starting fresh after that PR.

PR #732, opened 2016-06-06
Added debug page

tenacex:
Added debug page that logs events and other important stats. It uses the CDN version of Video.js and HLS.


debug/audio-track-selector.js
```diff
@@ -0,0 +1,114 @@
+(function(videojs) {
```
imbcmdth: You shouldn't need this file at all. This functionality is included in the latest releases of video.js natively.

debug/index.html
```diff
@@ -0,0 +1,333 @@
...
+        });
+
+        videojs.Hls.displayStats(document.querySelector('.switching-stats'), player);
+        videojs.Hls.displayCues(document.querySelector('.segment-timeline'), player);
+
+      });
+    }
+
+    (function(window, videojs) {
+      videojs.options.flash.swf = '../../node_modules/videojs-swf/dist/video-js.swf';
```
imbcmdth: This file is loaded by video.js from the CDN there is no need to set this.

imbcmdth:
Also, this should be referenced someplace from the main (index) page and made to look like that page.

PR #735, opened 2016-06-08
Intelligent seeking over un-closeable buffer gaps - Part Deux 

imbcmdth:
Skip over gaps after giving the HLS library adequate time to try and close the gap. The idea is to emulate safari's behavior of "playing through" any missing segments. Instead of playing through them, we will wait the duration of the gap before seeking to the next region of the buffer containing video data.

Based on #697 but with some fixes and minor refactors to make it more robust and use fewer tracking state-variables.


src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+      });
+
+      // Set of conditions that reset the gap-skipper logic
+      [
+        'seeking',
+        'seeked',
+        'pause',
+        'playing',
+        'error'
+      ].forEach((event) => {
```
BrandonOCasey: I think you can actually pass an array of events to on, so we won't have to loop over them here.

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
+/**
+ * @file gap-skipper.js
+ */
```
BrandonOCasey: Should GapSkipper have a dispose function that gets called by HLS?

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    this.player = videojs(options.tech.options_.playerId);
+    this.tech_ = options.tech;
+    this.consecutiveUpdates = 0;
+    this.timer = null;
+    this.lastRecordedTime = null;
+
+    if (options.debug) {
+      this.logger_ = videojs.log.bind(videojs, '<gap-skipper>');
```
BrandonOCasey: Debugging code! NICE!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+      ].forEach((event) => {
+        this.player.on(event, () => {
+          this.cancelTimer_();
+        });
+      });
+    });
+  }
+
+  /**
+  * Cancels any pending timers and resets the 'timeupdate' mechanism
```
BrandonOCasey: doc comment spacing is off for all of the functions in this file

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+      this.logger_ = videojs.log.bind(videojs, '<gap-skipper>');
+    }
+
+    this.player.one('canplaythrough', () => {
+      this.player.on('waiting', () => {
+        this.setTimer_();
+      });
+
+      // The purpose of this function is to emulate the "waiting" event on
+      // browsers that do not emit it when they are stalled waiting for
```
BrandonOCasey: maybe add in specifics about firefox and the issue on their board here?

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+
+    this.logger_('timer triggered');
+
+    if (nextRange.length === 0) {
+      return;
+    }
+
+    this.logger_('currentTime:', currentTime, 'scheduled currentTime:', scheduledCurrentTime, 'nextRange start:', nextRange.start(0));
+
+    if (currentTime !== scheduledCurrentTime) {
```
BrandonOCasey: maybe merge this if statement and the one above it since they do the same thing?

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+    let nextRange = Ranges.findNextRange(buffered, currentTime);
+
+    if (nextRange.length === 0) {
+      return;
+    }
+
+    this.logger_('nextRange start:', nextRange.start(0));
+
+    if (this.timer !== null) {
+      return;
```
BrandonOCasey: merge these two if statements as well (since they both only do a return)

src/gap-skipper.js
```diff
@@ -0,0 +1,163 @@
...
+    this.consecutiveUpdates = 0;
+    this.timer = null;
+
+    this.logger_('timer triggered');
+
+    if (nextRange.length === 0) {
+      return;
+    }
+
+    this.logger_('currentTime:', currentTime, 'scheduled currentTime:', scheduledCurrentTime, 'nextRange start:', nextRange.start(0));
```
BrandonOCasey: something tells me the linter doesn't like the length of this line...

src/gap-skipper.js
imbcmdth: Added a dispose and refactored some stuff for that purpose.

src/gap-skipper.js
imbcmdth: I'm not sure which behavior is right. Or if Firefox is wrong. That is why I left as just an explanation of why we have two ways to trigger this gap-skipping behavior.

src/gap-skipper.js
```diff
@@ -0,0 +1,179 @@
...
+   * @constructor
+   * @param {object} options an object that includes the tech and settings
+   */
+  constructor(options) {
+    if (!options.tech.options_.playerId) {
+      return;
+    }
+
+    this.player_ = videojs(options.tech.options_.playerId);
+    this.tech_ = options.tech;
```
dmlap: This member variable is never used. In general, we should be try to minimize dependencies on the player and techs. They're both really complex, stateful objects and they're shared among many different actors. Definitely don't store it if you're not going to use it.

src/gap-skipper.js
```diff
@@ -0,0 +1,179 @@
...
+    this.consecutiveUpdates = 0;
+    this.lastRecordedTime = null;
+    this.timer_ = null;
+
+    if (options.debug) {
+      this.logger_ = videojs.log.bind(videojs, '<gap-skipper>');
+    }
+    this.logger_('<initialize>');
```
dmlap: Won't this output `<gap-skipper> <initialize>`? Are you planning on feeding your log statements to an XML parser? :)

src/gap-skipper.js
dmlap: It seems like you could get away with using only the tech and skipping the player reference here. Given this is going to be used by the Source Handler, it seems like only referencing the tech would be preferable.

src/gap-skipper.js
imbcmdth: Nope. Without closing tags I'd need something less strict to be able to parse this ..an HTML parser!

src/gap-skipper.js
imbcmdth: Very true.

src/gap-skipper.js
```diff
@@ -0,0 +1,179 @@
...
+
+    // The purpose of this function is to emulate the "waiting" event on
+    // browsers that does not emit it when they are stalled waiting for
+    // more data
```
dmlap: I don't think `waiting` is actually the event we're looking for here. My reading of the spec says [waiting](https://html.spec.whatwg.org/multipage/embedded-content.html#event-media-waiting) is reserved for scenarios where the user or user agent has triggered a condition that requires buffering before playback but there's no evidence that buffering won't be 100% successful. For instance, seeking out of buffer would trigger waiting because it's an intentional decision to move into a state that requires buffering. 

[stalled](https://html.spec.whatwg.org/multipage/embedded-content.html#event-media-stalled) indicates that buffering is being attempted but no progress is being made. This sounds closer to the scenario we're dealing with ("I'm trying to buffer but failing!") but still not right.

[A related issue](https://github.com/w3c/media-source/issues/88) was recently filed against the media sources spec. If the application developer (us) is doing the content fetching, are `progress`, `waiting`, and `stalled` events well-defined? I lean towards "no" which makes me think we shouldn't overload an existing event for this functionality. All this is an extraordinarily long-winded way of saying, I think we should make our own event, possibly trigger it in response to `waiting` and `stalled`, and use that new event for gap skips.

imbcmdth: [stalled](https://html.spec.whatwg.org/multipage/embedded-content.html#event-media-stalled) is definitely not the right event either. That only applies when the user-agent is also responsible for the fetching of media data. With MSE that responsibility is no longer in it's hands.

There are two main reasons that we can't use `stalled`:
1. What the event means in terms of "how long the data has been not forthcoming" is pretty much left to the browser (though the spec does helpfully say "about 3 seconds").
2. The event can be fired while we are playing. In fact, it often is fired while there is plenty of data in the buffer and the condition usually resolves itself (data arrives) without the browser taking any extra steps.

Waiting is the event IMO that should fire when player has reached the end of the buffer and is waiting for MSE to provide more data.

I agree that we probably shouldn't be firing waiting ourselves. I will stop triggering waiting myself and call the waitingHandler function directly from the work-around for non-waiting-triggering browsers.

src/gap-skipper.js
```diff
@@ -0,0 +1,176 @@
...
+      } else if (currentTime === this.lastRecordedTime) {
+        this.consecutiveUpdates++;
+      } else {
+        this.consecutiveUpdates = 0;
+        this.lastRecordedTime = currentTime;
+      }
+    };
+
+    // Set of events that reset the gap-skipper logic and clear the timeout
+    let timerCancelEvents = [
```
dmlap: This could be a `const` and moved out of the constructor.

src/gap-skipper.js
```diff
@@ -0,0 +1,176 @@
...
+      'seeking',
+      'seeked',
+      'pause',
+      'playing',
+      'error'
+    ];
+
+    let cancelTimerHandler = this.cancelTimer_.bind(this);
```
dmlap: Why are all the other handlers declared inline in the constructor except for this one?

src/gap-skipper.js
```diff
@@ -0,0 +1,176 @@
...
+   * is enabled globally
+   *
+   * @private
+   */
+  logger_() {}
+
+  /**
+   * A noop to ensure there is always have a dispose function even if there
+   * was no playerId in the global options and therefore the gapSkipper was
+   * never properly initialized
```
dmlap: This sounds like `playerId` is a required parameter. You should throw an error instead of returning a broken object in this case.

test/gap-skipper.test.js
```diff
@@ -0,0 +1,104 @@
...
+  for (let i = 0; i < 10; i++) {
+    this.player.tech_.trigger('timeupdate');
+  }
+  this.clock.tick(2000);
+
+  // checks that player doesn't seek before timer expires
+  QUnit.equal(this.player.currentTime(), 10, 'Player doesnt seek over gap pre-timer');
+  this.clock.tick(10000);
+
+  // check that pkayer jumped the gap
```
dmlap: `s/pkayer/player/`

dmlap:
Some minor comments but otherwise LGTM

PR #741, opened 2016-06-10
Timeout requests

tenacex:
This feature will set the timeout property on a segment request to zero if we are currently requesting from the only non-blacklisted playlist. This will help us avoid entering an infinitely stalled state.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -155,6 +155,19 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     loader.trigger('error');
   };
 
+  // returns the number of enabled playlists on the master playlist object
+  enabledPlaylists = function() {
+    let playlists = loader.master.playlists;
+    let count = 0;
+    let i;
+    for (i = 0; i < playlists.length; i++) {
+      if (playlists[i].excludeUntil <= Date.now()) {
+        count += 1;
+      }
+    }
+    return count;
+  };
+
   // update the playlist loader's state in response to a new or
   // updated playlist.
   haveMetadata = function(xhr, url) {
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -235,6 +235,9 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
   };
 
   loader.onLowestRendition = function() {
+    if (!loader.media()) {
+      return false;
+    }
     return loader.master.playlists.filter((element, index, array) => {
       return element.attributes.BANDWIDTH <= loader.media().attributes.BANDWIDTH ? true : false;
     }).length > 1 ? false : true;
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -229,13 +229,13 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
 
     // returns the number of enabled playlists on the master playlist object
   loader.enabledPlaylists = function() {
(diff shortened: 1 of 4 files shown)


src/segment-loader.js
```diff
@@ -504,6 +504,11 @@ export default class SegmentLoader extends videojs.EventTarget {
    *
    * @private
    */
+
+  disableTimeout() {
+    this.dontTimeout = true;
```
imbcmdth: We have been trying to avoid adding more and more state to the segment loader (even though we haven't been doing a great job at that!) Is there another way to accomplish this goal?

src/playlist-loader.js
```diff
@@ -227,6 +227,20 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     }
   };
 
+    // returns the number of enabled playlists on the master playlist object
+  loader.enabledPlaylists = function() {
+    let playlists = loader.master.playlists;
+    let count = 0;
+    let i;
+
+    for (i = 0; i < playlists.length; i++) {
```
imbcmdth: This could be more concisely written with a `filter` and returning the resultant array's length.

You should avoid using plain-ol'-`for` loops whenever possible.

imbcmdth:
In addition to the tests not passing, this is also missing the other acceptance criteria requirement - that we don't time-out when we are already on the lowest rendition.

tenacex:
@imbcmdth I thought we agreed that if we had only one more enabled playlist we were on the lowest rendition?

src/playlist-loader.js
tenacex: Changed.

src/segment-loader.js
tenacex: Trying to figure out an alternative.

src/segment-loader.js
```diff
@@ -542,7 +542,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     // decrease in network performance or a server issue.
 
     // don't timeout if we are on the last un-blacklisted playlist
-    segmentInfo.dontTimeout ? requestTimeout = 0 : requestTimeout = (segment.duration * 1.5) * 1000;
+    if (segmentInfo.dontTimeout) {
+      requestTimeout = 0;
+    } else {
+      requestTimeout = (segment.duration * 1.5) * 1000;
+    }
```
imbcmdth: This is not at all the solution was discussed at kickoff.

See the `this.hls_.xhr` function being invoked a few lines below this one? The first parameter is an object. The idea was that the base of that object would come from the MasterPlaylistController and have things like `timeout` and `withCredentials` set *there*. Then that object get passed along to the SegmentLoaders whenever a playlist update happened so that it was always current. In SegmentLoader, the default xhr options from MasterPlaylistController would get merged with an object with the `uri` and `responseType` properties whenever we need to call `xhr` to create the final xhr-related options that we will pass along to the `xhr` function.

tenacex: Sorry, I didn't fully understand you I guess. One problem I see with that is the request timeout isn't really calculated until we know the segment duration, which is only in the segment loader. Since we won't know the timeout request time within the MPC, I think there are only 2 ways to approach passing it in. We could either pass in an enable/disable timeout property, kind of like I used above. We then could use this switch to decide what timeout to use. as an alternative we could use the goal segment duration as the calculator for the timeout request, since I believe the playlists belonging to the MPC know this. Any preference?

PR #754, opened 2016-06-23
Fix desyncing audio and video when using alternate audio (not muxed t…

gesinger:
NOT READY FOR MERGE. This requires changes within mux.js in order to function appropriately. Namely, gaps may be created with this change, and mux.js must fill those gaps to allow the video to continue playing.
## Description

Bug fix for desyncing audio and video when using alternate audio.
When audio and video are not muxed together, two segment loaders act independently to retrieve segments. When a discontinuity is reached, each one independently determines a timestamp offset, and that is used as the baseMediaDecodeTime in mux.js. However, since the baseMediaDecodeTimes do not necessarily match, the difference will create desyncing.
## Specific Changes proposed

First, this enables expired time tracking for the audio segment loader. Then, when encountering discontinuities, segment loaders pause, trigger a discontinuity event, and wait to be told to continue with a timestamp offset. Master playlist controller listens for the discontinuity event, determines if there are two active segment loaders, and if there are, waits for the second, chooses the greatest timestamp offset, then calls into the segment loaders with that timestamp offset to continue.
## Requirements Checklist
  - [ ] Docs/guides updated

…ogether)

Track expired time in audio segment loader. Pause segment loader on discontinuity and wait for timestamp offset from master playlist controller. When using audio and video muxed together, master playlist controller will immediately use the only timestamp offset available. When using separate audio and video, the greatest of the two timestamp offsets will be selected and given to both segment loaders, syncing the two together.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -154,9 +154,13 @@ export default class MasterPlaylistController extends videojs.EventTarget {
       this.trigger('progress');
     });
 
+    this.mainSegmentLoader_.on('discontinuity', this.handleDiscontinuity_.bind(this));
+
     this.mainSegmentLoader_.on('error', () => {
       this.blacklistCurrentPlaylist(this.mainSegmentLoader_.error());
     });
+
+    this.audioSegmentLoader_.on('discontinuity', this.handleDiscontinuity_.bind(this));
 
     this.audioSegmentLoader_.on('error', () => {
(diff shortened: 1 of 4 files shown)


gesinger:
Although we still have to tackle this problem, we haven't seen it cause too much trouble in the real world, and we'll most likely approach it after we try to bring videojs-contrib-media-sources back into this project.

PR #759, opened 2016-06-29
remove no longer used audio button from utils/stats

BrandonOCasey:
## Description
Before video.js had support for an audio button to select multiple tracks we added a custom javascript plugin to show such a button. We need to remove it now that video.js has it's own audio button.

## Specific Changes proposed
removed audio button js and removed the reference to the file in the stats `index.html` page

## Requirements Checklist


imbcmdth:
LGTM

PR #763, opened 2016-06-29
Fix rendition selection

tenacex:
An additional call to load can correctly reset the segment-loader state to "READY" without any negative consequences. This allows us to use manual rendition selection much quicker without entering a frozen state. 

WAITING ON TESTS UNTIL APPROVAL OF THIS METHOD


forbesjo:
Looks like the freezing is caused by video underflow, #774 fixes this. Closing

PR #769, opened 2016-07-06
Made properties of hls.stats enumerable.

mrocajr:
## Description
Made all properties of hls.stats enumerable in order to be able to dynamically pull the whole object into analytics rather than targeting individual properties

## Specific Changes proposed
Made all properties of hls.stats enumerable

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -409,19 +409,19 @@ class HlsHandler extends Component {
     });
 
     Object.defineProperties(this.stats, {
-      'bandwidth': {
+      bandwidth: {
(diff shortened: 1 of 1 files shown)


src/videojs-contrib-hls.js
```diff
@@ -409,16 +409,20 @@ class HlsHandler extends Component {
     });
 
     Object.defineProperty(this.stats, 'bandwidth', {
```
gkatsev: while we're at it, why not change this to `defineProperties`?

mrocajr: Sounds good.

imbcmdth:
LGTM

PR #770, opened 2016-07-06
Timeout requests

mjneil:
## Description
We should always try to play even when faced with the worst network conditions possible. Right now there is a specific scenario that doesn't follow the "Always Be Playin'" mantra - when we get to the lowest rendition, we still allow segments to timeout which means people with really bad connections can wait forever and never get a video to play.

This is an extension of #741, I just didn't know how to add his PR to my work with git.

## Specific Changes proposed
The MasterPlaylistController controls when segment requests should timeout and passes that information to the segment loader. Segment requests will stop timing out if
1. If we are in the lowest rendition
2. If there is only one enabled stream left in the master playlist

## Requirements Checklist


src/playlist-loader.js
```diff
@@ -227,6 +228,44 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
   };
 
+  /**
+   * Returns the number of enabled playlists on the master playlist object
+   *
+   * @return {Number} number of eneabled playlists
+   */
+  loader.enabledPlaylists = function() {
+    return loader.master.playlists.filter((element, index, array) => {
+      return element.excludeUntil <= Date.now() || !element.excludeUntil ? true : false;
```
dmlap: You can rely on javascript's implicit boolean conversions to avoid the ternary:
```js
element.excludeUntil <= Date.now() || !element.excludeUntil
```
Since your second expression is a pre-requisite of the first, I'd swap the order:
```js
!element.excludeUntil || element.excludeUntil <= Date.now()
```
The `||` operator will short-circuit and skip the date comparison which is nice (but not important), too.

src/playlist-loader.js
```diff
@@ -227,6 +228,44 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
+   */
+  loader.onLowestEnabledRendition = function() {
+    if (!loader.media()) {
+      return false;
+    }
+
+    let currentPlaylist = loader.media().attributes.BANDWIDTH;
+
+    return loader.master.playlists.filter((element, index, array) => {
+      let enabled = typeof element.excludeUntil === 'undefined' ||
```
dmlap: This is very similar to `enabledPlaylists()`. Why not reuse that here?

src/playlist-loader.js
```diff
@@ -227,6 +228,44 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     }
   };
 
+  /**
+   * Returns the number of enabled playlists on the master playlist object
+   *
+   * @return {Number} number of eneabled playlists
+   */
+  loader.enabledPlaylists = function() {
```
dmlap: I'd err towards markings methods private unless you have a good use-case for adding them to our public API. We indicate private methods in video.js projects by appending an underscore to the method name. You could also convert this into a function by taking the master playlist as the first argument and then just skip exporting it.

test/master-playlist-controller.test.js
```diff
@@ -484,6 +484,27 @@ QUnit.test('updates the duration after switching playlists', function() {
...
+QUnit.test('removes request timeout when segment timesout on lowest rendition',
+function() {
+  this.masterPlaylistController.mediaSource.trigger('sourceopen');
+
+  // master
+  standardXHRResponse(this.requests.shift());
+  // media
+  standardXHRResponse(this.requests.shift());
+
+  this.masterPlaylistController.masterPlaylistLoader_.onLowestEnabledRendition = () => {
```
dmlap: You could set the initial bandwidth estimate to something really low to cause the loader to actually select the lowest rendition and skip this mocking.

src/playlist-loader.js
mjneil: Originally the check was to call `enabledPlaylists()` to determine if there was only one enabled playlist left and to call `onLowestRendition()` to check if the lowest rendition was active. This would fail if all but one rendition were disabled, but the enabled one was not the lowest. Since I'm not too familiar with how the player selects a playlist, this may not be a use case we have to check for? By adding the check for being enabled in this filter, it fixes that problem and removes the need to have to call `enabledPlaylists()` in the check at all. I left `enabledPlaylists()` in since I was under the impression the method was desired. If that is not the case we can remove it.

src/master-playlist-controller.js
```diff
@@ -143,14 +149,23 @@ export default class MasterPlaylistController extends videojs.EventTarget {
 
     this.masterPlaylistLoader_.on('mediachange', () => {
       let media = this.masterPlaylistLoader_.media();
+      let requestTimeout = (this.masterPlaylistLoader_.targetDuration * 1.5) * 1000;
 
       this.mainSegmentLoader_.abort();
 
+      // If we don't have any more available playlists, we don't want to
+      // timeout the request.
+      if (this.masterPlaylistLoader_.onLowestEnabledRendition_()) {
```
imbcmdth: This is a total nitpick but `on*` functions names are historically used to denote an event handler. Something like `isLowestEnabledRendition_` would be potentially less confusing.

test/segment-loader.test.js
```diff
@@ -647,76 +647,83 @@ QUnit.test('live playlists do not trigger ended', function() {
...
-  QUnit.ok(!this.requests[1].withCredentials, 'overrode segment request withCredentials');
-  videojs.options.hls = hlsOptions;
-});
+/**
+ * Segment loader no longer handles withCredentials, but instead
+ * receives this information from the MasterPlaylistController.
+ * TODO: Remove these tests from the file (kept for now until I know
+ *       there is not another reason segment loader needs this option).
+*/
+// QUnit.test('respects the global withCredentials option', function() {
```
imbcmdth: I wish we could find a way to run this kind of test. There were cases in the past where the `withCredentials` option was not being properly handled. Can you think of a way to test that the right `withCred.` option is passed to a SegmentLoader?

mjneil: This test looks like it is basically doing a similar test, except the option is being set at player creation level and not loader (which is how it needs to be done with this change now)
https://github.com/videojs/videojs-contrib-hls/blob/master/test/videojs-contrib-hls.test.js#L1204
Could add a second test to make sure that passing in the option would override the global default to cover these cases?

imbcmdth: Yup, sounds great! Then I would be ok with removing the commented out block completely.

imbcmdth:
LGTM

forbesjo:
Looks good to me

PR #773, opened 2016-07-08
Prevent trimming buffer between segment boundaries

marguinbc:
## Description
VideoJS is throwing PLAYER_ERR_TIMEOUT errors when a user replays a video or seeks to an area where the buffered content has been removed between segment boundaries. 

Steps to reproduce.
1) Load a video of a long-enough duration to trigger trimming of the buffer. For example, 1 min 50 seconds.
2) Let the video play through to the end without seeking so that the buffer is trimmed from the start of the video. In my example case, the video was trimmed to 10.2 seconds but the first segment ended at 9.8 seconds. This caused a gap of .4 seconds which was not enough to force the second segment to reload
3) Replay the video. 
4) The video stalled after the first segment played, resulting in a PLAYER_ERR_TIMEOUT

It is likely the same scenario will play out when seeking in longer videos if seekable.start() or currentTime-60 causes the removeToTime to fall between segment boundaries.

## Specific Changes proposed
This suggested fix attempts to determine if the removeToTime is between segment boundaries by adding the duration of segments up to removeToTime and comparing that duration to removeToTime. If removeToTime is between segments, this fix chooses the duration up to the previous segment as the removeToTime.

Two tests were also updated to reflect that we are trimming to the segment prior to seekable.start() or seeking one segment less than the default 60 seconds.

Seeking input on the viability of this change.

## Requirements Checklist
- [x ] Feature implemented / Bug fixed
- [x ] If necessary, more likely in a feature request than a bug fix
  - [x ] Unit Tests updated or fixed
  - [ ] Docs/guides updated


src/segment-loader.js
```diff
@@ -529,6 +529,28 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+
+      if (removeDuration >= removeToTime) {
+        removeToTime = previousDuration;
+        break;
+      }
+    }
+
     if (removeToTime > 0) {
       this.sourceUpdater_.remove(0, removeToTime);
     }
```
imbcmdth: This whole function is starting to get rather large. How do you feel about pulling the back-buffer removal code (all of it) into it's own function?

marguinbc: Sure thing!

marguinbc:
@imbcmdth, broke out the functionality that checks for the portion to be trimmed as trimBuffer_() - feel free to re-review.

imbcmdth:
LGTM

marguinbc:
QA passed. Looking for one more core contributor to review and add a LGTM

forbesjo:
LGTM

PR #774, opened 2016-07-08
Skip gaps caused by video underflow

gesinger:
## Description
In the event that there is a small gap in the video buffer, Chrome will consider it video underflow, and will continue playing audio for ~3 seconds while video remains frozen. The current time of the player will update to reflect the audio playback. Due to the audio overplay, the gap skipper will think we are in a buffered region, and will not skip over the gap.

## Specific Changes proposed
This adds support for skipping over gaps caused by a browser's interpretation of video underflow.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/gap-skipper.js
+++ b/src/gap-skipper.js
@@ -163,29 +163,18 @@ export default class GapSkipper {
     // Chrome's video froze at 10 seconds, where the video buffer encountered the gap,
     // however, the audio continued playing until it reached ~3 seconds past the gap
     // (13 seconds), at which point it stops as well. Since current time is past the
-    // gap, findNextRange will return no ranges. To check for this issue, we see if
-    // we can get a next range from 4 seconds back (4 seconds to account for variance).
-    let nextRangeFromPossibleGap = Ranges.findNextRange(buffered, currentTime - 4);
-
-    if (nextRangeFromPossibleGap.length === 0) {
-      // There was no gap, and no available next buffer to jump to.
-      return null;
-    }
-
-    // Now that we've verified there is a gap, we need to check to see if our current
-    // time reflects audio overplay from the gap. We do this by seeing if the gap is
-    // small enough, and that it is somewhere within the 3 second range (3 seconds +/-
-    // 1 second).
-
+    // gap, findNextRange will return no ranges.
+    //
+    // To check for this issue, we see if there is a small gap that is somewhere within
+    // a 3 second range (3 seconds +/- 1 second) back from our current time.
     let gaps = Ranges.findGaps(buffered);
 
     for (let i = 0; i < gaps.length; i++) {
       let start = gaps.start(i);
       let end = gaps.end(i);
 
-      if (
-        // gap is small
-        end - start < 1 &&
+      // gap is small
+      if (end - start < 1 &&
         // gap is 3 seconds back +/- 1 second
         currentTime - start < 4 && currentTime - end > 2) {
         return {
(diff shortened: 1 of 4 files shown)


test/ranges.test.js
```diff
@@ -211,3 +230,53 @@ QUnit.test('calculates the percent buffered for segments ' +
 
   QUnit.equal(percentBuffered, 95, 'calculated the buffered amount correctly');
 });
+
+QUnit.test('finds next range', function() {
```
imbcmdth: Nice!

forbesjo:
Looks good to me

src/gap-skipper.js
```diff
@@ -144,6 +144,60 @@ export default class GapSkipper {
...
+    // video buffer:   0 => 10.1, 10.2 => 20
+    // audio buffer:   0 => 20
+    // overall buffer: 0 => 10.1, 10.2 => 20
+    // current time: 13
+    //
+    // Chrome's video froze at 10 seconds, where the video buffer encountered the gap,
+    // however, the audio continued playing until it reached ~3 seconds past the gap
+    // (13 seconds), at which point it stops as well. Since current time is past the
+    // gap, findNextRange will return no ranges. To check for this issue, we see if
+    // we can get a next range from 4 seconds back (4 seconds to account for variance).
```
dmlap: Good comment.

src/gap-skipper.js
```diff
@@ -144,6 +144,60 @@ export default class GapSkipper {
...
+    let nextRangeFromPossibleGap = Ranges.findNextRange(buffered, currentTime - 4);
+
+    if (nextRangeFromPossibleGap.length === 0) {
```
dmlap: This seems duplicative of the tests below. Is the intent to short-circuit the rest of this function when there is no possibility of a video gap? Is this code performance sensitive enough to warrant that?

src/gap-skipper.js
```diff
@@ -154,8 +208,23 @@ export default class GapSkipper {
...
+      // stuck in a gap due to video underflow.
+      let gap = this.gapFromVideoUnderflow_(buffered, currentTime);
+
+      if (gap) {
+        this.logger_('setTimer_:',
+                     'Encountered a gap in video',
+                     'from: ', gap.start,
+                     'to: ', gap.end,
+                     'seeking to current time: ', currentTime);
+        this.tech_.setCurrentTime(currentTime);
```
dmlap: This seems worth a comment about why seeking to `currentTime` resolves this issue.

src/ranges.js
```diff
@@ -64,6 +63,28 @@ const findNextRange = function(timeRanges, time) {
 };
 
 /**
+ * Returns gaps within a list of TimeRanges
+ * @param {TimeRanges} buffered - the TimeRanges object
+ * @return {TimeRanges} a TimeRanges object with gaps
```
dmlap: "with gaps" is not very precise.

src/gap-skipper.js
```diff
@@ -144,6 +144,60 @@ export default class GapSkipper {
...
+    // small enough, and that it is somewhere within the 3 second range (3 seconds +/-
+    // 1 second).
+
+    let gaps = Ranges.findGaps(buffered);
+
+    for (let i = 0; i < gaps.length; i++) {
+      let start = gaps.start(i);
+      let end = gaps.end(i);
+
+      if (
```
dmlap: We don't normally newline on `(`

dmlap: @imbcmdth how do you feel about this style? If I'm the only one that has trouble parsing it, I'll retrain my brain.

src/gap-skipper.js
gesinger: You're right. I tried to separate steps a little bit for clarity (and possibly for a bit of efficiency), but it may not even be much more efficient, and it ends up being clearer without this step. Removing.

src/gap-skipper.js
gesinger: I thought it might've been more readable, but I think it is just as readable moving it all back up.

gesinger:
Updated.

imbcmdth:
LGTM

PR #785, opened 2016-07-20
Add ability to manually pause playlist and segment loaders

ntadej:
## Description

Our HLS streams expire if not played. Thus when pausing video, player should not request on server anymore as stream will expire soon. To keep last frame on video when paused and prevent loading new playlists/segments all playlist and segment loaders should be paused.

This improves and replaces PR #349.
## Specific Changes proposed

`MasterPlaylistController` class already contained `pauseLoading` function, which was only executed on error. It was extended to also pause master playlist loader and all audioTrack loaders.

A public `pauseLoading` function was added to contrib to add user ability to pause loaders.

Not sure if unit tests are needed here. I can update docs/guides/examples when approved.
## Requirements Checklist
- [x] Feature implemented
  - [ ] Docs/guides updated

### DIFF
--- a/src/hls-audio-track.js
+++ b/src/hls-audio-track.js
@@ -85,6 +85,19 @@ export default class HlsAudioTrack extends AudioTrack {
   }
 
   /**
+   * pause all media loaders
+   */
+  pause() {
+    for (let i = 0; i < this.mediaGroups_.length; i++) {
+      let mgl = this.mediaGroups_[i];
+
+      if (mgl.loader) {
+        mgl.loader.pause();
+      }
+    }
+  }
+
+  /**
    * Dispose of this audio track and
    * the playlist loader that it holds inside
    */
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -330,9 +330,7 @@ class HlsHandler extends Component {
       this.setCurrentTime(this.tech_.currentTime());
     });
     this.on(this.tech_, 'error', function() {
-      if (this.masterPlaylistController_) {
-        this.masterPlaylistController_.pauseLoading();
-      }
+      this.pauseLoading();
     });
 
     this.audioTrackChange_ = () => {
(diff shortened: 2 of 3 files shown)


gkatsev:
Looks like the tests are failing due to the linter which should end up getting fixed by https://github.com/videojs/videojs-contrib-hls/pull/788

ntadej:
According to Apple documentation, this should not be present in live playlists:
The EXT-X-ENDLIST tag is not present in the Live playlist, indicating that new media files will be added to the index file as they become available.

We want to replicate the following iOS feature:
https://developer.apple.com/reference/avfoundation/avplayeritem/1388752-canusenetworkresourcesforlivestr?language=objc

When live stream is paused, we do not want it to continue requesting any playlists or segments from server,

src/videojs-contrib-hls.js
```diff
@@ -330,9 +330,7 @@ class HlsHandler extends Component {
       this.setCurrentTime(this.tech_.currentTime());
     });
     this.on(this.tech_, 'error', function() {
-      if (this.masterPlaylistController_) {
```
gesinger: Was there a specific reason this condition was removed?

src/hls-audio-track.js
```diff
@@ -84,6 +84,19 @@ export default class HlsAudioTrack extends AudioTrack {
     }
   }
 
+  /**
+   * pause all media loaders
+   */
+  pause() {
```
gesinger: This file no longer exists.

gesinger (commented):
Sorry this PR has gone on for a while without much response.

The main reason we don't pause loading of manifest data is to try to keep things up-to-date. For instance, if a live stream is paused, and we missed discontinuities or timing info, we may not be able to sync well with the current stream on resume.

As to the comment on live playlists and EXT-X-ENDLIST, that tag is not only used for VOD, but it is also used at the end of an HLS live stream to denote that the live stream has finished.

That said, if you still need this as an option, we can still get pauseLoading functionality in, though it would be good to have a test, to document it, and to note that this may lead to some issues if used in certain streaming scenarios.

And thank you for the contribution.

forbesjo:

PR #788, opened 2016-07-20
move global devDependency to normal dependency

BrandonOCasey:
## Description
global is used in the code as a normal dependency, we need to put it in the normal dependency list.

## Specific Changes proposed
change package.json to match description

## Requirements Checklist

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -9,6 +9,7 @@ import resolveUrl from './resolve-url';
 import {mergeOptions} from 'video.js';
 import Stream from './stream';
 import m3u8 from 'm3u8-parser';
+import window from 'global/window';
 
 /**
   * Returns a new array of segments that is the result of merging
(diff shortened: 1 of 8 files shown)


src/playlist-loader.js
```diff
@@ -9,6 +9,7 @@ import resolveUrl from './resolve-url';
 import {mergeOptions} from 'video.js';
 import Stream from './stream';
 import m3u8 from 'm3u8-parser';
+import {window} from 'global';
```
gkatsev: window is the default export from `global` or `global/window`, does this work? don't we want to do `import window from 'global/window'`? (or `import window from 'global'`)

BrandonOCasey: The build shows that it works. I think that `import window from 'global/window';` makes sense too and if you have a strong preference to it I can change it. I think we use it similarly in other projects. I don't really like `import window from 'global'` as it may be confusing since global can be used to require more than just window.

forbesjo:
Looks good to me

src/playlist-loader.js
gkatsev: No strong preference, `import window from 'global/window'` is nice because that's what we use elsewhere.

BrandonOCasey: ok I agree and will change it to that

PR #791, opened 2016-07-21
Cue Tags

gesinger:
## Description
Add support for passing data via video cues when encountering #EXT-X-CUE-OUT, #EXT-X-CUE-OUT-CONT, and #EXT-X-CUE-IN in m3u8. Requires `useCueTags` to be set to true in HLS options.

## Requirements Checklist
  - [X] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Play back HLS with video.js, even where it's not natively supported.
       - [Source](#source)
     - [List](#list)
       - [withCredentials](#withcredentials)
-      - [useTagCues](#usetagcues)
+      - [useCueTags](#usecuetags)
   - [Runtime Properties](#runtime-properties)
     - [hls.playlists.master](#hlsplaylistsmaster)
     - [hls.playlists.media](#hlsplaylistsmedia)
@@ -213,14 +213,16 @@ is set to `true`.
 See html5rocks's [article](http://www.html5rocks.com/en/tutorials/cors/)
 for more info.
 
-##### useTagCues
+##### useCueTags
 * Type: `boolean`
 * can be used as an initialization option
 
-When the `useTagCues` property is set to `true,` a text track is created with
+When the `useCueTags` property is set to `true,` a text track is created with
 the id 'hls-segment-metadata' and kind 'metadata'. The track is then added to
-`player.textTracks()`. Whenever a segment is playing, its associated tags will
-be listed as a tags property under its active cue. Changes in active cue may be
+`player.textTracks()`. Whenever a segment associated with a cue tag is playing,
+the cue tags will be listed as a properties inside of a stringified JSON object
+under its active cue's `text` property. The properties that are currently
+supported are cueOut, cueOutCont, and cueIn. Changes in active cue may be
 tracked by following the Video.js cue points API for text tracks. For example:
 
 ```javascript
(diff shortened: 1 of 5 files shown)


gesinger:
Note that this simply removes all old cues and adds new cues when loading a new playlist. We can change this to de-duping if it is preferable, but it seemed to make sense as when seeking back with a new playlist we'd want the new playlist's cues rather than the old. However, this could also cause some false cues to be triggered if we are seeking back into buffered range of an old playlist. We can discuss specifics around that case.

README.md
```diff
@@ -212,6 +213,49 @@ is set to `true`.
...
+  cuesTrack,
+  i;
+
+for (i = 0; i < textTracks.length; i++) {
+  if (textTracks[i].label === 'hls-segment-metadata') {
+    cuesTrack = textTracks[i];
+  }
+}
+
+cuesTrack.oncuechange = function() {
```
gkatsev: please use `cuesTrack.addEventListener('cuechange'` over `oncuechange` if possible.

README.md
```diff
@@ -212,6 +213,49 @@ is set to `true`.
...
+  var
```
gkatsev: I've been starting to use es6 in code samples as well when it's reasonable. So, these would just end up being `let`s. But no strong preference either way.

src/master-playlist-controller.js
```diff
@@ -804,4 +814,42 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+    }
+
+    let startTime;
+    let endTime = 0;
+
+    for (let i = 0; i < media.segments.length; i++) {
+      let segment = media.segments[i];
+
+      startTime = endTime;
+      endTime = startTime + segment.duration;
```
imbcmdth: Change the duration of the cue to be a short, non-zero value to be an "instantaneous" type cue. Follow the example in contrib-media-sources for ID3-tag cues.

gkatsev: The "short" value should be like 0.5s.
Also, we might want to consider making these have a start time of 0 and an end time of the cue time. That's what that w3c spec draft recommends for cuepoints.

gesinger: ID3 uses a duration of "0" (start time and end time are equal).

The goal is that we want to trigger these cue points once we cross between segments (for us, we are considering it the beginning of the segment the tag precedes). We only need it to trigger the cue change and remain active long enough for a listener of the event to grab the cue.

@gkatsev is the recommendation for 0.5 seconds the minimum for it remaining active long enough for a user to retrieve and make use of?

gkatsev: Yes, but also so that when we're using the emulated text tracks, we don't lose the cue altogether.

gesinger:
Updated.

gesinger:
Rebased on master and fixed linting errors.

README.md
```diff
@@ -212,6 +213,44 @@ is set to `true`.
...
+
+for (let i = 0; i < textTracks.length; i++) {
+  if (textTracks[i].label === 'hls-segment-metadata') {
+    cuesTrack = textTracks[i];
+  }
+}
+
+cuesTrack.addEventListener('cuechange', function() {
+  let activeCues = cuesTrack.activeCues;
+  let activeCue;
```
gkatsev: extra `let activeCue;`

imbcmdth:
LGTM

PR #794, opened 2016-07-22
Guard against negative media index due to manifest vs actual timing differences

gesinger:
## Description

One of the possible reasons for us getting a negative media index is that we've recorded actual segment timing information and it differs enough from the manifest's timing information that we can't properly place the first segment. In the event that we're at time 0 and the video isn't live, we can safely assume that we should get the first segment.

## Requirements Checklist
  - [ ] Docs/guides updated


imbcmdth:
LGTM

PR #796, opened 2016-07-26
Set the linter to a specific version

forbesjo:
This is to avoid breaking the build if there is a subtle breaking rule change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "shelljs": "^0.5.3",
     "sinon": "1.10.2",
     "uglify-js": "^2.5.0",
-    "videojs-standard": "^4.0.0",
+    "videojs-standard": "4.0.2",
     "watchify": "^3.6.0"
   }
 }


gesinger:
LGTM

misteroneill:
You can do this, but I wouldn't suggest it. There was a recent change to the linter that accidentally introduced breaking changes (my fault!), but typically this shouldn't happen. I'd suggest reconsidering.

gkatsev:
Also, this actually won't prevent issues because of the transitivity of dependencies.
Because videojs-standard depends on eslint-config-videojs with a `^` so it'll  pick up any patch and minor changes.
What we should've done is revert the change, release a patch version with the reversion, and then make a major update so consumers can update at will.
Should we just do it now anyway? @misteroneill

forbesjo:
I'll open PRs against the HLS projects with `^5.0.0` after the video.js linting process has finished up

PR #802, opened 2016-08-01
fix cue tags start times to account for expired time

mjneil:
## Description
Need to account for the sliding window playlist in a live stream to properly set the start time of each cue in the cueTags track

## Requirements Checklist


gesinger:
Looks good, but I think a test would be good to add.

mjneil:
Closing in favor of #804

PR #804, opened 2016-08-02
Ad cues

mjneil:
## Description
Provide a better cue structure for ad related cues. This PR only handles creating cues for `#EXT-X-CUE-OUT`, `#EXT-X-CUE-OUT-CONT` and `#EXT-X-CUE-IN` tags. Other ad related cues will eventually be represented in this format as well.

This will also make #802 unnecessary.

Current TextTrack layout:
```
----------------------
  ^co   ^coc   ^ci 
```

Proposed TextTrack layout:
```
-----------------------
  ^^^^^^^^^^ad-cue
```

Proposed cue-format:
```
{
  startTime: 0,
  endTime: 120,
  adStartTime: -10,
  adEndTime: 240
}
```

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/ad-cue-tags.js
+++ b/src/ad-cue-tags.js
@@ -20,10 +20,10 @@ const findAdCue = function(track, mediaTime) {
 };
 
 const initAdCueTrack = function(controller) {
-      controller.cueTagsTrack_ = controller.tech_.addTextTrack('metadata',
-        'hls-segment-metadata');
-      controller.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
-      controller.tech_.textTracks().addTrack_(controller.cueTagsTrack_);
+  controller.cueTagsTrack_ = controller.tech_.addTextTrack('metadata',
+    'hls-segment-metadata');
+  controller.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
+  controller.tech_.textTracks().addTrack_(controller.cueTagsTrack_);
 };
 
 const updateAdCues = function(media, track, offset = 0) {
--- a/src/ad-cue-tags.js
+++ b/src/ad-cue-tags.js
@@ -20,7 +20,8 @@ const findAdCue = function(track, mediaTime) {
 };
 
 const initAdCueTrack = function(controller) {
-      controller.cueTagsTrack_ = controller.tech_.addTextTrack('metadata', 'hls-segment-metadata');
+      controller.cueTagsTrack_ = controller.tech_.addTextTrack('metadata',
+        'hls-segment-metadata');
       controller.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
       controller.tech_.textTracks().addTrack_(controller.cueTagsTrack_);
 };
--- a/src/ad-cue-tags.js
+++ b/src/ad-cue-tags.js
@@ -61,7 +61,7 @@ const updateAdCues = function(media, track, offset = 0) {
       if ('cueOut' in segment) {
         cue = new window.VTTCue(mediaTime,
                                 mediaTime + segment.duration,
-                                '');
+                                segment.cueOut);
         cue.adStartTime = mediaTime;
         // Assumes tag format to be
         // #EXT-X-CUE-OUT:30
--- a/src/ad-cue-tags.js
+++ b/src/ad-cue-tags.js
@@ -19,13 +19,6 @@ const findAdCue = function(track, mediaTime) {
   return null;
 };
 
-const initAdCueTrack = function(controller) {
-  controller.cueTagsTrack_ = controller.tech_.addTextTrack('metadata',
-    'hls-segment-metadata');
-  controller.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
-  controller.tech_.textTracks().addTrack_(controller.cueTagsTrack_);
-};
-
 const updateAdCues = function(media, track, offset = 0) {
   if (!media.segments) {
     return;
@@ -46,7 +39,6 @@ const updateAdCues = function(media, track, offset = 0) {
     }
 
     if (cue) {
-
       if ('cueIn' in segment) {
         // Found a CUE-IN so end the cue
         cue.endTime = mediaTime;
@@ -94,13 +86,11 @@ const updateAdCues = function(media, track, offset = 0) {
         track.addCue(cue);
       }
     }
-
     mediaTime += segment.duration;
   }
 };
 
 export default {
-  initAdCueTrack,
   updateAdCues,
   findAdCue
 };
--- a/src/ad-cue-tags.js
+++ b/src/ad-cue-tags.js
@@ -1,0 +1,106 @@
+/**
+ * @file ad-cue-tags.js
+ */
+import window from 'global/window';
+
+/**
+ * Searches for an ad cue that overlaps with the given mediaTime
+ */
+const findAdCue = function(track, mediaTime) {
+  let cues = track.cues;
+
+  for (let i = 0; i < cues.length; i++) {
+    let cue = cues[i];
+
+    if (mediaTime >= cue.adStartTime && mediaTime <= cue.adEndTime) {
+      return cue;
+    }
+  }
+  return null;
+};
+
+const initAdCueTrack = function(controller) {
+      controller.cueTagsTrack_ = controller.tech_.addTextTrack('metadata', 'hls-segment-metadata');
+      controller.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
+      controller.tech_.textTracks().addTrack_(controller.cueTagsTrack_);
+};
+
+const updateAdCues = function(media, track, offset = 0) {
+  if (!media.segments) {
+    return;
+  }
+
+  let mediaTime = offset;
+  let cue;
+
+  for (let i = 0; i < media.segments.length; i++) {
+    let segment = media.segments[i];
+
+    if (!cue) {
+      // Since the cues will span for at least the segment duration, adding a fudge
+      // factor of half segment duration will prevent duplicate cues from being
+      // created when timing info is not exact (e.g. cue start time initialized
+      // at 10.006677, but next call mediaTime is 10.003332 )
+      cue = findAdCue(track, mediaTime + (segment.duration / 2));
+    }
+
+    if (cue) {
+
+      if ('cueIn' in segment) {
+        // Found a CUE-IN so end the cue
+        cue.endTime = mediaTime;
+        cue.adEndTime = mediaTime;
+        mediaTime += segment.duration;
+        cue = null;
+        continue;
+      }
+
+      if (mediaTime < cue.endTime) {
+        // Already processed this mediaTime for this cue
+        mediaTime += segment.duration;
+        continue;
+      }
+
+      // otherwise extend cue until a CUE-IN is found
+      cue.endTime += segment.duration;
+
+    } else {
+      if ('cueOut' in segment) {
+        cue = new window.VTTCue(mediaTime,
+                                mediaTime + segment.duration,
+                                '');
+        cue.adStartTime = mediaTime;
+        // Assumes tag format to be
+        // #EXT-X-CUE-OUT:30
+        cue.adEndTime = mediaTime + parseFloat(segment.cueOut);
+        track.addCue(cue);
+      }
+
+      if ('cueOutCont' in segment) {
+        // Entered into the middle of an ad cue
+        let adOffset;
+        let adTotal;
+
+        // Assumes tag formate to be
+        // #EXT-X-CUE-OUT-CONT:10/30
+        [adOffset, adTotal] = segment.cueOutCont.split('/').map(parseFloat);
+
+        cue = new window.VTTCue(mediaTime,
+                                mediaTime + segment.duration,
+                                '');
+        cue.adStartTime = mediaTime - adOffset;
+        cue.adEndTime = cue.adStartTime + adTotal;
+        track.addCue(cue);
+      }
+    }
+
+    mediaTime += segment.duration;
+  }
+};
+
+export default {
+  initAdCueTrack,
+  updateAdCues,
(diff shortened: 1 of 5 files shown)


src/master-playlist-controller.js
```diff
@@ -34,6 +34,22 @@ const parseCodecs = function(codecs) {
...
+  let cues = track.cues;
+
+  for (let i = 0; i < cues.length; i++) {
+    let cue = cues[i];
+
+    if (mediaTime >= cue.adStartTime && mediaTime <= cue.adEndTime) {
+      return cue;
+    }
+  }
+  return undefined;
```
gesinger: Better to return null here

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+    let mediaTime = offset;
+    let track = this.cueTagsTrack_;
+    let cue;
 
     for (let i = 0; i < media.segments.length; i++) {
       let segment = media.segments[i];
 
-      if ('cueOut' in segment || 'cueOutCont' in segment || 'cueIn' in segment) {
-        let cueJson = {};
+      if (cue !== undefined) {
```
gesinger: When checking for undefined, I think our linter expects `typeof cue === 'undefined'`

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+        if ('cueIn' in segment) {
+          // Found a CUE-IN so end the cue
+          if (mediaTime !== cue.endTime) {
+            // If the end time does not match the current mediaTime, a CUE-IN may have
+            // been inserted earlier than expected, so the cue should be updated.
+            cue.endTime = mediaTime;
+            cue.adEndTime = mediaTime;
+          }
+          mediaTime += segment.duration;
+          cue = undefined;
```
gesinger: Better to set to null instead of undefined. Or if you want undefined, I think our linting expects `void 0`

src/master-playlist-controller.js
gesinger: Also, for this check, whether undefined or null, checking for falsey should be fine. `if (!cue)`

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
-        // lose it as an active cue by the time a user retrieves the active cues).
-        this.cueTagsTrack_.addCue(new window.VTTCue(mediaTime,
-                                                    mediaTime + 0.5,
-                                                    JSON.stringify(cueJson)));
+        if ('cueOutCont' in segment) {
+          // Entered into the middle of an ad cue
+          let adOffset;
+          let adTotal;
+
+          [adOffset, adTotal] = segment.cueOutCont.split('/').map((value) => {
```
gesinger: Are we OK making the assumption here that all cueOutCont tags will be of the format #EXT-X-CUE-OUT-CONT:5/10 ? If we are, we should add a comment that we expect that form, and add that to the README

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
 
-      if ('cueOut' in segment || 'cueOutCont' in segment || 'cueIn' in segment) {
-        let cueJson = {};
+      if (cue !== undefined) {
 
-        if ('cueOut' in segment) {
-          cueJson.cueOut = segment.cueOut;
+        if ('cueIn' in segment) {
+          // Found a CUE-IN so end the cue
+          if (mediaTime !== cue.endTime) {
```
gesinger: If this always relies on mediaTime, do we need a check?

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+          i--;
+          continue;
+        }
+
+        if ('cueOut' in segment) {
+          cue = new window.VTTCue(mediaTime,
+                                  mediaTime + segment.duration,
+                                  '');
+          cue.adStartTime = mediaTime;
+          cue.adEndTime = mediaTime + parseFloat(segment.cueOut);
```
gesinger: Same comment as below regarding assumption around the format of the CUE-OUT

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+        if ('cueOutCont' in segment) {
+          // Entered into the middle of an ad cue
+          let adOffset;
+          let adTotal;
+
+          [adOffset, adTotal] = segment.cueOutCont.split('/').map((value) => {
+            parseFloat(value);
+          });
+
+          cue = new window.VTTCue(mediaTime,
```
gesinger: Should we be creating another cue on cueOutCont? Wouldn't the cue be an extension of the above created cueOut cue?

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+
+        // otherwise extend cue until a CUE-IN is found
+        cue.endTime += segment.duration;
+
+      } else {
+        // Since the cues will span for at least the segment duration, adding a fudge
+        // factor of half segment duration will prevent duplicate cues from being
+        // created when timing info is not exact (e.g. cue start time initialized
+        // at 10.006677, but next call mediaTime is 10.003332 )
+        cue = findAdCue(track, mediaTime + (segment.duration / 2));
```
gesinger: The code may be easier to read if, instead of doing the find here and decrementing i to redo the loop, move a conditional check up to do

```
if (!cue) {
  cue = findAdCue(...);
}
```

around the beginning of the loop, then continue with the rest of the code.

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
+          }
+          mediaTime += segment.duration;
+          cue = undefined;
+          continue;
         }
-        if ('cueOutCont' in segment) {
-          cueJson.cueOutCont = segment.cueOutCont;
+
+        if (mediaTime < cue.endTime) {
+          // Already processed this mediaTime for this cue
```
gesinger: It may just be me, but the wording here threw me off a bit. Are you saying that we've already adjusted the current cue's end time to account for this segment's duration on a prior call to updateCues_?

src/master-playlist-controller.js
```diff
@@ -831,40 +847,81 @@ export default class MasterPlaylistController extends videojs.EventTarget {
...
   }
 
-  updateCues_(media) {
+  updateCues_(media, offset = 0) {
     if (!this.useCueTags_ || !media.segments) {
       return;
     }
 
-    while (this.cueTagsTrack_.cues.length) {
-      this.cueTagsTrack_.removeCue(this.cueTagsTrack_.cues[0]);
```
gesinger: I see that we're never removing cues now. Are we OK with the assumption that different playlists will always have cues at the same segments and times (included in this is that segments have the same duration across playlists)?

test/master-playlist-controller.test.js
```diff
@@ -767,35 +772,189 @@ QUnit.test('update tag cues', function() {
...
+  QUnit.equal(cueTagsTrack.cues.length, 1,
+    'adds a single cue for entire ad when entering mid cue-out-cont');
+
+  testCue = cueTagsTrack.cues[0];
+  QUnit.equal(testCue.startTime, 0, 'cue starts at 0');
+  QUnit.equal(testCue.endTime, 20, 'cue ends at start time plus duration');
+  QUnit.equal(testCue.adStartTime, -10, 'cue ad starts at -10');
+  QUnit.equal(testCue.adEndTime, 20, 'cue ad ends at 20');
+
+  cueTagsTrack.removeCue(testCue);
```
gesinger: I don't believe you have to remove the cue at the end of the test

gesinger:
A couple tests that would be good to add:
- Some tests for findAdCue (it is a simple function, but always good to have tests for it, even though it is implicitly tested by the updateCues_ tests)
- Tests with multiple ad cues

src/master-playlist-controller.js
mjneil: This is the case where in a live stream, you enter the stream and the first tag you see is a cueOutCont. There would be no cueOut to extend to since we haven't seen a cue out.

src/master-playlist-controller.js
mjneil: Yes that is correct.

For example, if in the stream we have this playlist

```
---------------------
  O      C1      C2
```
and then we load another playlist as the stream progresses and the sliding window moves
```
---------------------
  C1     C2      C3
```
The duration of the segments with tags `C1` and `C2` have already been added to the cue's end time, so we want to skip over that segment

src/master-playlist-controller.js
mjneil: I think its ok to make the assumption for now since the parsing only affects `adStartTime` and `adEndTime` which don't actually affect the functionality of the cue, but only give some extra metadata to the user. `parseFloat` will also ignore anything after it sees the first non-numerical character, so as long as that data is first, it won't be affected.

imbcmdth:
I think this should be moved so that there are a minimal number of changes to `master-playlist-controller.js` which has already grown into an unwieldy large file.

src/ad-cue-tags.js
```diff
@@ -0,0 +1,107 @@
...
+    let cue = cues[i];
+
+    if (mediaTime >= cue.adStartTime && mediaTime <= cue.adEndTime) {
+      return cue;
+    }
+  }
+  return null;
+};
+
+const initAdCueTrack = function(controller) {
```
gesinger: It's a bit odd to pass the caller to this function so that it can set a value on the caller. If we want to separate responsibility, we can either create an AdCueTagManager (or something similar) that maintains its own state (keeping hold of the cueTagsTrack) with MasterPlaylistController having only an instance of the manager, or create the cue tags track in MasterPlaylistController (move the innards of this function to MasterPlaylistController) and continue to pass in that reference as you do later on. I think the latter might make the most sense, as these functions are merely utility functions, and MasterPlaylistController is the one who should know a bit more about its state, and what gets added to Video.js type APIs (i.e., the new text track).

src/ad-cue-tags.js
```diff
@@ -0,0 +1,107 @@
...
+    if (!cue) {
+      // Since the cues will span for at least the segment duration, adding a fudge
+      // factor of half segment duration will prevent duplicate cues from being
+      // created when timing info is not exact (e.g. cue start time initialized
+      // at 10.006677, but next call mediaTime is 10.003332 )
+      cue = findAdCue(track, mediaTime + (segment.duration / 2));
+    }
+
+    if (cue) {
+
```
gesinger: Extra space here (and at the end of the block)

test/ad-cue-tags.test.js
```diff
@@ -0,0 +1,292 @@
...
+      },
+      {
+        duration: 10
+      }
+    ]
+  }, this.track);
+
+  QUnit.equal(this.track.cues.length, 2, 'correctly created 2 cues for the ads');
+  QUnit.equal(this.track.cues[0].startTime, 30, 'cue created at correct start time');
+  QUnit.equal(this.track.cues[0].endTime, 60, 'cue created at correct end time');
```
gesinger: Minor note, but phrasing for end times seems a bit off

Also, while they should be the same, can't hurt to check adStartTimes and adEndTimes as well

test/ad-cue-tags.test.js
```diff
@@ -0,0 +1,292 @@
...
+      adEndTime: 55
+    },
+    {
+      adStartTime: 100,
+      adEndTime: 120
+    }
+  ];
+
+  let cue;
+
```
gesinger: Also worth it to have one more checking that it matches start time.

src/master-playlist-controller.js
```diff
@@ -62,7 +62,8 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     this.mode_ = mode;
     this.useCueTags_ = useCueTags;
     if (this.useCueTags_) {
-      this.cueTagsTrack_ = this.tech_.addTextTrack('metadata', 'hls-segment-metadata');
+      this.cueTagsTrack_ = this.tech_.addTextTrack('metadata',
+        'hls-segment-metadata');
```
imbcmdth: We should rename this TextTrack to 'ad-cues' as it no longer contains 'hls-segment-metdata'.

src/segment-loader.js
```diff
@@ -503,14 +503,16 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+   *
+   * @param {Object} segmentInfo - the current segment
+   * @returns {Number} removeToTime - the end point in time, in seconds
+   * that the the buffer should be trimmed.
    */
-  loadSegment_(segmentInfo) {
-    let segment;
-    let keyXhr;
-    let segmentXhr;
+  trimBuffer_(segmentInfo) {
```
imbcmdth: I think a rebase pulled in something it should not have.. this code appears to be from the last commit in master.

test/videojs-contrib-hls.test.js
```diff
@@ -2016,7 +2016,9 @@ QUnit.test('cleans up the buffer when loading live segments', function() {
   QUnit.strictEqual(this.requests[0].url, 'liveStart30sBefore.m3u8',
                     'master playlist requested');
   QUnit.equal(removes.length, 1, 'remove called');
-  QUnit.deepEqual(removes[0], [0, seekable.start(0)],
+  // segment-loader removes up to the segment prior to seekable.start
+  // to avoid crossing segment-boundaries
+  QUnit.deepEqual(removes[0], [0, seekable.start(0) - 10],
```
imbcmdth: This also came from that last commit in master.

PR #810, opened 2016-08-08
Change timing of SegmentLoader abort on media change

mjneil:
## Description
Quickly going in and out of fullscreen causes a race condition that causes the SegmentLoader to get stuck in the `'WAITING'` state after a media change. This results in never requesting more segments. Aborting the SegmentLoader before pausing corrects this issue.

## Requirements Checklist

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -155,14 +155,13 @@ export default class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.masterPlaylistLoader_.on('mediachanging', () => {
+      this.mainSegmentLoader_.abort();
       this.mainSegmentLoader_.pause();
     });
 
     this.masterPlaylistLoader_.on('mediachange', () => {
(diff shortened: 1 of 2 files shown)


dmlap:
It seems wrong to me that the ordering of these two calls on segment loader causes a problem. What changes would be required to segment loader so that this code wouldn't be so sensitive?

mjneil:
We could set the state of the segmentLoader to 'READY' on an `abort()` every time instead of only when it is not paused.

mjneil:
Looking into this more, the way segment loader currently handles being paused and unpaused is a big part of the reason why aborting before pausing solves this. We could change the state logic within segment loader to allow a paused state. This would make the state of the segment loader much more clear and hopefully less error prone in the long run. I have a branch with the proposed state changes [here](https://github.com/mjneil/videojs-contrib-hls/tree/pause-states). The race condition is the above fixes no longer appears regardless of the order of calls with the more defined states.

PR #820, opened 2016-08-15
SegmentLoader States

mjneil:
## Description

Changes the state logic in `SegmentLoader` so that the states are explicit and not combined together. This will solve issues being addressed in #810 
[Proposed State Diagram](https://docs.google.com/drawings/d/1nJJFh9Yq6cZXmjUL8suy5hVq6Y53LwzO6RczVMSxFRY/edit?usp=sharing)
## Specific Changes proposed

Remove the concept of being PAUSED and in another state such as READY or WAITING.
Adds three states, `PAUSED`, `PAUSING` and `PRELOAD`
- `PAUSED` indicates the loader has paused monitoring of the buffer
- `PAUSING` indicates that the loader will move to the `PAUSED` state once it has finished its current outstanding request.
- `PRELOAD` indicates the loader can begin fetching segments early if it has enough information despite possibly not being fully initialized.
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -187,10 +187,11 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    if (this.state === 'WAITING') {
+    if (this.state === 'PRELOAD') {
+      this.state = 'INIT';
+    } else if (this.state === 'WAITING') {
       this.state = 'READY';
-    }
-    if (this.state === 'PAUSING') {
+    } else if (this.state === 'PAUSING') {
       this.state = 'PAUSED';
     }
 
@@ -256,7 +257,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // buffering now
     if (this.sourceUpdater_ &&
         media &&
-        this.state === 'INIT') {
+        this.state === 'PRELOAD') {
       this.state = 'READY';
       return this.fillBuffer_();
     }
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -746,11 +764,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     let segmentInfo;
     let segment;
 
-    this.state = 'DECRYPTING';
-
     segmentInfo = this.pendingSegment_;
     segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex];
 
+    // this.state = 'DECRYPTING';
```
imbcmdth: Remove these commented lines completely.

dmlap:
This doesn't make things clearer for *me*, but if you guys think it's easier to work with I don't object.

dmlap:
What happens when you call pause() in the init  state?

gkatsev:
Make sure that links are publicly available

mjneil:
Calling pause in the init states doesn't have an effect and will stay in the init state. The loader is still initializing and does not have all the information needed to load segments.

dmlap:
Could you elaborate on the purpose of the `PRELOAD`? What information is required to move the loader into `READY`?

mjneil:
`PRELOAD` is similar to `INIT`. Both require a SourceUpdater and a playlist. When in the `PRELOAD` state, the loader will begin fetching segments as soon as both have been set, where as in the `INIT` state, a call to `load()` is required even after both have been set.

dmlap:
@mjneil so INIT is basically PAUSE? :)

@gesinger seems like a reasonable suggestion to me.

mjneil:
@gesinger Is the suggestion that calling pause would abort the current process and pause from future loads instead of just pausing from future loads? Once we send the segment to the source buffer for appending, PAUSE can no longer be an immediate action since we have to finish the processing of metadata post append. The current issues with the loader are not during this time period so this is ok, however we would either need to keep some notion of a PAUSING state or go back to not having PAUSED or PAUSING be an explicit loader state. 

@dmlap yes, with the slight difference of not having a SourceUpdater or playlist. If we go with @gesinger's suggestion with taking PAUSED and PAUSING back out of the states, we can make the INIT state more clear.

PR #822, opened 2016-08-16
Changes to the underflow-detection in the gap-skipper to be less restrictive about the size of the gaps it is able to skip when an underflow is detected

imbcmdth:
## Description
Right now the gap-skipper detects video underflow conditions on Chrome but only skips then when there is a gap in the buffer that is less than 1 second long. There are conditions under which the gap can be greater than 1 second (a GOP being removed from the coded-frame buffer for instance) and this change allows the detection and skipping of gaps that exceed 1 second.

## Requirements Checklist
  - [ ] Docs/guides updated


gesinger:
LGTM

PR #824, opened 2016-08-18
Prevent loading first segment of live manifest - seek to and load segment at current time

marguinbc:
## Description
Currently in setupFirstPlay() we are performing a load() followed by a seek to the current live point in the manifest. The first segment is never played, resulting in wasted bandwidth as the segment is downloaded and ignored. Speaking to core contributors, this logic appears to be left over from when earlier versions of FF could not seek without first loading some segment and has since been fixed.

This suggested fix moves the load() call to immediately after setting the seekable range to the current Time to avoid loading the first segment. This results in loading the first segment at the current point in time.

Tests were also adjusted to reflect this new behavior. 

## Specific Changes proposed
Moving load() after seeking in setupFirstPlay()

## Requirements Checklist
- [ x ]  Bug fixed
  - [ x ] Unit Tests updated or fixed
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "videojs-contrib-hls",
-  "version": "3.5.1",
+  "version": "3.5.2",
   "description": "Play back HLS with video.js, even where it's not natively supported",
   "main": "es5/videojs-contrib-hls.js",
   "engines": {


gesinger:
@marguinbc looks like there's a linting error for trailing spaces.

marguinbc:
Thanks @gesinger! I found and fixed it, but now I am behind master again so I need to rebase, etc.

marguinbc:
Think I fixed my branch - reopening

PR #838, opened 2016-08-30
Cleanup mocked selectPlaylist

dmlap:
Leaving a mock version of selectPlaylist on the HlsHandler prototype leads to broken error handling in tests for encryption key fetching. Fixes current Travis test failure.


BrandonOCasey:
should we just set this back to the original playlist selector rather than deleting?

dmlap:
@BrandonOCasey there is no playlist selector on the prototype before this test runs. While I know `delete` is a no-no, this resets the prototype back to the state it was in before the test started mucking with it.

gesinger:
LGTM

PR #841, opened 2016-09-06
Replace linear search with binary search to incorporate large live streams

SuriGill:
## Description

The VideoJS player was slowing down and buffering in large HLS streams with small GOP sizes (1 hour+ streams). I traced the results to the buffer and more specifically how the buffer was being trimmed. I ended up replacing the linear search to find the time segment to trim the buffer for large streams with a binary search and the results are a lot faster.  
## Specific Changes proposed

Replacing linear search with binary search to improve the amount of iterations taken to find the time segment to trim the buffer at. 
## Requirements Checklist
  - [x] Docs/guides updated
  - [x] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))

… discard previous content on buffer for large scale hls streams

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -525,35 +525,42 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     // If we have a seekable range use that as the limit for what can be removed safely
     // otherwise remove anything older than 1 minute before the current play head
-    if (seekable.length &&
-        seekable.start(0) > 0 &&
-        seekable.start(0) < currentTime) {
+    if (seekable.length && seekable.start(0) > 0 && seekable.start(0) < currentTime) {
       removeToTime = seekable.start(0);
     } else {
       removeToTime = currentTime - 60;
     }
-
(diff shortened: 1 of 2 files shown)


mjneil:
lgtm

SuriGill:
Also you might see I had to change one of the tests because the VOD was actually 40 seconds long but it was trying to trim the buffer at 60 seconds. I'm guessing that was a bug because the buffer shouldn't have even loaded to that point when you skipped to 120 seconds even though it is 60 seconds before as the test describes. The live buffer trim test has the same expected results.

forbesjo:
Looks good to me

imbcmdth:
@SuriGill Thanks for the submission!

We put a lot of work into the areas around this search and the result was that we were able to remove it completely. As a result, the search performance is no longer an issue in the *latest versions* of videojs-contrib-hls. we have tested with streams that were 5 days (!!) long and saw no performance problems seeking around.

PR #842, opened 2016-09-08
Fix 'ended' event not firing after replay

mjneil:
## Description
The "ended" event fires when the video reaches to the end of the video for the first time (This is fine). However on a reply of the video (without refreshing the page), it does playback, but when the video reaches to the end of the video, it stops at the last frame and "ended" event does not fire.

## Specific Changes proposed
If the fetcher logic chose `mediaIndex === playlist.segments.length` as the next segment to fetch, the playlist is not live, the last segment in the playlist is already buffered and the media source has a ready state of `open`, then call `endOfStream()` on the media source to signal that its ready state should be changed to `ended`.

## Requirements Checklist

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -380,7 +380,6 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     let bufferedTime;
     let currentBufferedEnd;
-    let segment;
     let mediaIndex;
 
     if (!playlist.segments.length) {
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -399,25 +403,39 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        let endOfBuffer = buffered.end(buffered.length - 1);
+        let lastSegment = playlist.segments[playlist.segments.length - 1];
+
+        if (lastSegment.end &&
+            lastSegment.end < endOfBuffer + Ranges.TIME_FUDGE_FACTOR) {
+          return { isEndOfStream: true };
+        }
+      }
+      return null;
+    }
```
imbcmdth: Can we add the slight differences from this code to the code above for `detectEndOfStream` and move this check from here to `fillBuffer_`?

src/segment-loader.js
```diff
@@ -363,10 +363,14 @@ export default class SegmentLoader extends videojs.EventTarget {
...
    * @param {Number} currentTime - the playback position in seconds
+   * @param {Boolean} hasPlayed - if the player has played before
+   * @param {Number} expired - the seconds expired off the playlist
+   * @param {Number} timeCorrection - correction value to add to current time when
+   *  when determining media index to use
    * @returns {Object} a segment info object that describes the
    * request that should be made or null if no request is necessary
    */
-  checkBuffer_(buffered, playlist, currentTime) {
+  checkBuffer_(buffered, playlist, currentTime, hasPlayed, expired, timeCorrection) {
```
imbcmdth: I like that you made `checkBuffer_` no longer rely on `this`...

forbesjo:
LGTM

PR #844, opened 2016-09-14
Browserify and webpack tests

ruchigupta19:
Added tests for Browserify/Webpack


mjneil:
This should be updated to work like the contrib-media-sources [pull request](https://github.com/videojs/videojs-contrib-media-sources/pull/103)

PR #855, opened 2016-10-03
allow for initial bandwidth option of 0

mjneil:
## Description
Setting initial bandwidth option to zero uses default bandwidth and not zero

## Specific Changes proposed
Can now set to zero.

## Requirements Checklist

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -342,7 +342,7 @@ class HlsHandler extends Component {
     // start playlist selection at a reasonable bandwidth for
     // broadband internet
     // 0.5 MB/s
-    if (typeof this.options_.bandwidth === 'undefined') {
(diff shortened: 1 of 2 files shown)


src/videojs-contrib-hls.js
```diff
@@ -342,7 +342,9 @@ class HlsHandler extends Component {
     // start playlist selection at a reasonable bandwidth for
     // broadband internet
     // 0.5 MB/s
-    this.options_.bandwidth = this.options_.bandwidth || 4194304;
+    if (typeof this.options_.bandwidth === 'undefined') {
```
imbcmdth: People can put all sorts of terrible things in the configuration. I think checking if `bandwidth` is a number is a safer approach here.

imbcmdth:
Also this should have a test.

imbcmdth:
LGTM

Issue #776, opened 2016-07-13
While doing HLS AES 128 playback, the player request key for every chunk.


maddygoround:
## Description

When working with HLS 128 encrypted stream, the module make request for key for every chunk.
Why we need to make request every-time for chunk when we already have the key from the first request to key.

Following is the key object already in the segment.

`"{"method":"AES-128","uri":"10minscat.key","iv":{"0":89807546,"1":1232804682,"2":2466904456,"3":792036832},"resolvedUri":"http://server-url/10minscat.key","bytes":{"0":664287445,"1":1803566766,"2":1398338856,"3":643511093}}"`

Following is the screenshot of segment calls
![request](https://cloud.githubusercontent.com/assets/7420749/16804071/869a6e90-4927-11e6-8866-be24af1c9db2.png)

On other end streamroot HLS plugin makes the request for key only once.
https://github.com/streamroot/videojs5-hlsjs-source-handler
### Expected

We expect the calls to the key only once.
### Actual

The call for the key is done for each and every chunk.
### videojs-contrib-hls

videojs-contrib-hls 3.1.0
#### videojs

video.js 5.10.4
#### browsers

Chrome
*
#### Operating Systems

Windows10
*
### plugins

https://github.com/streamroot/videojs5-hlsjs-source-handler

gkatsev:
Do you have a link to your m3u8 or an example m3u8?

maddygoround:
Here it is - http://djyjlsv34fx05.cloudfront.net/10minscat/10minscat.m3u8

gkatsev:
Looks like a bug.

maddygoround:
good to know.

ellyjonez:
Any movement here? I've noticed this happening as well, on OSX Chrome and OSX Firefox. I am concerned it may be contributing to an uptick in complaints from our users about video stopping to buffer too much.

videojs version: 5.10.7
vjs-hls-contrib version: 3.5.3

here's a screenshot of what I'm seeing. (those POSTs to viewing_events.json is our event tracker at work):

<img width="721" alt="screen shot 2016-10-03 at 12 11 33 am" src="https://cloud.githubusercontent.com/assets/697972/19029848/88d0afaa-88fe-11e6-825e-90d0e2fd4acd.png">

colt082295:
+1 I'm currently having the same issue. It also seems like the server with the key (out of my control) isn't happy with requests, and throws 403 errors sometimes that eventually break the stream.

[...]

PR #865, opened 2016-10-14
Fix the issue #776

benvirus:
## Description

Fix the issue https://github.com/videojs/videojs-contrib-hls/issues/776
At first,please accept my apologies for my poor English.Thank you for your patience.
As we know When the segement is Encrypted,we will get the key for every segment.But the key is the same for the same video as usually.So the request is a waste.
I store the key in code and check if it is exist for every XMLHttpRequest.
## Specific Changes proposed
- Add a key.js to store the key we get.
- Check if the key is exist for every segement.So the segement-loader.js is changed.
## Requirements Checklist
  - [ ] Docs/guides updated
  - [ ] Example created ([Example Link](http://default.prod.dev.qiqiuyun.cn:9071/video-player/examples/demo-sdk-play-rate-display.html))

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -1,0 +1,38 @@
+/**
+ * @file key.js
+ *
+ * 
+ * Store the unique key for the video data;
+ * @type {Object}
+ * @private
+ */
+
+let keyObject = {
+    uri: '',
+    keyData: {}
+};
+
+/**
+ * Test if wo have stored the key
+ *
+ * @param  {String}  The uri of the key which we need to load;
+ * @return {Boolean} do we have stored the key which we want to load;
+ * @function getStoredKey
+ */
+export function getStoredKey(uri) {
+    if (keyObject.uri === uri) {
+        return keyObject.keyData[uri];
+    }
+    return false;
+}
+
+/**
+ * Store the key and uri
+ * @param {String} uri The uri of the key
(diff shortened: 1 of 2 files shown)


dmlap:
@benvirus do you have caching headers setup on your key URL? Also, have you turned of HTTP caching when the developer console is open? The browser should be caching keys automatically unless your headers specify otherwise.

ghaydarov:
Hi @benvirus, I implemented encryption with key, however, my links are visible and downloadable as opposed to yours, on your [example](http://default.prod.dev.qiqiuyun.cn:9071/video-player/examples/demo-sdk-play-rate-display.html) here. What are the plugins you are using for this and I am using HLSv3 on aws and serving with cloudfront cdn.  

Thanks

benvirus:
@ghaydarov We generate the m3u8 file everytime when the browser request the playlist. So we have chance to generate the key link everytime and we only allow one request to the key from every key link.The next time request from the same key link ,you will get a wrong key. For more, we encrypt our key with our own encryption algorithm and decript it in video-contrib-hls plugin.

From more ,Please email to benchen19921130@gmail.com without hesitation.

ghaydarov:
Thanks @benvirus , I emailed you about this. Thank you!

huanghanzhilian:
您好，我想问下如何使用您提交的代码？

PR #873, opened 2016-10-18
updated package.json

ruchigupta19:
fixes an error "Failed at the contextify@0.1.15 install script 'node-gyp rebuild'" which can be seen while installing dependencies.


forbesjo:
LGTM

PR #875, opened 2016-10-19
Simplified fetch algorithm as much as possible

imbcmdth:
## Description

Tore out the entire "fetch" algorithm and replaced it with something simpler and more robust (hopefully!) The process resulted in a lot of changes and the introduction of a "sync-point" mechanism. A "sync-point" is defined as a known mapping between display-time and a segment index that can be used to make a reasonable and conservative guess about what segment to load next. (It's conservative in that it always errors on the side of guessing too "early".)

Internally, we keep track of many potential sync-points and attempt the pick the one that will give us the most accurate guesses when ever we need to make a guess. Guessing is rare though and a major of the fetch-logic devolves into a simple walk forward through the playlist (`mediaIndex++`).


src/master-playlist-controller.js
```diff
@@ -689,6 +698,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.load();
 
       return true;
+    } else if (media &&
+        // 2) the video is a VOD
+        media.endList) {
+      this.hasPlayed_ = () => this.tech_.played().length !== 0;
```
imbcmdth: We might need to rethink `hasPlayed_` soon... it is complicated to get this working in all 4 cases:

* MSE & VOD
* MSE & Live
* Flash & VOD
* Flash & Live

src/segment-loader.js
```diff
@@ -354,66 +361,148 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    let segmentIndexArray = playlist.segments
+      .map((s, i) => {
+        return {
+          timeline: s.timeline,
+          segmentIndex: i
+        };
+      }).filter(s => s.timeline === this.currentTimeline_);
+
+    if (segmentIndexArray.length) {
+      return segmentIndexArray[Math.min(segmentIndexArray.length - 1, 1)].segmentIndex;
```
mjneil: This looks like if there is only one segment with the current timeline, we pick that, otherwise pick the second of the list. Why are these options better candidates than any of the other segments?

PR #876, opened 2016-10-19
Don't error if MediaSource isn't defined

dmlap:
If a CODECS attribute is present on an m3u8 but MSE isn't present, don't explode. In practice, this means only exclude variants based on codec for MSE mode.


imbcmdth:
LGTM

PR #877, opened 2016-10-21
use setTimeout in gap skipper instead of relying on timeupdate events

mjneil:
## Description
gap skipper previously relied on `timeupdate` events from the tech to detect when the player encounters a gap. However, if the tech is using native `timeupdate` events instead of manual ones like html does in the latest release of video.js, `timeupdate` events do not fire when the player stalls at a gap.

## Specific Changes proposed
use `setTimeout` within gap skipper to periodically check the current time to detect a gap instead of relying on `timeupdate` events from the tech.

## Requirements Checklist

### DIFF
--- a/src/gap-skipper.js
+++ b/src/gap-skipper.js
@@ -38,7 +38,7 @@ export default class GapSkipper {
     this.consecutiveUpdates = 0;
     this.lastRecordedTime = null;
     this.timer_ = null;
-    this.timeupdateTimeout_ = null;
+    this.checkCurrentTimeTimeout_ = null;
 
     if (options.debug) {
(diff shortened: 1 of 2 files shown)


imbcmdth:
LGTM as well

src/gap-skipper.js
```diff
@@ -38,31 +38,49 @@ export default class GapSkipper {
...
       this.cancelTimer_();
     };
   }
 
   /**
+   * Periodically check for timeupdates to see if a gap has been encountered.
+   *
+   * @private
+   */
+  checkTimeupdate_() {
```
mjneil: Made some rename changes to better reflect what they do

dmlap:
Ran tests on my machine since Travis is running into DNS issues, all good.

PR #878, opened 2016-10-21
Fixed preload="none" behavior and reduced the incidence of repeating initial segment requests

imbcmdth:
Changes to the simplified SegmentLoader to fix preload="none" never starting playback in a VOD. Also fixed an issue where we were resetting the SegmentLoader too often which manifest as downloading the first segment twice in the live-stream case.


forbesjo:
Looks good to me.

PR #884, opened 2016-10-25
Make sure packaged bundle works when used in other Browserify build systems.

arendjr:
## Description

This change includes browserify-derequire in the build step so that the resulting bundle can be used by other projects that using Browserify. Without this patch, the Browserify toolchain in other projects would pick up the `require()` calls in the bundle and attempt to resolve them, which will lead to failures.

For more info, please see: https://www.npmjs.com/package/browserify-derequire
## Specific Changes proposed

Add browserify-derequire to the build system.
## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
That seems like a reasonable change, however, why are you using the pre-build file in browserify? This project is built so that you should be able to just require it.

arendjr:
What do you mean by "pre-build" file? I'm taking the one that's built in the `dist/` directory, just not the one that's minified. I prefer to use the non-minified versions of libraries for development purposes, and I have my own build system minify them when I'm going to production.

gkatsev:
Yeah, why is that file (`dist/`) being used in a browserify build-system? This project is set up so that you can require it (https://github.com/videojs/videojs-contrib-hls/blob/master/package.json#L5) and when you run browserify on your own code, everything gets bundled together. Just trying to understand the usecase better.

arendjr:
Well, that's a good question. It was my understanding if I want to use it like that, I should install the project through NPM. But that doesn't seem very attractive as soon as I want to customize my video.js build, because the NPM version of videojs-contrib-hls will automatically pull in the stock version of video.js. So that's why I went this route. But I admit I'm just getting started setting the whole thing up, so if I'm missing something, I'd gladly hear what'd be an easier setup :)

gesinger:
Hey @arendjr , thanks for the PR, but if (as @gkatsev mentioned) the project is required through a browserify build system, the addition is not required, as it is all handled by browserify. We recommend in these cases not requiring from the `dist` dir directly, and instead using a browserify based build system. Thanks.

PR #888, opened 2016-10-28
Persist representations enable function and separate blacklisting from disabling

mjneil:
## Description

This PR includes two main fixes.
1. I've updated the closure for `Representation`'s enabled function to keep reference to the `PlaylistLoader` instead of the playlist object itself. This is because in `PlaylistLoader` overwrites the playlists objects with a new reference when it loads a playlist it hasnt loaded previously. This would create a new reference to an enabled function making the previous reference a user may have held onto obsolete.
2. Previously, the `Representation` enabled function would set `playlist.excludeUntil = Infinity` when disabling a playlist. However, when `MasterPlaylistController` blacklists playlists, it also uses `excludeUntil` to disable selection of that playlist. Since`Representation`s had no notion on whether a playlist was disabled through the enabled function or through blacklisting, it was possible to enable a previously blacklisted playlist, which could cause a player crash.

This change is also required for functionality of videojs-contrib-quality-levels
## Specific Changes proposed

`Representation`'s enabled function now creates closure around the `PlaylistLoader` object instead of the playlist object so the reference is not lost on new playlist loads.

Added a `disabled` property to playlist objects when using `Representation`'s enabled function so that manual enabling/disabling does not affect the blacklist status of a playlist.
## Requirements Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "url-toolkit": "^1.0.4",
     "video.js": "^5.10.1",
     "videojs-contrib-media-sources": "^4.1.2",
+    "videojs-contrib-quality-levels": "^1.0.0",
     "videojs-swf": "^5.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


gesinger:
LGTM

mjneil:
Ive introduced new things to this PR so previous LGTM are void

mjneil:
Going to close this and open a new PR as the scope of this PR has drastically changed

PR #893, opened 2016-11-04
Live falloff

gesinger:
## Description
Seeks to seekable end if we detect that we fell off the back end of a live playlist.

## Requirements Checklist
  - [ ] Docs/guides updated


src/always-be-playing.js
```diff
@@ -135,6 +121,71 @@ export default class GapSkipper {
...
+        this.seekedAtTime === currentTime) {
+      return;
+    }
+
+    // check to see if we fell out of the live window
+    if (playlist &&
+        !playlist.endList &&
+        seekable.length &&
+        currentTime < seekable.start(0)) {
+      this.seekedAtTime = currentTime;
```
mjneil: Do we actually need to keep track of `seekedAtTime`? I can't think of a scenario where we will reach the check `this.seekedAtTime === currentTime` and have that be true

src/always-be-playing.js
```diff
@@ -135,6 +121,71 @@ export default class GapSkipper {
   }
 
   /**
+   * Handler for `waiting` events from the player
+   *
+   * @private
+   */
+  waiting_() {
```
mjneil: There is a lot going on in this function. I think it would be helpful for readability if the different checks ("check to see if we fell out of the live window", "check for video underflow", "check for gap") were made out into their own functions

gesinger: Good call. Will try for that. Will make unit testing easier as well.

gesinger: It could be a relic of some old changes, but I remembered seeing waiting fired twice. I will switch it out and see if I can uncover why that was happening instead.

src/always-be-playing.js
gesinger: It could be a relic of some old changes, but I remembered seeing waiting fired twice. I will switch it out and see if I can uncover why that was happening instead.

src/always-be-playing.js
```diff
@@ -65,11 +57,15 @@ export default class GapSkipper {
   }
 
   /**
-   * Periodically check for timeupdates to see if a gap has been encountered.
+   * Periodically check for timeupdates
```
mjneil: Might be better to say "Periodically check the currentTime" since we are no longer using timeupdate events

mjneil:
LGTM

imbcmdth:
Added via #899

PR #894, opened 2016-11-04
Add playback tests

forbesjo:
This is the start to automating end to end playback tests.


test/playback.test.js
```diff
@@ -0,0 +1,74 @@
+import QUnit from 'qunit';
+import videojs from 'video.js';
+/* eslint-disable no-unused-vars */
+import { Hls } from '../src/videojs-contrib-hls';
+
+let when = function(element, type, cb, condition) {
+  element.on(type, function func() {
```
gkatsev: why not use `element.one`?

forbesjo: I want the function to fire when the targeted event is triggered and a certain condition is met. Using `one` will just remove it the first time the event is fired.

For example, I use this with `timeupdate` and some target time.

gkatsev: Oh, totally missed that. I guess I was focused on less code :P

forbesjo: Actually I might just use cue points since I haven't really used this function for anything other than doing something at a specific time.

test/playback.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+QUnit.test('Advanced Bip Bop', function() {
+  QUnit.expect(2);
+  QUnit.stop();
+  let player = this.player;
+
+  player.autoplay(true);
+
+  playFor(player, 2, function() {
+    QUnit.ok(true, 'played for three more seconds');
```
mjneil: should this be `'played for ~~three~~ two more seconds'` ?

forbesjo:
Whoops that's a leftover from something I was trying before

test/playback.test.js
forbesjo: That's what I get for copy/paste

test/playback.test.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+QUnit.module('Playback', {
+  beforeEach() {
+    QUnit.stop();
+    let video = document.createElement('video');
+
+    video.width = 600;
+    video.height = 300;
+    document.querySelector('#qunit-fixture').appendChild(video);
```
mjneil: Probably not an issue but do we want to remove the element from the DOM between tests?

forbesjo: qunit will automatically reset the fixture it provides https://qunitjs.com/cookbook/#keeping-tests-atomic

test/playback.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+let playFor = function(player, time, cb) {
+  let targetTime = player.currentTime() + time;
+
+  when(player, 'timeupdate', cb, () => player.currentTime() >= targetTime);
+};
+
+QUnit.module('Playback', {
+  beforeEach() {
+    QUnit.stop();
```
gesinger: Should we use async and done instead of stop and start?

gkatsev: yes.

forbesjo: I'll update Qunit

mjneil:
LGTM

PR #898, opened 2016-11-08
remove expired time tracking and use sync points to calculate seekable

mjneil:
## Description
Tracking expired time in live playback can be buggy and inaccurate. This change removes tracking expired time in playlist loader completely in favor of using sync points generated by the SyncController to calculate the amount of expired time. 

There is also a bugfix for autoplay on live playlists that are not a master playlist. Because interacting with the tech can be asynchronous, if a source was just a live media playlist and not a master playlist, if autoplay was turned on, a race condition would happen where we would try to call `addSeekableRange()` on the mediasource before the duration was properly set to `Infinity`

## Specific Changes proposed
 - Removed `expired_` and `trackExpiredTime_` from `PlaylistLoader`. This has the added benefit of not needing to listen to `firstplay` within `PlaylistLoader`.
 - Now check to see if `mediaSource.duration` is equal to `Infinity` before trying to call `mediaSource.addSeekableRange()`. If the duration is not `Infinity`, we wait and listen for the `durationchange` event on the tech to call `addSeekableRange()` once the duration is `Infinity`.
- `MasterPlaylistController` now caches the seekable range instead of calculating it every time `seekable()` is called. It now listens for `syncinfoupdate` events on the `SegmentLoader`s (which are listening for the same event on their `SyncController`) and will recalculate the seekable range when `syncinfoupdate` is fired.

## Requirements Checklist


imbcmdth:
The `addSeekableRange` function was based on a API proposal by @dmlap that wasn't accepted as-is. The replacement has a different function that performs similarly. I wonder if this is a good time to change the names in videojs-contrib-media-sources and here to match the new api in the MSE specification.

AFAIK, no browser implements the new API natively though we should come up with a plan to support the native API incrementally as it becomes available.

## API
* [setLiveSeekableRange](https://w3c.github.io/media-source/#dom-mediasource-setliveseekablerange)
* [clearLiveSeekableRange](https://w3c.github.io/media-source/#dom-mediasource-clearliveseekablerange)

src/playlist.js
```diff
@@ -232,14 +232,15 @@ export const sumDurations = function(playlist, startIndex, endIndex) {
...
   let endSequence;
-
-  if (typeof expired !== 'number') {
-    expired = 0;
-  }
+  let expiredSyncPoint;
+  let firstSyncPoint;
+  let expired;
+  let expiredStart;
+  let segments;
```
imbcmdth: Wow, that is a lot of state for one function! 😮  I feel that at that point it might make more sense to break this into a few smaller functions.

Also the same comment as above about creating these inline.

src/playlist.js
```diff
@@ -250,6 +251,51 @@ export const seekable = function(playlist, expired) {
...
+      break;
+    }
+  }
+
+  // We have no sync information for this playlist so we can't create a seekable range
+  if (!expiredSyncPoint && !firstSyncPoint) {
+    return createTimeRange();
+  }
+
+  // If we dont have an expired sync point, estimate expired time based on first sync,
```
gesinger: Just my opinion, but this comment seems to just repeat a lot of the logic, and is more at risk of becoming outdated/losing precision. Better to have individual comments within each of the logical blocks, closer to their relevant code, specifying high level concepts of what is going on.

gesinger: And it seems like the comment is mixed (order-wise) from the conditionals below.

src/master-playlist-controller.js
```diff
@@ -278,7 +279,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.masterPlaylistLoader_.on('loadedplaylist', () => {
       let updatedPlaylist = this.masterPlaylistLoader_.media();
-      let seekable;
+      let onDurationchange;
+      let addSeekableRange;
```
mjneil: Good point, old habits die hard

src/master-playlist-controller.js
```diff
@@ -371,6 +390,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.blacklistCurrentPlaylist(this.mainSegmentLoader_.error());
     });
 
+    this.mainSegmentLoader_.on('syncinfoupdate', () => {
+      this.onSyncInfoUpdate_();
+    });
+
+    this.audioSegmentLoader_.on('syncinfoupdate', () => {
+      this.onSyncInfoUpdate_();
+    });
```
mjneil: The current implementation of seekable uses both the main playlist and audio playlist, taking the intersection of the two as the seekable range, so I wanted to keep behavior as clsoe as possible. If the main playlist had a seekable window of 50 - 150 seconds, but the audio playlist had a seekable window of 75 - 125 seconds, I think it would make sense to not allow seeking outside of the audio range.

src/master-playlist-controller.js
```diff
@@ -846,39 +866,48 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      return;
     }
 
     if (this.audioPlaylistLoader_) {
-      audioSeekable = Hls.Playlist.seekable(this.audioPlaylistLoader_.media(),
-                                            this.audioPlaylistLoader_.expired_);
+      audioSeekable = Hls.Playlist.seekable(this.audioPlaylistLoader_.media());
       if (audioSeekable.length === 0) {
-        return audioSeekable;
+        return;
```
mjneil: I guess this depends on if we want seekable to depend on the audio playlist seekable range ( see conversation above with @imbcmdth ). If the answer is yes, then if the audio playlist doesn't have a seekable range, then we probably don't want the player to think it has a seekable range when there is no seekable range for the alternate audio

src/playlist-loader.js
```diff
@@ -534,73 +519,4 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
 
 PlaylistLoader.prototype = new Stream();
 
- /**
-  * Update the PlaylistLoader state to reflect the changes in an
-  * update to the current media playlist.
-  *
-  * @param {Object} update the updated media playlist object
-  */
-PlaylistLoader.prototype.updateMediaPlaylist_ = function(update) {
```
mjneil: It felt pretty good to delete this

src/playlist.js
mjneil: I'll reword/structure the comments here to make more sense. Just to clarify, the only issue here is the comments, or is there issues in the code logic as well that is at risk of losing precision?

src/master-playlist-controller.js
gesinger: I think you're right (on this and the comment above), we do want the intersection if an audio playlist is available.

src/playlist.js
gesinger: I believe it is just the comments, as I think it makes sense to prioritize firstSyncPoint.

src/master-playlist-controller.js
imbcmdth: Sounds good. I think long run we have to look at combining the separate SyncControllers between segment loaders so that they can use the same information when choosing and placing segments. Specifically, I am thinking about the mapping of timestamps to display time.

src/master-playlist-controller.js
imbcmdth: I concede.

src/master-playlist-controller.js
```diff
@@ -846,39 +866,48 @@ export class MasterPlaylistController extends videojs.EventTarget {
    * @return {TimeRange} the seekable range
    */
   seekable() {
+    if (!this.seekable_) {
```
mjneil: I want to say I had a reason for doing this, but I can't recall what it would've been. Looking at it now don't really see any harm in setting `this.seekable_ = videojs.createTimeRanges()` at the start. I'll give it a shot and make sure it doesn't break things.

src/playlist.js
```diff
@@ -218,6 +218,94 @@ export const sumDurations = function(playlist, startIndex, endIndex) {
...
 };
+
+/**
+ * Returns an array with two sync points. The first being an expired sync point, which is
+ * the most recent segment with timing sync data that has fallen off the playlist. The
+ * second is a segment sync point, which is the first segment that has timing sync data in
+ * the current playlist.
+ *
+ * @param {Object} playlist a media playlist object
+ * @returns {Array} an array of length two containing two sync points, an expired sync
```
gesinger: An object with names will make it easier to read the result.

mjneil:
Can you clarify odd scenarios?

gesinger:
Nevermind. I was trying to think of cases where a sync point showed what should be segments going over target duration (an invalid playlist, but curious to know the result), but I believe that should be covered.

PR #899, opened 2016-11-08
Seek to live point if we fall off the back of a live playlist

gesinger:
Backported simple-fetcher PR https://github.com/videojs/videojs-contrib-hls/pull/893 to master

### DIFF
--- a/src/always-be-playing.js
+++ b/src/always-be-playing.js
@@ -1,10 +1,10 @@
 /**
- * @file gap-skipper.js
+ * @file always-be-playing.js
  */
 import Ranges from './ranges';
 import videojs from 'video.js';
 
-// Set of events that reset the gap-skipper logic and clear the timeout
+// Set of events that reset the always-be-playing time check logic and clear the timeout
 const timerCancelEvents = [
   'seeking',
   'seeked',
@@ -14,34 +14,26 @@ const timerCancelEvents = [
 ];
 
 /**
- * The gap skipper object handles all scenarios
- * where the player runs into the end of a buffered
- * region and there is a buffered region ahead.
- *
- * It then handles the skipping behavior by setting a
- * timer to the size (in time) of the gap. This gives
- * the hls segment fetcher time to close the gap and
- * resume playing before the timer is triggered and
- * the gap skipper simply seeks over the gap as a
- * last resort to resume playback.
- *
- * @class GapSkipper
+ * @class AlwaysBePlaying
  */
-export default class GapSkipper {
+export default class AlwaysBePlaying {
   /**
-   * Represents a GapSKipper object.
+   * Represents an AlwaysBePlaying object.
    * @constructor
    * @param {object} options an object that includes the tech and settings
    */
   constructor(options) {
     this.tech_ = options.tech;
+    this.seekable = options.seekable;
+    this.playlist = options.playlist;
+
     this.consecutiveUpdates = 0;
     this.lastRecordedTime = null;
     this.timer_ = null;
     this.checkCurrentTimeTimeout_ = null;
 
     if (options.debug) {
-      this.logger_ = videojs.log.bind(videojs, 'gap-skipper ->');
+      this.logger_ = videojs.log.bind(videojs, 'always-be-playing ->');
     }
     this.logger_('initialize');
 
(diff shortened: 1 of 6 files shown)


src/always-be-playing.js
```diff
@@ -135,6 +120,93 @@ export default class GapSkipper {
...
+      return;
+    }
+
+    this.checkGap_(nextRange, currentTime);
+  }
+
+  checkFellOutOfLiveWindow_(playlist, seekable, currentTime) {
+    if (!playlist ||
+        playlist.endList ||
+        !seekable.length ||
```
dmlap: Since seekable.start(0) will always be zero for VOD, I don't think you need to worry about whether the playlist is live or not which would eliminate the need to pass in `playlist` here.

src/always-be-playing.js
```diff
@@ -135,6 +120,93 @@ export default class GapSkipper {
...
+        currentTime >= seekable.start(0)) {
+      return false;
+    }
+
+    let livePoint = seekable.end(seekable.length - 1);
+
+    this.logger_(`Fell out of live window at time ${currentTime}. Seeking to ` +
+                 `live point (seekable end) ${livePoint}`);
+    this.cancelTimer_();
+    this.tech_.setCurrentTime(livePoint);
```
dmlap: Design comment: I like the practice of splitting side-effecting operations from pure ones, all other things being equal. Instead of writing a function that checks-and-seeks-if necessary, you could write a check function and a seek-to-live function.

src/always-be-playing.js
```diff
@@ -135,6 +120,93 @@ export default class GapSkipper {
   }
 
   /**
+   * Handler for `waiting` events from the player
+   *
+   * @private
+   */
+  waiting_() {
```
dmlap: It seems like this method name and jsdoc may be out of date.

test/always-be-playing.test.js
```diff
@@ -132,78 +126,172 @@ QUnit.test('skips over gap in Chrome due to video underflow', function() {
...
   // happened, but we should still allow the player to try to continue from that spot.
   QUnit.deepEqual(
-    this.gapSkipper.gapFromVideoUnderflow_(
+    this.alwaysBePlaying.gapFromVideoUnderflow_(
       videojs.createTimeRanges([[0, 10], [10.1, 12.9]]), 13),
     {start: 10, end: 10.1},
     'returns gap even when current time is not in buffered range');
 });
+
+QUnit.test('check fell out of live window', function() {
```
dmlap: These tests would probably be less complicated if you split the side effects of the check out of the check itself.

src/always-be-playing.js
```diff
@@ -14,34 +14,26 @@ const timerCancelEvents = [
...
- * the hls segment fetcher time to close the gap and
- * resume playing before the timer is triggered and
- * the gap skipper simply seeks over the gap as a
- * last resort to resume playback.
- *
- * @class GapSkipper
+ * @class AlwaysBePlaying
  */
-export default class GapSkipper {
+export default class AlwaysBePlaying {
```
gesinger: Makes sense, and I'm cool with it. Will come up with some alternatives.

src/always-be-playing.js
gesinger: I think the original reason I used side-effects was because video underflow would have to pass back a value in addition to whether or not the condition happened (to prevent having to re-calculate the value). That said, you're still right, and if it doesn't get too messy I can just pass back an object containing all necessary information.

gesinger: Or just use the value itself.

src/always-be-playing.js
gesinger: Nice call. Definitely cleans things up.

src/always-be-playing.js
```diff
@@ -65,11 +57,15 @@ export default class GapSkipper {
   }
 
   /**
-   * Periodically check for timeupdates to see if a gap has been encountered.
+   * Periodically check current time to see if playback stopped
    *
    * @private
    */
   monitorCurrentTime_() {
+    if (!this.tech_.el_) {
```
gesinger: I have to go back and try to re-create it. I was seeing it happen before, but maybe it was something else. If I can't reproduce, I'll remove the check.

dmlap:
LGTM

forbesjo:
LGTM

PR #901, opened 2016-11-09
Handle falling off front of live playlist in playback-watcher

gesinger:
## Description
Note: currently doing more testing with this, but wanted to get this out to review

Handles cases where we fall off the front of a playlist, for instance, when playlist requests are slow.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -122,9 +122,21 @@ export default class PlaybackWatcher {
   waiting_() {
     let seekable = this.seekable();
     let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
 
     if (this.tech_.seeking() || this.timer_ !== null) {
       return;
+    }
+
+    if (this.pastLiveWindow_(seekable, currentTime. buffered)) {
+      let seekableEnd = seekable.end(seekable.length - 1);
+
+      // Through slow playlist requests or some other means, we managed to find ourselves
+      // off the front of the live window. It doesn't matter how we got here, but we need
+      // to get back to our safe seekable range.
+      this.logger_(`Fell past live window at time ${currentTime}. Seeking to ` +
+                   `seekable end ${seekableEnd}`);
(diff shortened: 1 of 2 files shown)


src/playback-watcher.js
```diff
@@ -122,11 +122,23 @@ export default class PlaybackWatcher {
   waiting_() {
     let seekable = this.seekable();
     let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
 
     if (this.tech_.seeking() || this.timer_ !== null) {
       return;
     }
 
+    if (this.pastLiveWindow_(seekable, currentTime. buffered)) {
```
forbesjo: `currentTime, buffered`

forbesjo:
(ignore my approval, I was messing around with the github review feature and clicked the wrong thing)

src/playback-watcher.js
```diff
@@ -122,11 +122,23 @@ export default class PlaybackWatcher {
...
+      let seekableEnd = seekable.end(seekable.length - 1);
+
+      // Through slow playlist requests or some other means, we managed to find ourselves
+      // off the front of the live window. It doesn't matter how we got here, but we need
+      // to get back to our safe seekable range.
+      this.logger_(`Fell past live window at time ${currentTime}. Seeking to ` +
+                   `seekable end ${seekableEnd}`);
+      this.tech_.setCurrentTime(seekableEnd);
```
mjneil: I notice you call `cancelTimer_` in the other two cases that you `setCurrentTime` without setting a timer (fall off live window and video underflow). Should we do that before setting current time here too?

forbesjo: Looks like we're doing the same thing for falling out of the live window, maybe the logic should be combined.

imbcmdth:
We decided not to do this for now as it adds complications to work around a problem with the stream that should exist in the first place.

PR #902, opened 2016-11-10
Immediately reload the source when an error is emitted by the player

imbcmdth:
## Description
When an error is triggered on the player, reload the source and try playback again. There are a large class of errors that MSE can throw which are unrecoverable. This is the fastest way to start playing back again in those cases.

For VOD content, it seeks to the previous `currentTime` value and should resume from exactly where it was left off.

For live content, playback should resume from the usual live-point.

### DIFF
--- a/src/reload-source-on-error.js
+++ b/src/reload-source-on-error.js
@@ -26,7 +26,7 @@ const reloadSourceOnError = function() {
 
   player.on('error', reloadSource);
   player.on('dispose', () => {
-      player.off('error', reloadSource);
+    player.off('error', reloadSource);
   });
 };
 
--- a/src/reload-source-on-error.js
+++ b/src/reload-source-on-error.js
@@ -1,0 +1,33 @@
+/**
+ * Reload the source when an error is detected as long as there
+ * wasn't an error previously within the last 30 seconds
+ */
+const reloadSourceOnError = function() {
+  const player = this;
+  let lastCalled = 0;
+  const reloadSource = function() {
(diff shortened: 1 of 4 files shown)


test/reload-source-on-error.test.js
```diff
@@ -0,0 +1,102 @@
+import QUnit from 'qunit';
+import videojs from 'video.js';
+import reloadSourceOnError from '../src/reload-source-on-error';
+
+reloadSourceOnError
```
dmlap: ?

dmlap:
LGTM

forbesjo:
LGTM

dmlap:
Still looks good.

src/reload-source-on-error.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Reload the source when an error is detected as long as there
+ * wasn't an error previously within the last 30 seconds
+ */
+const reloadSourceOnError = function() {
+  const player = this;
+  let lastCalled = 0;
+  const reloadSource = function() {
+    let tech = player.tech({ IWillNotUseThisInPlugins: true });
+    let sourceObj = tech.currentSource();
```
forbesjo: `currentSource_`

forbesjo:
LGTM

mkhazov:
Thanks for this feature, it's very helpful!
Probably `reloadSourceOnError` plugin need to be mentioned in the library's documentation.

PR #903, opened 2016-11-10
fix trim buffer to compare correct segments

mjneil:
## Description
Buffer was not being trimmed in live playlist since both duration and seekable are calculated in the same way. When we try to make sure we aren't discarding a partial segment, we start at index 0 which gives a previous segment of -1. This change starts at 1 so we compare segment 1 to segment 0 to properly set removeToTime


imbcmdth:
LGTM

PR #907, opened 2016-11-10
Updated the reloadSourceOnError plugin

imbcmdth:
1. Allow the ability to pass a `getSource` function that can be used to provide a new source to load on error
2. Added the ability to override the default minimum time between errors in seconds
3. Plugin now cleans up event bindings when initialized multiple times
4. Better doc-comments!

### DIFF
--- a/src/reload-source-on-error.js
+++ b/src/reload-source-on-error.js
@@ -1,34 +1,74 @@
-/**
- * Reload the source when an error is detected as long as there
- * wasn't an error previously within the last 30 seconds
- */
-const reloadSourceOnError = function() {
-  const player = this;
+import videojs from 'video.js';
+
+const defaultOptions = {
+  errorInterval: 30
+};
+
+const initPlugin = function(player, options) {
   let lastCalled = 0;
-  const reloadSource = function() {
+  let seekTo = 0;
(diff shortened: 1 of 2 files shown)


src/reload-source-on-error.js
```diff
@@ -1,34 +1,121 @@
...
       return;
     }
     lastCalled = Date.now();
 
-    if (seekTo) {
-      player.one('loadedmetadata', () => {
-        player.currentTime(seekTo);
-      });
+    if (localOptions.getSource &&
+        typeof localOptions.getSource === 'function') {
```
dmlap: Alternative: put `getSource` in the default options.

test/reload-source-on-error.test.js
```diff
@@ -103,3 +107,48 @@ QUnit.test('only allows a retry once every 30 seconds', function() {
 
   QUnit.equal(this.player.src.calledWith.length, 2, 'player.src was called twice');
 });
+
+QUnit.test('allows you to override the default retry interval', function() {
```
dmlap: Amazing tests!

dmlap:
Minor comments. LGTM

PR #908, opened 2016-11-11
Don't try to reload if the getSource function returns null or undefined

imbcmdth:
Sometimes `currentSource_` is null which is a valid source. It will clear the player error state though it eventually causes a timeout (because of the invocation of `play`). This change just doesn't try to reload the source at all if the object returned by `getSource` is null or undefined which would have the effect of passing the error (likely a PLAYER_ERR_TIMEOUT) through to display in the error modal.


mjneil:
LGTM

PR #909, opened 2016-11-11
resolve-url.js now uses an iframe to contain the base and anchor elements used to resolve relative urls

imbcmdth:
## Description

Before this fix, we were using a temporary base element (created on-the-fly) to change the relative url resolution of the entire page. We did this for every single url that we needed to resolve (potentially hundreds or thousands per playlist). Some foxy browsers were particularly slow at this when the page contained many anchor tags.

This change sandboxes the base element and anchor into an iframe so that the host-page is unaffected and performance shouldn't suffer as much. In addition, this will only create a single iframe, base, and anchor element for each PlaylistLoader and reuse them for the entire lifetime of the playlist.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -5,7 +5,7 @@
  * M3U8 playlists.
  *
  */
-import resolveUrl from './resolve-url';
+import resolveUrlFactory from './resolve-url';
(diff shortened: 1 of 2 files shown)


mjneil:
LGTM

PR #911, opened 2016-11-14
Fix resolve url function in IEs (by doing what hls.js is doing)

forbesjo:
## Description
The new resolve url function is having problems in IEs, fixing this by doing what hls.js is doing. This needs attribution and fixed tests.


gesinger:
Are there any unit tests we can pull over/modify as well?

forbesjo:
Pulling over the tests

imbcmdth:
LGTM

PR #913, opened 2016-11-15
Use url-toolkit for url manipulation

forbesjo:
## Description
The url resolver in hls.js is now published as url-toolkit. We should start using that and contribute our ideas.

## Requirements Checklist


dmlap:
LGTM

PR #918, opened 2016-11-17
Added support for the throughput of the whole system to the ABR algorithm

imbcmdth:
## Description
Right now the ABR algorithm only takes into account the network bandwidth when making a determination about which variant stream to select. The issue with that is that the ABR algorithm is ignoring the time it takes to decrypt, transmux, and append a segment.

With MSE this isn't too worrying since the bandwidth of that second part of the segment loading process is huge - testing shows that it is in excess of 150 Mb/s. As a result, it's effect on the total system bandwidth is relatively minor.

When using Flash, things are very very different. The reason for this is two-fold: First, Flash is typically only used as a fallback on IE and IE's JavaScript engine doesn't exactly have stellar performance. The second problem is Flash itself.

The mechanism for transferring data into Flash from JavaScript is slow and, worse, it blocks the main browser thread. To get around this problem we chunk up the data and send it piece by piece to Flash. That chunking has the effect of limiting the amount of data we are able to send to Flash per unit of time - creating a hard limit on that component of the system bandwidth.

The equation for determining the aggregate system bandwidth for segment loading is:

```
systemBandwidth = 1 / (1 / networkBandwidth + 1 / appendBandwidth)
```

An example will help to show how this affects variant selection. Let's assume that playback is happening on a computer with ~4Mb/s of network bandwidth.

Starting with the MSE case, we have:

```
1 / (1 / 4Mbps + 1 / 150Mbps) = 3.896Mbps
```

...showing that network bandwidth is a pretty good proxy for system bandwidth when the append bandwidth is so large. Since we apply a scale value to account for network variance we would select a variant around 3.3Mb/s and playback would be fine. 

Now for the Flash side:

```
1 / (1 / 4Mbps + 1 / 5Mbps) = 2.222Mbps
```

...using only the network bandwidth in this situation would result in a pretty poor playback experience. Currently, we'd select the same 3.3Mb/s variant and there is no way that the system could keep up. The results would be stuttering during playback and stalls.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1024,20 +1024,25 @@ export default class SegmentLoader extends videojs.EventTarget {
 
   /**
    * Records the current throughput of the decrypt, transmux, and append
-   * portion of the semgment pipeline. `throughput.rate` is an unweighted
-   * average of the throughput. `throughput.count` is the number of samples
-   * in the average.
+   * portion of the semgment pipeline. `throughput.rate` is a the cumulative
+   * moving average of the throughput. `throughput.count` is the number of
+   * data points in the average.
    *
    * @private
    * @param {Object} segmentInfo the object returned by loadSegment
    */
   recordThroughput_(segmentInfo) {
     let rate = this.throughput.rate;
+    // Add one to the time to ensure that we don't accidentally attempt to divide
+    // by zero in the case where the throughput is ridiculously high
     let segmentProcessingTime =
       1 + ((new Date()).getTime() - segmentInfo.startOfLoad_);
+    // Multiply by 8000 to convert from bytes/millisecond to bits/second
     let segmentProcessingThroughput =
       Math.floor((segmentInfo.byteLength / segmentProcessingTime) * 8 * 1000);
 
+    // This is just a cumulative moving average calculation:
+    //   newAvg = oldAvg + (sample - oldAvg) / (sampleCount + 1)
     this.throughput.rate +=
       (segmentProcessingThroughput - rate) / (++this.throughput.count);
   }
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -755,7 +755,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (request === this.xhr_.segmentXhr) {
       // the segment request is no longer outstanding
       this.xhr_.segmentXhr = null;
-      segmentInfo.startOfLoad_ = (new Date()).getTime();
+      segmentInfo.startOfLoad_ = Date.now();
 
       // calculate the download bandwidth based on segment request
       this.roundTrip = request.roundTripTime;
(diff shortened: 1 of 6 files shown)


imbcmdth:
Requires https://github.com/videojs/videojs-contrib-media-sources/pull/119

src/segment-loader.js
```diff
@@ -754,6 +755,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (request === this.xhr_.segmentXhr) {
       // the segment request is no longer outstanding
       this.xhr_.segmentXhr = null;
+      segmentInfo.startOfLoad_ = (new Date()).getTime();
```
mjneil: You can use `Date.now()` instead of creating a new object

src/videojs-contrib-hls.js
```diff
@@ -399,12 +399,51 @@ class HlsHandler extends Component {
           this.masterPlaylistController_.selectPlaylist = selectPlaylist.bind(this);
         }
       },
+      throughput: {
+        get() {
+          return this.masterPlaylistController_.mainSegmentLoader_.throughput.rate;
+        },
+        set(throughput) {
+          this.masterPlaylistController_.mainSegmentLoader_.throughput.rate = throughput;
+          this.masterPlaylistController_.mainSegmentLoader_.throughput.count = 1;
```
mjneil: Why does a reset of throughput set the count to 1, but reseting the bandwidth below resets the throughput to a count of 0?

src/segment-loader.js
```diff
@@ -754,6 +755,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (request === this.xhr_.segmentXhr) {
       // the segment request is no longer outstanding
       this.xhr_.segmentXhr = null;
+      segmentInfo.startOfLoad_ = Date.now();
```
mjneil: `startOfLoad_` is a bit misleading since this is actually a timestamp post-load

test/videojs-contrib-hls.test.js
```diff
@@ -465,10 +465,65 @@ QUnit.test('downloads media playlists after loading the master', function() {
...
+  });
+  this.player.tech_.hls.bandwidth = 20e10;
+  this.player.tech_.hls.throughput = 20e10;
+  // 1 / ( 1 / 20e10 + 1 / 20e10) = 10e10
+  QUnit.strictEqual(this.player.tech_.hls.systemBandwidth,
+                    10e10,
+                    'systemBandwidth is the combination of bandwidth and throughput');
+});
+
+QUnit.test('setting bandwidth resets throughput', function() {
```
mjneil: This is the same test as line 472

src/segment-loader.js
```diff
@@ -754,6 +755,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (request === this.xhr_.segmentXhr) {
       // the segment request is no longer outstanding
       this.xhr_.segmentXhr = null;
+      segmentInfo.startOfAppend = Date.now();
```
mjneil: in a fragmented mp4 and/or encrypted stream case, we have to wait for all our requests (init segment and/or key), depending on the order we receive the responses, setting the date here will affect the throughput calculation. Not sure if this behavior is ok or not

imbcmdth: Unfortunately bandwidth is only the segment request. So the rest of the process is something I current consider "overhead" so I can capture key response overhead.

dmlap:
LGTM

PR #777, opened 2016-07-13
Upgrade aes-decrypter to use webcrypto for HLSe decryption where available.

dconnolly:
## Description

Use the aes-decrypter@2.0.0 with new api, which uses webcrypto to decrypt AES-CBC encrypted segments where available (Chrome, Firefox, Edge, Opera).
## Specific Changes proposed

Please list the specific changes involved in this pull request.
## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [x] If necessary, more likely in a feature request than a bug fix
  - [x] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))
- [ ] Reviewed by Two Core Contributors

Cast request.response ArrayBuffer to Uint8Array for explicit passing to
decrypt

PR #922, opened 2016-11-23
Revert "Upgrade aes-decrypter to use webcrypto for HLSe decryption where available. (#777)"

imbcmdth:
This reverts commit b9692e29de401a5af50a8bcbc1b04bc2008e1f4f.

We might have problems with the new web-crypto support. So this reverts those changes in contrib-hls while we get aes-decrypter sorted out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "test/"
   ],
   "dependencies": {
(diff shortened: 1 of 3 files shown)


gesinger:
LGTM

PR #927, opened 2016-11-30
Added GOAL_BUFFER_LENGTH_BEFORE_PLAYING option

servercimen:
## Description
Currently, it is not possible to configure goal buffer length for when the player hasn't started played. `1` is hard coded. This pull request moves that value to a config option.

## Specific Changes proposed
Move hard coded value to  config option.

## Requirements Checklist
- [X ] Feature implemented / Bug fixed
  - [ ] Docs/guides updated


src/segment-loader.js
```diff
@@ -433,7 +433,8 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     // if the video has not yet played once, and we already have
     // one segment downloaded do nothing
-    if (!hasPlayed && bufferedTime >= 1) {
+
+    if (!hasPlayed && bufferedTime >= Config.GOAL_BUFFER_LENGTH_BEFORE_PLAYING) {
```
gesinger: We now have a function, `goalBufferLength` in MasterPlaylistController that may be a good spot for this to exist.

forbesjo:
Thank you for your PR but we wont be adding this setting at this time since there doesn't seem to be much demand.

PR #928, opened 2016-12-01
Fix a segment hop in live

gesinger:
## Description
In a case where the live playlist updates during processing of the first segment request, it is possible that we'd skip over the next segment and proceed down the playlist. This accounts for the change in media sequence during update end.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -997,10 +997,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     currentMediaIndex +=
       segmentInfo.playlist.mediaSequence - this.playlist_.mediaSequence;
 
-    if (!segmentInfo.isSyncRequest) {
-      this.mediaIndex = currentMediaIndex;
-      this.fetchAtBuffer_ = true;
-    }
+    this.mediaIndex = currentMediaIndex;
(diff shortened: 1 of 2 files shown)


mjneil:
LGTM

src/segment-loader.js
```diff
@@ -992,16 +992,16 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-      this.mediaIndex = segmentInfo.mediaIndex;
-      this.fetchAtBuffer_ = true;
-    }
-
     let currentMediaIndex = segmentInfo.mediaIndex;
 
     currentMediaIndex +=
       segmentInfo.playlist.mediaSequence - this.playlist_.mediaSequence;
 
+    if (!segmentInfo.isSyncRequest) {
```
mjneil: Can we actually get to this check if the segment is a sync request? In `handleSegment_` we return if it is a sync request and never append, so I would think this check is unnecessary.

imbcmdth: I think Matt means "Can we actually get rid of this check" which I agree with!

mjneil:
LGTM official this time

PR #929, opened 2016-12-02
Representations and Quality Levels

mjneil:
## Description
This PR introduces QualityLevels to contrib-hls.

The first part of this addition is to fix the representations mixin to create closure on the playlist loader instead of the playlist object, since the playlist loader creates a new object reference when loading a new playlist. This would cause outdated references to the representations enabled functions. 

The next change is separating the notion of disabled and blacklisted. We disable playlists for a multitude of reasons. Sometimes we want to blacklist a playlist permanently, for a short time, or to manually have a choice of which playlist is loaded. Since blacklisting and disabling/enabling use the same state variable, there is no way to know the difference between enabling a playlist that was disabled manually or blacklisted permanently.

Lastly, this change brings in videojs-contrib-quality-levels to expose the qualityLevels api on the player and populate the list of quality levels with the available representations in the source. This requires [these changes](https://github.com/videojs/videojs-contrib-quality-levels/pull/4) in contrib-quality-levels.

## Specific Changes proposed
- make representation references persist on new playlist loads
- separate disabling from blacklisting
- bring in the qualityLevels api

## Requirements Checklist
- [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
     "sinon": "1.10.2",
     "uglify-js": "^2.5.0",
     "videojs-standard": "^4.0.3",
+    "videojs-contrib-quality-levels": "^2.0.0",
     "watchify": "^3.6.0",
     "webpack": "^1.13.2"
   }
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "url-toolkit": "^1.0.4",
     "video.js": "^5.15.1",
     "videojs-contrib-media-sources": "^4.1.3",
+    "videojs-contrib-quality-levels": "^1.0.0",
     "videojs-swf": "^5.0.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -96,7 +96,6 @@
     "url-toolkit": "^1.0.4",
     "video.js": "^5.15.1",
     "videojs-contrib-media-sources": "^4.1.3",
-    "videojs-contrib-quality-levels": "^1.0.0",
     "videojs-swf": "^5.0.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "sinon": "1.10.2",
     "uglify-js": "^2.5.0",
     "videojs-standard": "^4.0.3",
-    "videojs-contrib-quality-levels": "^2.0.0",
+    "videojs-contrib-quality-levels": "^2.0.1",
     "watchify": "^3.6.0",
     "webpack": "^1.13.2"
   }
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "sinon": "1.10.2",
     "uglify-js": "^2.5.0",
     "videojs-standard": "^4.0.3",
-    "videojs-contrib-quality-levels": "^2.0.1",
+    "videojs-contrib-quality-levels": "^2.0.2",
     "watchify": "^3.6.0",
     "webpack": "^1.13.2"
   }
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -229,8 +229,10 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
    * @return {Number} number of eneabled playlists
    */
   loader.enabledPlaylists_ = function() {
-    return loader.master.playlists.filter((element, index, array) => {
-      return !element.excludeUntil || element.excludeUntil <= Date.now();
+    return loader.master.playlists.filter((playlist) => {
+      let blacklisted = playlist.excludeUntil && playlist.excludeUntil > Date.now();
+
+      return (!playlist.disabled && !blacklisted);
     }).length;
   };
 
@@ -249,8 +251,8 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     let currentBandwidth = loader.media().attributes.BANDWIDTH || 0;
 
     return !(loader.master.playlists.filter((playlist) => {
-      let enabled = typeof playlist.excludeUntil === 'undefined' ||
-                      playlist.excludeUntil <= Date.now();
+      let blacklisted = playlist.excludeUntil && playlist.excludeUntil > Date.now();
+      let enabled = (!playlist.disabled && !blacklisted);
 
       if (!enabled) {
         return false;
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -7,6 +7,7 @@
  */
 import resolveUrl from './resolve-url';
 import {mergeOptions} from 'video.js';
+import { isEnabled } from './playlist.js';
 import Stream from './stream';
 import m3u8 from 'm3u8-parser';
 import window from 'global/window';
@@ -229,11 +230,7 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
    * @return {Number} number of eneabled playlists
    */
   loader.enabledPlaylists_ = function() {
-    return loader.master.playlists.filter((playlist) => {
-      let blacklisted = playlist.excludeUntil && playlist.excludeUntil > Date.now();
-
-      return (!playlist.disabled && !blacklisted);
-    }).length;
+    return loader.master.playlists.filter(isEnabled).length;
   };
 
   /**
@@ -251,8 +248,7 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     let currentBandwidth = loader.media().attributes.BANDWIDTH || 0;
 
     return !(loader.master.playlists.filter((playlist) => {
-      let blacklisted = playlist.excludeUntil && playlist.excludeUntil > Date.now();
-      let enabled = (!playlist.disabled && !blacklisted);
+      const enabled = isEnabled(playlist);
 
       if (!enabled) {
         return false;
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -92,6 +92,7 @@ let renditionSelectionMixin = function(hlsHandler) {
     return playlists
       .master
       .playlists
+      .filter((media) => !isBlacklisted(media))
       .map((e, i) => new Representation(hlsHandler, e, e.uri));
   };
 };
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -1,3 +1,4 @@
+import { isBlacklisted, isEnabled } from './playlist.js';
 /**
  * Enable/disable playlist function. It is intended to have the first two
  * arguments partially-applied in order to create the final per-playlist
@@ -11,10 +12,10 @@
  * or if undefined returns the current enabled-state for the playlist
  * @return {Boolean} The current enabled-state of the playlist
  */
-let enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
-  let playlist = loader.master.playlists[playlistUri];
-  let blacklisted = playlist.excludeUntil && playlist.excludeUntil > Date.now();
-  let currentlyEnabled = (!playlist.disabled && !blacklisted);
+const enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
+  const playlist = loader.master.playlists[playlistUri];
+  const blacklisted = isBlacklisted(playlist);
+  const currentlyEnabled = isEnabled(playlist);
 
   if (typeof enable === 'undefined') {
     return currentlyEnabled;
--- a/src/rendition-mixin.js
+++ b/src/rendition-mixin.js
@@ -91,7 +91,7 @@ let renditionSelectionMixin = function(hlsHandler) {
     return playlists
       .master
       .playlists
-      .map((e, i) => new Representation(hlsHandler, e, i));
+      .map((e, i) => new Representation(hlsHandler, e, e.uri));
   };
 };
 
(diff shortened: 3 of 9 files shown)


src/videojs-contrib-hls.js
```diff
@@ -78,6 +79,57 @@ const safeGetComputedStyle = function(el, property) {
...
+ * @param {PlaylistLoader} playlistLoader PlaylistLoader containing the new media info.
+ * @function handleHlsMediaChange
+ */
+const handleHlsMediaChange = function(qualityLevels, playlistLoader) {
+  let newPlaylist = playlistLoader.media();
+  let selectedIndex = getHlsPlaylistIndex(newPlaylist, playlistLoader.master.playlists);
+
+  qualityLevels.selectedIndex_ = selectedIndex;
+  qualityLevels.trigger({
+    selectedIndex,
```
gesinger: Is this guaranteed to be the same index as the index in QualityLevelList?

src/videojs-contrib-hls.js
```diff
@@ -524,6 +578,26 @@ class HlsHandler extends Component {
   }
 
   /**
+   * Initializes the quality levels and sets listeners to update them.
+   *
+   * @method setupQualityLevels_
+   * @private
+   */
+  setupQualityLevels_() {
+    let _player = videojs(this.tech_.options_.playerId);
```
mjneil: not really, was just following suit of already existing code https://github.com/videojs/videojs-contrib-hls/blob/master/src/videojs-contrib-hls.js#L287

src/videojs-contrib-hls.js
mjneil: If the list is changes (e.g. a quality level is removed with `removeQualityLevel`, then it wont be the same. We do set the id of the quality level to the same as the index of its representation. With the QualityLevelList api is setup, i could guarantee this in a round about way of finding the qualitylevel by id, then get the index of that level in the underlying list. Or the quality level api could be updated for the getQualityLevelById to instead return an index.

test/rendition-mixin.test.js
```diff
@@ -174,19 +192,20 @@ QUnit.test('setting a representation to disabled sets excludeUntil to Infinity',
 
   renditions[0].enabled(false);
 
-  assert.equal(playlists[0].excludeUntil, Infinity, 'rendition has an infinite excludeUntil');
-  assert.equal(playlists[1].excludeUntil, 0, 'rendition has an excludeUntil of zero');
+  assert.equal(playlists[0].disabled, true, 'rendition has been disabled');
+  assert.equal(playlists[1].disabled, undefined, 'rendition has not been disabled');
```
mjneil: added (not sure why github didnt pick this up as outdated along with the other comments)

src/videojs-contrib-hls.js
```diff
@@ -106,11 +86,17 @@ const getHlsPlaylistIndex = function(media, playlists) {
  */
 const handleHlsMediaChange = function(qualityLevels, playlistLoader) {
   let newPlaylist = playlistLoader.media();
-  let selectedIndex = getHlsPlaylistIndex(newPlaylist, playlistLoader.master.playlists);
+  let i;
```
gesinger: Not a big deal, but I generally like to keep the index vars within the scope of the loop when possible, as outside of it the context is often lost, even when the code is close together. Maybe we can use selectedIndex as the var?

src/videojs-contrib-hls.js
```diff
@@ -106,11 +86,17 @@ const getHlsPlaylistIndex = function(media, playlists) {
...
 
-  qualityLevels.selectedIndex_ = selectedIndex;
+  for (i = qualityLevels.length - 1; i >= 0; i--) {
```
gesinger: Any specific reason for going down rather than up?

mjneil: Was just trying to avoid an extra check. If no quality level is found with the correct id, i will become -1 at the end of the loop, which is what we want to set selectedIndex if there isnt a selected QualityLevel, but along with your comment above, it definitely makes things less clear and its nothing extra to restructure this in a more clear way

src/videojs-contrib-hls.js
```diff
@@ -78,6 +78,45 @@ const safeGetComputedStyle = function(el, property) {
...
+
+  for (let i = 0; i < qualityLevels.length; i++) {
+    if (qualityLevels[i].id === newPlaylist.uri) {
+      selectedIndex = i;
+      break;
+    }
+  }
+
+  qualityLevels.selectedIndex_ = selectedIndex;
+  qualityLevels.trigger({
```
gesinger: Do we want to trigger if -1?

mjneil: Yes, although selectedIndex shouldn't ever be -1 unless a user manually changes the quality level list, if listening to the change event, this will allow them to see the invalid state and correct or catch any errors instead of this changing out from under them

src/playlist-loader.js
```diff
@@ -249,8 +251,8 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     let currentBandwidth = loader.media().attributes.BANDWIDTH || 0;
 
     return !(loader.master.playlists.filter((playlist) => {
-      let enabled = typeof playlist.excludeUntil === 'undefined' ||
-                      playlist.excludeUntil <= Date.now();
```
imbcmdth: These two lines are the same thing as the lines 233-235. I think making these two lines a function that takes a playlist parameter would be better (something like `isPlaylistEnabled_(playlist) {...}`. Then the `enabledPlaylists_` function becomes `return loader.master.playlists.filter(isPlaylistEnabled_).length` and these two lines become just something like `let enabled = isPlaylistEnabled_(playlist);`

src/rendition-mixin.js
```diff
@@ -11,21 +11,22 @@
  * or if undefined returns the current enabled-state for the playlist
  * @return {Boolean} The current enabled-state of the playlist
  */
-let enableFunction = (playlist, changePlaylistFn, enable) => {
-  let currentlyEnabled = typeof playlist.excludeUntil === 'undefined' ||
-                         playlist.excludeUntil <= Date.now();
+let enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
+  let playlist = loader.master.playlists[playlistUri];
+  let blacklisted = playlist.excludeUntil && playlist.excludeUntil > Date.now();
+  let currentlyEnabled = (!playlist.disabled && !blacklisted);
```
imbcmdth: Another perfect place for a hypothetical `isPlaylistEnabled_` function. 👍

src/playlist-loader.js
mjneil: Good call, I'm thinking an `isEnabled(playlist)` exported from playlist.js since we have other playlist utility functions there

src/rendition-mixin.js
mjneil: The function is not as useful here since we still need to get the intermediate information of whether the playlist is blacklisted or not. In the other places we only care about if the playlist is enabled or not, but here we want to make sure that the playlist isn't blacklisted even if `enable` differs from `currentlyEnabled` before calling the `changePlaylistFn`

src/videojs-contrib-hls.js
```diff
@@ -103,10 +142,9 @@ Hls.STANDARD_PLAYLIST_SELECTOR = function() {
   // filter out any playlists that have been excluded due to
   // incompatible configurations or playback errors
   sortedPlaylists = sortedPlaylists.filter((localVariant) => {
-    if (typeof localVariant.excludeUntil !== 'undefined') {
-      return now >= localVariant.excludeUntil;
-    }
-    return true;
+    let blacklisted = localVariant.excludeUntil && localVariant.excludeUntil > now;
+
+    return (!localVariant.disabled && !blacklisted);
```
imbcmdth: You know what would be really great here!?

package.json
```diff
@@ -96,6 +96,7 @@
     "url-toolkit": "^1.0.4",
     "video.js": "^5.15.1",
     "videojs-contrib-media-sources": "^4.1.3",
+    "videojs-contrib-quality-levels": "^1.0.0",
```
forbesjo: This should be removed

mjneil: Good catch, slipped through the rebase

src/rendition-mixin.js
```diff
@@ -87,7 +92,8 @@ let renditionSelectionMixin = function(hlsHandler) {
     return playlists
       .master
       .playlists
-      .map((e, i) => new Representation(hlsHandler, e, i));
+      .filter((media) => !isBlacklisted(media))
```
imbcmdth: I thought we wouldn't filter blacklisted media here?

mjneil: So I added this to filter out media that was blacklisted in `excludeIncompatibleVariants_` so that we don't add quality levels for a playlist that will never be selected. It's a good point that since every time representations is called, it rebuilds its list with this filter, so temporary blacklisting will get removed here. With respect to quality levels this isn't an issue since we only call representations once and then work with the quality level list. Could filter on `media.excludeUntil !== Infinity` instead to remove the permanent blacklisted media but still allow temporary blacklists through.

gesinger: This may get overly complicated, but would it be better for us to use something else to identify incompatible variants versus temporary suspensions? It may make more contextual sense, but if it ends up being too many options, the check for not Infinity could be an OK solution for now.

src/videojs-contrib-hls.js
```diff
@@ -78,6 +78,45 @@ const safeGetComputedStyle = function(el, property) {
 };
 
 /**
+ * Updates the selcetedIndex of the QualityLevelList when a mediachange happens in hls.
```
gesinger: selectedIndex

src/videojs-contrib-hls.js
gesinger: Makes sense 👍

gesinger: Will this trigger a change event even if a playlist refreshes but doesn't change?

test/rendition-mixin.test.js
```diff
@@ -135,35 +149,39 @@ QUnit.test('returns representations with width and height if present', function(
   assert.equal(renditions[2].height, undefined, 'rendition has a height of undefined');
 });
 
-QUnit.test('representations are disabled if their excludeUntil is after Date.now', function(assert) {
+QUnit.test('blacklisted playlists are not included in the representations list', function(assert) {
   let hlsHandler = makeMockHlsHandler([
     {
       bandwidth: 0,
-      excludeUntil: Infinity
+      excludeUntil: Infinity,
```
gesinger: We should add another with excludeUntil between 0 and Infinity.

test/rendition-mixin.test.js
```diff
@@ -135,35 +149,39 @@ QUnit.test('returns representations with width and height if present', function(
...
     }
   ]);
 
   RenditionMixin(hlsHandler);
 
   let renditions = hlsHandler.representations();
 
-  assert.equal(renditions[0].enabled(), false, 'rendition is not enabled');
-  assert.equal(renditions[1].enabled(), true, 'rendition is enabled');
+  assert.equal(renditions.length, 1, 'rblacklisted rendition not added');
```
gesinger: blacklisted

test/rendition-mixin.test.js
gesinger: 👍

src/rendition-mixin.js
```diff
@@ -11,21 +12,22 @@
  * or if undefined returns the current enabled-state for the playlist
  * @return {Boolean} The current enabled-state of the playlist
  */
-let enableFunction = (playlist, changePlaylistFn, enable) => {
-  let currentlyEnabled = typeof playlist.excludeUntil === 'undefined' ||
-                         playlist.excludeUntil <= Date.now();
+const enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
+  const playlist = loader.master.playlists[playlistUri];
```
mjneil: Yes, the original function used just the playlist but this PR needed to change that since the call to `bind` below would reference an outdated playlist object since `PlaylistLoader` overwrites the reference when it gets a new playlist. This ensures we are acting on the correct object.

src/videojs-contrib-hls.js
mjneil: Nope, we fire a `loadedplaylist` event every time we refresh a playlist, but only fire `mediachange` when the playlist uri changes. This function is only called on those `mediachange` events.

src/rendition-mixin.js
mjneil: I think the current decision is that the current solution is OK for now, and we can revisit it later.

mjneil:
RE: Also would probably be worth adding tests for enableFunction

There are already tests to ensure that calling `representation.enable()` calls the `fastQualityChange_` function and tests to make sure the `fastQualityChange_` function is functioning properly,  so I don't think more tests need to be added

test/rendition-mixin.test.js
```diff
@@ -135,35 +149,44 @@ QUnit.test('returns representations with width and height if present', function(
...
+    },
+    {
+      bandwidth: 0,
+      excludeUntil: 0,
+      uri: 'media1.m3u8'
     },
     {
       bandwidth: 0,
-      excludeUntil: 0
+      excludeUntil: Date.now() + 999999,
```
gesinger: Minor, but might be useful to have one with expired excludeUntil

mjneil: That is actually covered with the above playlist `excludeUntil: 0`

gesinger: That is true, and maybe in this case it would be overdoing it to add another one, but usually it is good to have a non 0 test, as 0 is falsy and often is an edge case.

gesinger:
LGTM

PR #939, opened 2016-12-21
Processing segment reachable even after playlist update removes it

gesinger:
## Description
In live scenarios, a refresh may render a processing segment unreachable. This attaches the segment to the segmentInfo object so the segment will always be around.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -526,7 +526,9 @@ export default class SegmentLoader extends videojs.EventTarget {
       // The timeline that the segment is in
       timeline: segment.timeline,
       // The expected duration of the segment in seconds
-      duration: segment.duration
+      duration: segment.duration,
+      // retain the segment in case the playlist updates while doing an async process
+      segment
     };
   }
 
@@ -674,7 +676,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.sourceUpdater_.remove(0, removeToTime);
     }
 
-    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex];
+    segment = segmentInfo.segment;
 
     // optionally, request the decryption key
     if (segment.key) {
@@ -756,7 +758,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
 
     segmentInfo = this.pendingSegment_;
-    segment = segmentInfo.playlist.segments[segmentInfo.mediaIndex];
+    segment = segmentInfo.segment;
 
     // if a request times out, reset bandwidth tracking
     if (request.timedout) {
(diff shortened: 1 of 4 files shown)


imbcmdth:
When a playlist is refreshed, the segments in the new playlist are copies of the old segments. That means that if playlist is refreshed between segment request and xhr completing, we will annotate the old segment object at: 

https://github.com/gesinger/videojs-contrib-hls/blob/b426cc22fe27de3d96a53d5586c0ab3a451d3ad1/src/sync-controller.js#L310-L314

gesinger:
Good call @imbcmdth . Here are a couple approaches (I think (1) is better, but there are definitely alternatives from these as well):

1) When we update segment loader's playlist with a refresh and change the media index of segmentInfo, we also change the segment reference. We'd have to check that the segment exists on the new playlist (media index did not turn negative). If it doesn't exist, we simply leave the old segment reference (it'll save info onto a segment that is missing from the playlist, but no harm will be done). This is the simplest change from the current code.

2) Instead of saving the segment reference on segmentInfo, we continue to use mediaIndex and the playlist reference but save old segments that are not officially part of the playlist anymore for the case where they fell off. This may lead to misleading playlist objects, as we will save information that technically no longer exists in the playlist.

test/segment-loader.test.js
```diff
@@ -1014,6 +1014,153 @@ function(assert) {
...
+  assert.equal(loader.pendingSegment_.uri, '0.ts', 'first segment pending');
+
+  // wrap up the first request to set mediaIndex and start normal live streaming
+  this.requests[0].response = new Uint8Array(10).buffer;
+  this.requests.shift().respond(200, null, '');
+  mediaSource.sourceBuffers[0].buffered = videojs.createTimeRanges([[0, 10]]);
+  mediaSource.sourceBuffers[0].trigger('updateend');
+
+  assert.equal(loader.state, 'WAITING', 'in waiting state');
+  assert.equal(loader.pendingSegment_.uri, '1.ts', 'second segment pending');
```
mjneil: I think it would be good to also check `loader.pendingSegment_.segment.uri` is what you expect whenver checking `loader.pendingSegment_.uri` to to make sure the new segment object we are attaching to `segmentInfo` is the correct reference

src/segment-loader.js
```diff
@@ -279,6 +279,13 @@ export default class SegmentLoader extends videojs.EventTarget {
 
         if (segmentInfo && !segmentInfo.isSyncRequest) {
```
mjneil: Why do we only do these adjustments for non-sync requests. It probably doesn't matter if we don't save the info on an outdated segment object when its a sync-request because we don't append it and if we do need to append it in the future, we are going to have to re-request/probe/sync it anyway. However, I think it probably doesn't hurt to make sure the timing info we get from the probe of the sync request to be put onto the segment object. It might help make other calculations that fallback to target duration when start and end times are unavailable be more accurate

imbcmdth:
LFGTM

PR #940, opened 2016-12-22
Map legacy AVC codecs to their modern equivalents when excluding inco…

gesinger:
…mpatible playlists

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -79,8 +79,10 @@ const parseCodecs = function(codecs) {
  *
  * @param codecString {String} the codec string
  * @return {String} the codec string with old apple-style codecs replaced
+ *
+ * @private
  */
-export const mapLegacyAvcCodecs = function(codecString) {
+export const mapLegacyAvcCodecs_ = function(codecString) {
   return codecString.replace(/avc1\.(\d+)\.(\d+)/i, (match) => {
     return translateLegacyCodecs([match])[0];
   });
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -7,6 +7,7 @@ import Ranges from './ranges';
 import videojs from 'video.js';
 import AdCueTags from './ad-cue-tags';
 import SyncController from './sync-controller';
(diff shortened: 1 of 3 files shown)


src/master-playlist-controller.js
```diff
@@ -72,6 +73,12 @@ const parseCodecs = function(codecs) {
   return result;
 };
 
+export const mapLegacyAvcCodecs = (codecString) => {
```
dmlap: jsdoc comment?

mjneil: This doesn't really affect anything since this function doesn't use `this`, but I think it's safer to not use arrow function syntax for top level functions just because it creates special behavior

dmlap: Why export this if it's only used in this module?

gesinger: Exported for unit tests

mjneil: exported for use in unit tests

gesinger:
Updated

src/master-playlist-controller.js
```diff
@@ -73,6 +74,19 @@ const parseCodecs = function(codecs) {
 };
 
 /**
+ * Replace codecs in the codec string with the old apple-style `avc1.<dd>.<dd>` to the
+ * standard `avc1.<hhhhhh>`.
+ *
+ * @param codecString {String} the codec string
+ * @return {String} the codec string with old apple-style codecs replaced
+ */
+export const mapLegacyAvcCodecs = function(codecString) {
```
mjneil: Can we add an underscore and @private to this since we don't want this to be a part of the api that we are exposing

src/master-playlist-controller.js
```diff
@@ -7,6 +7,7 @@ import Ranges from './ranges';
 import videojs from 'video.js';
 import AdCueTags from './ad-cue-tags';
 import SyncController from './sync-controller';
+import { translateLegacyCodecs } from 'videojs-contrib-media-sources/es5/codec-utils';
```
mjneil: I think we should plan to not do a deep import in the future, but it is ok for now

mjneil:
LGTM

PR #959, opened 2017-01-13
Fix any possible fillBuffer_ race conditions by debouncing all fillBuffers_

imbcmdth:
## Description
On a rendition switch, SegmentLoader seems to be able to enter a state where it fetches the same segment twice at the same time causing bandwidth starvation. The requests then timeout and an emergency rendition switch which can trigger the same behavior *again*.

### What's gone wrong?

The root problem is that `handleUpdateEnd_` can trigger a rendition switch (by emitting the 'progress' event). Eventually, a rendition switch results in `load` being called on the SegmentLoader which has the side effect of calling `fillBuffer_` and starting a request for a segment. At the same time, the original `handleUpdateEnd_` function calls `fillBuffer_` itself triggering a second request.

### Why doesn't this happen every time we change renditions?

The issue is "masked" by two different mechanisms each of which can stop the second request from happening:
1. If the rendition we are switching to has never been loaded before, the PlaylistLoader first must make a request for the renditions playlist/manifest. That introduces asynchronicity that results in the SegmentLoader being in a state (waiting) so that `fillBuffer_` is never called by `SegmentLoader#load`.
2. If there is buffered data more than 30 seconds before the playhead, the first call to `fillBuffer_` will, via `loadSegment_`, result in a `SourceUpdater.remove` that causes the `SourceBuffer#updating` to be `true`. When the next `fillBuffer_` runs, it will exit early because the sourceBuffer is updating.

## Specific Changes proposed
* Convert all calls to `fillBuffer_` into calls to `monitorBuffer_`
* Rename `monitorBuffer_` to `monitorBufferTick_` which becomes the 500ms buffer check timer-loop
* Make a new `monitorBuffer_` that schedule an immediate (1 ms) timer for `monitorBufferTick_`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -203,7 +203,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // next segment
     if (!this.paused()) {
       this.state = 'READY';
-      this.fillBuffer_();
+      this.monitorBuffer_();
     }
   }
 
@@ -252,7 +252,6 @@ export default class SegmentLoader extends videojs.EventTarget {
     }
 
     this.state = 'READY';
-    this.fillBuffer_();
   }
 
   /**
@@ -349,12 +348,25 @@ export default class SegmentLoader extends videojs.EventTarget {
   }
 
   /**
+   * (re-)schedule monitorBufferTick_ to run as soon as possible
+   *
+   * @private
+   */
+  monitorBuffer_() {
+    if (this.checkBufferTimeout_) {
+      window.clearTimeout(this.checkBufferTimeout_);
+    }
+
+    this.checkBufferTimeout_ = window.setTimeout(this.monitorBufferTick_.bind(this), 1);
+  }
+
+  /**
    * As long as the SegmentLoader is in the READY state, periodically
    * invoke fillBuffer_().
    *
    * @private
    */
-  monitorBuffer_() {
+  monitorBufferTick_() {
     if (this.state === 'READY') {
       this.fillBuffer_();
     }
@@ -363,7 +375,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       window.clearTimeout(this.checkBufferTimeout_);
     }
 
-    this.checkBufferTimeout_ = window.setTimeout(this.monitorBuffer_.bind(this),
+    this.checkBufferTimeout_ = window.setTimeout(this.monitorBufferTick_.bind(this),
                                                  CHECK_BUFFER_DELAY);
   }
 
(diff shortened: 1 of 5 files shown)


src/segment-loader.js
```diff
@@ -349,12 +348,25 @@ export default class SegmentLoader extends videojs.EventTarget {
   }
 
   /**
+   * (re-)schedule monitorBufferTick_ to run as soon as possible
+   *
+   * @private
+   */
+  monitorBuffer_() {
```
gesinger: Why do we need this in addition to monitorBufferTick_? It seems that by changing calls from fillBuffer_ to monitorBuffer_, we already get the guard that checks whether the state is READY.

fillBuffer_, if it starts a segment request, will synchronously get to the WAITING state. This means that any subsequent calls to monitorBuffer would be safe with the READY guard.

imbcmdth: The real difference is in how long in the future things are scheduled. Calling `monitorBuffer_` forces the tick to happen next (it either schedules or reschedules the timer tick to be just one millisecond). Whereas `monitorBufferTick_` always schedules the next time to be 500ms in the future.

Originally, I had a piece of code with an optional parameter to `monitorBuffer_` to set/reset different timer durations in one function but I decided that was "too clever".

gesinger: As a followup, spoke with @imbcmdth , and he pointed out that monitorBuffer_ contextually is async, and is safer  as async when there are events that are being listened to that may happen around a segment's end of processing (as an example, if monitorBuffer_ was synchronous, by the time listeners received the events, a new segment may have already been requested and a lot of state changed).

test/segment-loader.test.js
```diff
@@ -350,23 +365,26 @@ QUnit.test('adjusts the playlist offset if no buffering progress is made', funct
...
   sourceBuffer.trigger('updateend');
+  this.clock.tick(1);
 
   // so the loader should try the next segment
   assert.equal(this.requests[0].url, '1.ts', 'moved ahead a segment');
 
   // verify stats
   assert.equal(loader.mediaBytesTransferred, 20, '20 bytes');
-  assert.equal(loader.mediaTransferDuration, 2, '2 ms (clocks above)');
+  assert.equal(loader.mediaTransferDuration, 1, '1 ms (clocks above)');
```
gesinger: Why did this decrease?

imbcmdth: Prior to this change, the request happened immediately. As a result, there were two clock ticks between the request and the response. Now, the request happens later and only one tick happens now (the first tick becomes the tick that triggers the request).

forbesjo:
LGTM

PR #972, opened 2017-01-23
run decryption in a webworker

mjneil:
## Description
Currently trying to play an encrypted source in a background tab causes the player to crash. This seems to be due to browsers throttling setTimeouts to 1000ms delays while in a background tab. Since our decryption uses a lot of timeouts, it cannot preform fast enough in a background tab. This change moves the decryption work into a web worker since timeouts are not throttled while in a worker.

## Specific Changes proposed
* Move decryption to a web worker for the segment loaders to use

## Requirements Checklist

### DIFF
--- a/src/bin-utils.js
+++ b/src/bin-utils.js
@@ -31,6 +31,37 @@ const formatAsciiString = function(e) {
 };
 
 /**
+ * Creates an object for sending to a web worker modifying properties that are TypedArrays
+ * into seperated properties for the buffer, byteOffset, and byteLength.
+ *
+ * @param {Object} message
+ *        Object of properties and values to send to the web worker
+ * @return {Object}
+ *         Modified message with TypedArray values expanded
+ * @function transferableMessage
+ */
+const transferableMessage = function(message) {
+  const transferable = {};
+  const keys = Object.keys(message);
+
+  keys.forEach((key) => {
+    const value = message[key];
+
+    if (ArrayBuffer.isView(value)) {
+      transferable[key] = {
+        bytes: value.buffer,
+        byteOffset: value.byteOffset,
+        byteLength: value.byteLength
+      };
+    } else {
+      transferable[key] = value;
+    }
+  });
+
+  return transferable;
+};
+
+/**
  * utils to help dump binary data to the console
  */
 const utils = {
@@ -59,7 +90,8 @@ const utils = {
       result += textRange(ranges, i) + ' ';
     }
     return result;
-  }
+  },
+  transferableMessage
 };
 
 export default utils;
--- a/src/bin-utils.js
+++ b/src/bin-utils.js
@@ -32,19 +32,18 @@ const formatAsciiString = function(e) {
 
 /**
  * Creates an object for sending to a web worker modifying properties that are TypedArrays
- * into seperated properties for the buffer, byteOffset, and byteLength.
+ * into a new object with seperated properties for the buffer, byteOffset, and byteLength.
  *
  * @param {Object} message
  *        Object of properties and values to send to the web worker
  * @return {Object}
  *         Modified message with TypedArray values expanded
- * @function transferableMessage
+ * @function createTransferableMessage
  */
-const transferableMessage = function(message) {
+const createTransferableMessage = function(message) {
   const transferable = {};
-  const keys = Object.keys(message);
 
-  keys.forEach((key) => {
+  Object.keys(message).forEach((key) => {
     const value = message[key];
 
     if (ArrayBuffer.isView(value)) {
@@ -91,7 +90,7 @@ const utils = {
     }
     return result;
   },
-  transferableMessage
+  createTransferableMessage
 };
 
 export default utils;
--- a/src/decrypter-worker.js
+++ b/src/decrypter-worker.js
@@ -1,0 +1,58 @@
+import window from 'global/window';
+import {Decrypter} from 'aes-decrypter';
+
+/**
+ * Callback method to send the decrypted segment bytes back to the main thread
+ *
+ * @param {Object} err
+ *        Decryption error. Will always be null with aes-decrypter ^1.0.3
+ * @param {Uint8Array} bytes
+ *        Decrypted byte array
+ * @function postDecrypted
+ */
+const postDecrypted = function(err, bytes) {
+  window.postMessage({
+    action: 'done',
+    source: this.source,
+    bytes: bytes.buffer,
+    byteOffset: bytes.byteOffset,
+    byteLength: bytes.byteLength
+  }, [bytes.buffer]);
+};
+
+/**
+ * Our web wroker interface so that things can talk to aes-decrypter
+ * that will be running in a web worker. the scope is passed to this by
+ * webworkify.
+ *
+ * @param {Object} self
+ *        the scope for the web worker
+ */
+const Worker = function(self) {
+  self.onmessage = function(event) {
+    const data = event.data;
+    if (data.action === 'decrypt') {
+      const encrypted = new Uint8Array(data.encrypted.bytes,
+                                     data.encrypted.byteOffset,
+                                     data.encrypted.byteLength);
+      const key = new Uint32Array(data.key.bytes,
+                                data.key.byteOffset,
+                                data.key.byteLength / 4);
+      const iv = new Uint32Array(data.iv.bytes,
+                               data.iv.byteOffset,
+                               data.iv.byteLength / 4);
+      const context = {
+        source: data.source
+      };
+
+      new Decrypter(encrypted,
+                    key,
+                    iv,
+                    postDecrypted.bind(context));
+    }
+  };
+};
+
+export default (self) => {
+  return new Worker(self);
+};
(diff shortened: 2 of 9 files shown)


src/decrypter-worker.js
```diff
@@ -0,0 +1,47 @@
+import window from 'global/window';
+import {Decrypter} from 'aes-decrypter';
+
+/**
+ * Our web wroker interface so that things can talk to aes-decrypter
```
ESPNMichaelKidd: typo wroker -> worker

imbcmdth:
Can you update the README to point out that this change means we no longer support IE10.

src/decrypter-worker.js
```diff
@@ -0,0 +1,47 @@
...
+ * webworkify.
+ *
+ * @param {Object} self
+ *        the scope for the web worker
+ */
+const Worker = function(self) {
+  self.onmessage = function(event) {
+    const data = event.data;
+
+    if (data.action === 'decrypt') {
```
gesinger: Is the action specifier necessary since we're only decrypting?

src/master-playlist-controller.js
```diff
@@ -256,7 +262,22 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // combined audio/video or just video when alternate audio track is selected
     this.mainSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
     // alternate audio track
+    segmentLoaderOptions.loaderType = 'audio';
     this.audioSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
+
+    this.decrypter_.onmessage = (event) => {
+      switch (event.data.source) {
+      case 'main':
```
gesinger: Style preference, but maybe indent the cases.

test/videojs-contrib-hls.test.js
```diff
@@ -1466,6 +1468,18 @@ QUnit.test('the source handler supports HLS mime types', function(assert) {
...
+QUnit.test('source handler does not support sources when IE 10 or below', function(assert) {
+  videojs.browser.IE_VERSION = 10;
+
+  ['html5', 'flash'].forEach(function(techName) {
+    assert.ok(!HlsSourceHandler(techName).canHandleSource({
+      type: 'aPplicatiOn/x-MPegUrl'
+    }), 'does not support when browser is IE10');
+  });
+
+  videojs.browser.IE_VERSION = this.old.IE_VERSION;
```
gesinger: If we're resetting in afterEach, do we need to do it here?

test/videojs-contrib-hls.test.js
```diff
@@ -1466,6 +1468,18 @@ QUnit.test('the source handler supports HLS mime types', function(assert) {
   });
 });
 
+QUnit.test('source handler does not support sources when IE 10 or below', function(assert) {
+  videojs.browser.IE_VERSION = 10;
+
+  ['html5', 'flash'].forEach(function(techName) {
+    assert.ok(!HlsSourceHandler(techName).canHandleSource({
+      type: 'aPplicatiOn/x-MPegUrl'
```
gesinger: I know it shouldn't matter, but since this test is specific to IE 10 testing, may want to keep this is proper casing

src/master-playlist-controller.js
```diff
@@ -235,6 +237,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.syncController_ = new SyncController();
 
+    this.decrypter_ = work(Decrypter);
```
gesinger: Why do we create one decrypter here to share among loaders instead of creating a decrypter in each loader?

src/segment-loader.js
```diff
@@ -923,22 +925,53 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-                      this.handleSegment_();
-                    }).bind(this));
-      /* eslint-enable */
+      this.decrypter_.postMessage({
+        action: 'decrypt',
+        source: this.loaderType_,
+        encrypted: {
+          bytes: segmentInfo.encryptedBytes.buffer,
+          byteOffset: segmentInfo.encryptedBytes.byteOffset,
+          byteLength: segmentInfo.encryptedBytes.byteLength
```
imbcmdth: I think it might be time for a helper function to make these objects. Something so that we can do:

```
  ...
  encrypted: transferTypedArray(segmentInfo.encryptedBytes),
  key: transferTypedArray(segmentInfo.key.bytes),
  iv: transferTypedArray(segmentInfo.key.iv)
},
...
```

src/master-playlist-controller.js
```diff
@@ -8,6 +8,8 @@ import videojs from 'video.js';
 import AdCueTags from './ad-cue-tags';
 import SyncController from './sync-controller';
 import { translateLegacyCodecs } from 'videojs-contrib-media-sources/es5/codec-utils';
+import work from 'webworkify';
```
imbcmdth: I think `work` is a bit generic and `worker` would be a more descriptive name for the function.

src/master-playlist-controller.js
mjneil: vjsstandard linting rules require the cases to be the same indentation of the switch. I too prefer the cases indented

src/master-playlist-controller.js
mjneil: This is to limit the number of workers we are creating. We currently already have one worker for transmuxing. Creating the decrypter in each loader brings that number up to 3, but there really isnt any reason we can use the same worker since decryption doesn't need state. Also, since we always create an audio segment loader, but the common use case is muxed video and audio source, the audio loader is often not used, but would still create a wasted worker

src/master-playlist-controller.js
```diff
@@ -256,7 +262,22 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // combined audio/video or just video when alternate audio track is selected
     this.mainSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
     // alternate audio track
+    segmentLoaderOptions.loaderType = 'audio';
     this.audioSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
+
+    this.decrypter_.onmessage = (event) => {
+      switch (event.data.source) {
```
imbcmdth: Matt and I discussed turning this into a set of if/else instead of an *evil* switch statement!! Just noting that incase someone else notices the switch statement and feels ill.

src/bin-utils.js
```diff
@@ -31,6 +31,37 @@ const formatAsciiString = function(e) {
 };
 
 /**
+ * Creates an object for sending to a web worker modifying properties that are TypedArrays
+ * into seperated properties for the buffer, byteOffset, and byteLength.
```
gesinger: maybe "...into a new object with separated properties..."

src/bin-utils.js
```diff
@@ -31,6 +31,37 @@ const formatAsciiString = function(e) {
...
+ *
+ * @param {Object} message
+ *        Object of properties and values to send to the web worker
+ * @return {Object}
+ *         Modified message with TypedArray values expanded
+ * @function transferableMessage
+ */
+const transferableMessage = function(message) {
+  const transferable = {};
+  const keys = Object.keys(message);
```
gesinger: Only used once. Probably simpler to just do `Object.keys(message).forEach(...`

src/bin-utils.js
```diff
@@ -31,6 +31,37 @@ const formatAsciiString = function(e) {
...
+ * Creates an object for sending to a web worker modifying properties that are TypedArrays
+ * into seperated properties for the buffer, byteOffset, and byteLength.
+ *
+ * @param {Object} message
+ *        Object of properties and values to send to the web worker
+ * @return {Object}
+ *         Modified message with TypedArray values expanded
+ * @function transferableMessage
+ */
+const transferableMessage = function(message) {
```
gesinger: Really minor, and up for debate (not sure which is actually better), but maybe it would make sense to call this function expandTypedArrays.

src/master-playlist-controller.js
```diff
@@ -256,7 +262,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // combined audio/video or just video when alternate audio track is selected
     this.mainSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
     // alternate audio track
+    segmentLoaderOptions.loaderType = 'audio';
     this.audioSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
+
+    this.decrypter_.onmessage = (event) => {
+      if (event.data.source === 'main') {
+        this.mainSegmentLoader_.handleDecrypted_(event.data);
+      } else {
```
gesinger: Just preference, but it might be good to keep this as `else if (event.data.source === 'audio')`

src/bin-utils.js
imbcmdth: `Transferrable` is the name given to the class of objects that can be zero-copied into a webworker. ArrayBuffers are one of those types but to be able to reconstruct them we need some more data. This function prepares them for transferrability. Personally, I think the name mostly makes sense but I think it might be better to call it `createTransferableMessage`.

imbcmdth:
Looking good to me.

PR #975, opened 2017-01-24
fix: introduce videojs 6 forward compatibility while maintaining backward compatibilty

### LINKED ISSUE #968: Compatibility with videojs 6.0 RC
## Description
Can't play m3u8 files in video.js 6.0.0-RC.0 but it still play in video.js 5.15.1
The video played and showed "No compatible source was found for this media."

## Steps to reproduce
1. Import media from d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8 and other m3u8 files
2. Press play but it show "No compatible source was found for this media."

### Error output
* Press play but it show "No compatible source was found for this media."

### videojs-contrib-hls version
* videojs-contrib-hls 4.1.1

### videojs version
* video.js 6.0.0-RC.0

### Browsers
* Google Chrome version 55.0.2883.87 m (64-bit)

### Platforms
* Windows 10 Pro version 1607 64bit

gkatsev:
This makes it backwards and forwards compatible to videojs 6 and as far back as [5.2](https://github.com/videojs/video.js/blob/master/CHANGELOG.md#520-2015-11-10) when the tech registry was added.

Fix #968

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -212,14 +212,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.cueTagsTrack_ = this.tech_.addTextTrack('metadata',
         'ad-cues');
       this.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
-
-      const textTracks = this.tech_.textTracks();
-
-      if (textTracks.addTrack) {
-        textTracks.addTrack(this.cueTagsTrack_);
-      } else {
-        textTracks.addTrack_(this.cueTagsTrack_);
-      }
     }
 
     this.audioTracks_ = [];
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -212,7 +212,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.cueTagsTrack_ = this.tech_.addTextTrack('metadata',
         'ad-cues');
       this.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
-      this.tech_.textTracks().addTrack_(this.cueTagsTrack_);
+
+      const textTracks = this.tech_.textTracks();
+
+      if (textTracks.addTrack) {
+        textTracks.addTrack(this.cueTagsTrack_);
+      } else {
+        textTracks.addTrack_(this.cueTagsTrack_);
+      }
     }
 
     this.audioTracks_ = [];
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -781,8 +781,10 @@ if (window.Uint8Array && flashTech) {
 videojs.HlsHandler = HlsHandler;
 videojs.HlsSourceHandler = HlsSourceHandler;
 videojs.Hls = Hls;
+if (!videojs.use) {
+  videojs.registerComponent('Hls', Hls);
+}
 videojs.m3u8 = m3u8;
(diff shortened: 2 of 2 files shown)


gkatsev:
Looks like this doesn't fix compat for vjs 6 since `Hls` a regular object is registered as a component which is currently disallowed.

BrandonOCasey:
I'm not sure that registerComponent call is actually needed

mjneil:
If users expect to be able to access hls with `videojs.getComponent('Hls')` then we would lose backwards compatibility with removing the register.

gkatsev:
A potential workaround is to not register `Hls` as a component if we're in a videojs 6 env. For example:
```js
if (!videojs.use) {
  videojs.registerComponent('Hls');
}
```
This keeps compat for 5.x and sets us to remove it from hls when 5.x support is dropped.

gkatsev:
Also, for better or worse, videojs 6.0 no longer checks for components on the `videojs` object, it won't error out/warn in the way that prompted using it in the first place.

gkatsev:
I think this PR could go in without answering the `registerComponent` question.

gkatsev:
And for `plugin`, we could either leave it or we can change it to be something like:
```js
if (videojs.registerPlugin) {
  videojs.registerPlugin(...);
} else {
  videojs.plugin(...);
}
```

gkatsev:
So, I updated the PR:
* use `getTech` instead of `getComponent`
* verify that the flash tech is included before attaching something to it
* if `registerPlugin` is available, use it instead
* if `videojs.use` is available, *don't* register "Hls" as a component
  this isn't too bad since `videojs.Hls` is still around. Plus, that one won't cause any issues since the code that checks for components on the videojs object is removed in videojs 6. Plus, given that this is a videojs project, I don't think it's too bad for it to be attaching itself as `videojs.Hls`.

BrandonOCasey:
@gkatsev you will also have to change `addTrack_` to `addTrack` on vjs6 https://github.com/videojs/videojs-contrib-hls/blob/master/src/master-playlist-controller.js#L217

gkatsev:
It seems to be working just fine for me? Maybe the version of vjs I was testing with is a bit outdated, hm...

BrandonOCasey:
Also a lot of the unit tests are failing on vjs6,

gkatsev:
Saw an issue with evented which should be fixed here: https://github.com/videojs/video.js/pull/3990

gkatsev:
Oh, I had to set the `useCueTags` option. Updated that line.

gkatsev:
I'm looking into the test failures with vjs 6.0. However, those are going to go into a separate PR

mjneil:
LGTM

src/master-playlist-controller.js
```diff
@@ -212,7 +212,6 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.cueTagsTrack_ = this.tech_.addTextTrack('metadata',
         'ad-cues');
       this.cueTagsTrack_.inBandMetadataTrackDispatchType = '';
-      this.tech_.textTracks().addTrack_(this.cueTagsTrack_);
```
BrandonOCasey: Shouldn't this have just been changed to `this.tech_.textTracks().addTrack(this.cueTagsTrack_);` rather than removed?

gkatsev: The track is added automatically inside of `addTextTrack`.

PR #976, opened 2017-01-24
Add a segment-metadata TextTrack that contains cues for the segments currently in the buffer

mjneil:
## Description
Currently, we do not expose or track segment information for the segments that are currently in the actual buffer. Being able to know which segment or rendition is currently being rendered on screen has been a much requested feature. This addition allows users to listen to `cuechange` events on the track and see the segment information for the currently rendered segment in the active cues list.

## Specific Changes proposed
* Adds a new TextTrack to the tech with the label `segment-metadata`
* The Cues in the track directly correspond to the current state of the buffer at a segment level
* As segments are appended to the buffer, a new cue is added to the track that has start and end times equal to the segment. If the segment being appended overlaps already buffered content, we remove any overlapping cues from the track so that the cues in the track are always the most up to date.
* whenever data is removed from the buffer, we remove any cues that also fall inside the removed area

## Requirements Checklist
- [ ] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,16 +451,16 @@ the web](http://www.html5rocks.com/en/tutorials/track/basics/).
 ### Segment Metadata
 You can get metadata about the segments currently in the buffer by using the `segment-metadata`
 text track. You can get the metadata of the currently rendered segment by looking at the
-tracks `activeCues` array. The metadata will be attached to the `cue.value` property and
+track's `activeCues` array. The metadata will be attached to the `cue.value` property and
 will have this structure
 
 ```javascript
 cue.value = {
-  uri: The Segment uri,
-  timeline: Timeline of the segment,
-  playlist: The Playlist uri,
-  start: Segment start time,
-  end: Segment end time
+  uri, // The Segment uri
+  timeline, // Timeline of the segment for detecting discontinuities
+  playlist, // The Playlist uri
+  start, // Segment start time
+  end // Segment end time
 };
 ```
 
--- a/README.md
+++ b/README.md
@@ -448,6 +448,50 @@ cue.value.data
 There are lots of guides and references to using text tracks [around
 the web](http://www.html5rocks.com/en/tutorials/track/basics/).
 
+### Segment Metadata
+You can get metadata about the segments currently in the buffer by using the `segment-metadata`
+text track. You can get the metadata of the currently rendered segment by looking at the
+tracks `activeCues` array. The metadata will be attached to the `cue.value` property and
+will have this structure
+
+```javascript
+cue.value = {
+  uri: The Segment uri,
+  timeline: Timeline of the segment,
+  playlist: The Playlist uri,
+  start: Segment start time,
+  end: Segment end time
+};
+```
+
+Example:
+Detect when a change in quality is rendered on screen
+```javascript
+let tracks = player.textTracks();
+let segmentMetadataTrack;
+
+for (let i = 0; i < tracks.length; i++) {
+  if (tracks[i].label === 'segment-metadata') {
+    segmentMetadataTrack = tracks[i];
+  }
+}
+
+let previousPlaylist;
+
+if (segmentMetadataTrack) {
+  segmentMetadataTrack.on('cuechange', function() {
+    let activeCue = segmentMetadataTrack.activeCues[0];
+
+    if (activeCue) {
+      if (previousPlaylist !== activeCue.playlist) {
+        console.log('Switched from rendition' + previousPlaylist + 'to rendition' + activeCue.playlist);
+      }
+      previousPlaylist = activeCue.playlist;
+    }
+  });
+}
+```
+
 ## Hosting Considerations
 Unlike a native HLS implementation, the HLS tech has to comply with
 the browser's security policies. That means that all the files that
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -252,9 +252,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       hasPlayed: () => this.hasPlayed_(),
       bandwidth,
       syncController: this.syncController_,
-      decrypter: this.decrypter_,
-      loaderType: 'main',
-      segmentMetadataTrack: this.segmentMetadataTrack_
+      decrypter: this.decrypter_
     };
 
     // setup playlist loaders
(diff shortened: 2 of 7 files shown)


src/master-playlist-controller.js
```diff
@@ -255,6 +260,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // setup segment loaders
     // combined audio/video or just video when alternate audio track is selected
     this.mainSegmentLoader_ = new SegmentLoader(segmentLoaderOptions);
+
+    // for now we don't want the audio segment loader to also add cues to the segment
+    // metadata track
+    delete segmentLoaderOptions.segmentMetadataTrack;
```
imbcmdth: I think it would be better to create two different objects.

Perhaps using `Object.create` to make the `mainSegmentLoader_`'s object with `segmentMetadataTrack` added to the base object.

src/segment-loader.js
```diff
@@ -1104,4 +1109,42 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+
+    const value = {
+      uri: segmentInfo.uri,
+      timeline: segmentInfo.timeline,
+      playlist: playlist.uri,
+      start,
+      end
+    };
+    const Cue = window.WebKitDataCue || window.VTTCue;
+    const data = JSON.stringify(value);
```
dmlap: It might be more useful to just merge these properties onto the created cue object.

src/segment-loader.js
```diff
@@ -1104,4 +1109,42 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+   * Adds a cue to the segment-metadata track with some metadata information about the
+   * segment
+   *
+   * @private
+   * @param {Object} segmentInfo
+   *        the object returned by loadSegment
+   * @method addSegmentMetadataCue_
+   */
+  addSegmentMetadataCue_(segmentInfo) {
+    if (!this.segmentMetadataTrack_ || !segmentInfo) {
```
gesinger: I don't think we have to be guarding against these.

src/segment-loader.js
```diff
@@ -1104,4 +1109,42 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      playlist: playlist.uri,
+      start,
+      end
+    };
+    const Cue = window.WebKitDataCue || window.VTTCue;
+    const data = JSON.stringify(value);
+    const cue = new Cue(start, end, data);
+
+    value.data = data;
+    cue.value = value;
```
gesinger: So cue ends up with a stringified `value` for its text, then has a value property which has the `value` object, including the stringified `value`?

test/segment-loader.test.js
```diff
@@ -782,6 +809,82 @@ QUnit.test('tracks segment end times as they are buffered', function(assert) {
...
+    };
+
+    assert.equal(track.cues.length, 3, 'one cue added for segment');
+    assert.deepEqual(JSON.parse(track.cues[2].value.data), expectedCue,
+      'added correct segment info to cue');
+
+    // verify stats
+    assert.equal(loader.mediaBytesTransferred, 30, '10 bytes');
+    assert.equal(loader.mediaRequests, 3, '1 request');
+  });
```
gesinger: Would also be worth testing the behavior with overlapping times from a different timeline.

src/segment-loader.js
mjneil: You're right with not needing a check for `segmentInfo` but i think for check for the track is still needed since both the main and audio segment loaders are instances of this class, but the audio segment loader doesnt have a segment metadata track, so we just want this function to be a noop in that case

gesinger: Good call on checking for `segmentMetadataTrack_`

src/segment-loader.js
mjneil: Yes and no. It mainly depends on the `const Cue = window.WebKitDataCue || window.VTTCue;` line. I was following what is done in contrib-media-sources when adding id3 tags, but `WebKitDataCue` is only in safari right now, and there are some small differences between `WebKitDataCue` and `VTTCue`. Mainly, `WebKitDataCue` does not have a `text` property but instead a `value`. For id3 tags this `value` property has a `key` and `data` property usually. My goal with these lines was to minimize the differences between the two cues so that you can just use `cue.value` in all the browsers. After having to explain this, it may just be better to only use `VTTCue` and not jump through all these hoops

test/segment-loader.test.js
mjneil: Adding a test for overlapping segments is definitely a good call. I don't think being on a different timeline makes a different and needs to be tested. At the point of creating the cues, the start and end times have been converted to display time, so the non-congruent timestamps across discontinuities are not a factor

gesinger: 👍

src/segment-loader.js
gesinger: I think keeping Safari support is a good idea (just in case). Maybe a comment and/or some kind of helper method would help.

mjneil: Well `VTTCue` is supported in safari, its just that `WebKitDataCue` is only supported in safari. Is it best practice to use these types of cues for certain purposes, or are they more general purpose and its not "wrong" to just use the `VTTCue` for safari as well?

mjneil: Deciding to stick with what we do for id3 metadata tags since these cues are another form of metadata

src/segment-loader.js
mjneil: do you mean if say you wanted the uri, accessing it should look like `cue.uri` instead of `cue.value.uri`?

README.md
```diff
@@ -458,6 +458,50 @@ cue.value.data
...
+### Segment Metadata
+You can get metadata about the segments currently in the buffer by using the `segment-metadata`
+text track. You can get the metadata of the currently rendered segment by looking at the
+tracks `activeCues` array. The metadata will be attached to the `cue.value` property and
+will have this structure
+
+```javascript
+cue.value = {
+  uri: The Segment uri,
+  timeline: Timeline of the segment,
```
gesinger: Might be useful to say what the timeline is

src/segment-loader.js
```diff
@@ -1104,4 +1109,43 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+   * @param {Object} segmentInfo
+   *        the object returned by loadSegment
+   * @method addSegmentMetadataCue_
+   */
+  addSegmentMetadataCue_(segmentInfo) {
+    if (!this.segmentMetadataTrack_) {
+      return;
+    }
+
+    const playlist = segmentInfo.playlist;
```
gesinger: Can get rid of this and just use `segmentInfo.playlist.uri` below

gesinger:
Tested, but with the same code I can't get it working on IE11/Windows 7 (not even getting `cuechange` events from the `segmentMetadataTrack`), whereas it works even for Flash (and HTML) on Chrome/Mac.

forbesjo:
Looks good to me

PR #977, opened 2017-01-24
Fixed mediaIndex tracking so that it is consistent when the playlist updates during a live stream

imbcmdth:
## What
During low-bandwidth conditions, segment requests can take long enough that the playlist reloads while we are fetching segments. This confused our handling of `mediaIndex` forcing the player to load the same segment that it loaded previously.

If network conditions are bad enough (and we are on the lowest rendition) then we can get into a situation where we refresh the playlist during each segment's request. The end result is that we will fetch the same segment over and over again without making progress. In extreme cases (when two playlist loads happened while fetching a segment) we can actually go backwards!!

## Why
There was code in `SegmentLoader#handleUpdateEnd_` that would adjust `segmentInfo.mediaIndex` before setting `this.mediaIndex`. The change was based on the difference between the saved playlist object and the current playlist's `mediaSequence`.

This action by `handleUpdateEnd_` was ONLY required when `this.mediaIndex` was `null` but we didn't have a check. When `this.mediaIndex` was not `null`, `SegmentLoader#playlist` was responsible for adjusting `mediaIndex`. 

For this PR, I decided to remove the code in `handleUpdateEnd_` and move all the responsibility of tracking `mediaIndex` to one place - `SegmentLoader#playlist`.

## Changes
1. Removed any code in `SegmentLoader#handleUpdateEnd_` that modified the `mediaIndex`
2. Reordered `SegmentLoader#playlist` to make it easier to follow and less deeply nested
3. All changes to both `mediaIndex`-es (SegmentLoader's and segmentInfo's) now happen solely in `SegmentLoader#playlist`

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -270,8 +270,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.playlist_ = newPlaylist;
     this.xhrOptions_ = options;
 
-    // when we haven't started playing yet, the start of the playlist is always
-    // our zero-time so force a sync update each time we get a new playlist
+    // when we haven't started playing yet, the start of a live playlist
+    // is always our zero-time so force a sync update each time the playlist
+    // is refreshed from the server
     if (!this.hasPlayed_()) {
       newPlaylist.syncInfo = {
         mediaSequence: newPlaylist.mediaSequence,
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -267,32 +267,12 @@ export default class SegmentLoader extends videojs.EventTarget {
     let oldPlaylist = this.playlist_;
     let segmentInfo = this.pendingSegment_;
 
-    if (this.mediaIndex !== null) {
-      // We reloaded the same playlist so we are in a live scenario
-      // and we will likely need to adjust the mediaIndex
-      if (oldPlaylist &&
-          oldPlaylist.uri === newPlaylist.uri) {
-        let mediaSequenceDiff = newPlaylist.mediaSequence - oldPlaylist.mediaSequence;
+    this.playlist_ = newPlaylist;
+    this.xhrOptions_ = options;
 
-        this.mediaIndex -= mediaSequenceDiff;
-
-        if (segmentInfo) {
-          segmentInfo.mediaIndex -= mediaSequenceDiff;
-
-          // we need to update the referenced segment so that timing information is
-          // saved for the new playlist's segment, however, if the segment fell off the
-          // playlist, we can leave the old reference and just lose the timing info
-          if (segmentInfo.mediaIndex >= 0) {
-            segmentInfo.segment = newPlaylist.segments[segmentInfo.mediaIndex];
-          }
-        }
-
-        this.syncController_.saveExpiredSegmentInfo(oldPlaylist, newPlaylist);
-      } else {
-        // We must "resync" the fetcher when we switch renditions
-        this.resyncLoader();
-      }
-    } else if (!this.hasPlayed_()) {
+    // when we haven't started playing yet, the start of the playlist is always
+    // our zero-time so force a sync update each time we get a new playlist
+    if (!this.hasPlayed_()) {
       newPlaylist.syncInfo = {
         mediaSequence: newPlaylist.mediaSequence,
         time: 0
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -267,47 +267,64 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // and we will likely need to adjust the mediaIndex
+    let mediaSequenceDiff = newPlaylist.mediaSequence - oldPlaylist.mediaSequence;
+
+    log('mediaSequenceDiff', mediaSequenceDiff);
+
+    // update the mediaIndex on the SegmentLoader
+    // this is important because we can abort a request and this value must be
+    // equal to the last appended mediaIndex
+    if (this.mediaIndex !== null) {
+      this.mediaIndex -= mediaSequenceDiff;
```
gesinger: In the case where this results in a mediaIndex < -1 (we fell two slots behind the end of the playlist), do we want to handle the case in any way (since we knowingly fell off the back), or let playback watcher catch that we fell off the back in its own time (or another resolution, e.g., we keep the index at -1 and hope it'll create a 1 segment gap that will be skipped over by the gap skipper)?

src/segment-loader.js
```diff
@@ -267,47 +267,64 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    this.xhrOptions_ = options;
 
-        this.syncController_.saveExpiredSegmentInfo(oldPlaylist, newPlaylist);
-      } else {
-        // We must "resync" the fetcher when we switch renditions
-        this.resyncLoader();
-      }
-    } else if (!this.hasPlayed_()) {
+    // when we haven't started playing yet, the start of the playlist is always
+    // our zero-time so force a sync update each time we get a new playlist
```
gesinger: "each time we get a new playlist" may be a bit confusing in this context

src/segment-loader.js
imbcmdth: I think it is ok to let playback watcher handle this scenario instead of introducing complexity here. What do you think?

gesinger: Yeah, it's probably fine to do that. The behavior at least will be the same as current.

But it may be an interesting idea for the future. Since we're in a place where we knowingly will fall back to playback-watcher, and the user will experience a pause in playback, maybe we should at least add a comment or TODO to consider taking action.

mjneil:
It would be nice to have a unit test that emulates the scenario that caused the issue.

test/segment-loader.test.js
```diff
@@ -606,6 +606,104 @@ QUnit.test('abort does not cancel segment processing in progress', function(asse
...
+QUnit.test('segmentInfo.mediaIndex is adjusted when live playlist is updated', function(assert) {
+  currentTime = 31;
+  loader.playlist(playlistWithDuration(50, {
+    mediaSequence: 0,
+    endList: false
+  }));
+  loader.mimeType(this.mimeType);
+  loader.load();
+  // Start at mediaIndex 2 which means that the next segment we request
+  // should mediaIndex 3
```
mjneil: These comments are not describing what is happening

test/segment-loader.test.js
```diff
@@ -606,6 +606,104 @@ QUnit.test('abort does not cancel segment processing in progress', function(asse
...
+  assert.equal(this.requests[0].url, '3.ts', 'requesting the segment at mediaIndex 3');
+
+  // Update the playlist shifting the mediaSequence by 2 which will result
+  // in a decrement of the mediaIndex by 2 to 1
+  loader.playlist(playlistWithDuration(50, {
+    mediaSequence: 2,
+    endList: false
+  }));
+
+  assert.equal(segmentInfo.mediaIndex, 1, 'segmentInfo.mediaIndex is updated to 1');
```
mjneil: Might be a good idea to also check the `segmentInfo.segment` reference was also updated

test/segment-loader.test.js
```diff
@@ -606,6 +606,104 @@ QUnit.test('abort does not cancel segment processing in progress', function(asse
...
+  this.requests[0].response = new Uint8Array(10).buffer;
+  this.requests.shift().respond(200, null, '');
+  this.clock.tick(1);
+  mediaSource.sourceBuffers[0].trigger('updateend');
+
+  assert.equal(loader.mediaIndex, 2, 'SegmentLoader.mediaIndex ends at 2');
+});
+
+QUnit.test('segmentInfo.mediaIndex is adjusted when live playlist is updated', function(assert) {
+  currentTime = 31;
```
mjneil: A short comment here explaining that you're setting currentTime to 31 so the loader selects segment `3.ts` would be helpful. It took me a while to figure out why `segmentInfo.mediaIndex` starts at 3 down below when `loader.mediaIndex = null` was being set

test/segment-loader.test.js
mjneil: This is already covered in a different test, ignore comment

mjneil:
LGTM

PR #978, opened 2017-01-25
Bubble progress events

dmlap:
If the segment request triggers progress events (that is, XHR2 is supported), bubble those up to the tech. This makes it clearer that buffering is happening even on very slow connections.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -707,6 +707,9 @@ export default class SegmentLoader extends videojs.EventTarget {
     });
 
     segmentXhr = this.hls_.xhr(segmentRequestOptions, this.handleResponse_.bind(this));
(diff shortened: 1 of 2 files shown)


imbcmdth:
LGTM

gkatsev:
We should figure out how to use the event object in the progress listener to have better bandwidth reporting. Particularly around aborted segments. The [progress event](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent) has a `loaded` and `total` attributes which contain the currently loaded amount and the total for what we're downloading.

This definitely should go in another PR, though, but, I thought it would be good to mention here.

dmlap:
@gkatsev yeah, that would be cool. Good to note for the future.

PR #980, opened 2017-01-25
Error early for misconfigured overrideNative

dmlap:
overrideNative requires emulated text tracks and silently fails if it's misconfigured. Instead, fail spectacularly and early and document that behavior.

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -329,6 +329,14 @@ class HlsHandler extends Component {
       }
     }
 
+    // overriding native HLS only works if audio tracks have been emulated
+    // error early if we're misconfigured:
+    if (videojs.options.hls.overrideNative &&
+        (tech.featuresNativeVideoTracks || tech.featuresNativeAudioTracks)) {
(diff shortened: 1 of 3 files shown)


mjneil:
LGTM

dmlap:
@mjneil I verified in Edge-- it's exploding as designed.

src/videojs-contrib-hls.js
```diff
@@ -329,6 +329,14 @@ class HlsHandler extends Component {
       }
     }
 
+    // overriding native HLS only works if audio tracks have been emulated
+    // error early if we're misconfigured:
+    if (videojs.options.hls.overrideNative &&
+        (tech.featuresNativeVideoTracks || tech.featuresNativeAudioTracks)) {
+      throw new Error('Overriding native HLS requires emulated tracks. ' +
+                      'See https://git.io/vMpjB');
```
gkatsev: this link to master without a specific revision. If the README changes the url will no longer be valid.

dmlap: True. I didn't want to have to update this with every release. I hope it doesn't end up being an issue but we'll see.

Issue #665, opened 2016-04-22
hls.xhr.beforeRequest seems do not work properly


md2k:
Hi Guys, i tried release 2.1.1 (over npm) and found that `beforeRequest` work correctly only:
- if I set it globally as videojs.Hls.xhr.beforeRequest
- if I set it during playback with `player.tech({ IWillNotUseThisInPlugins: true }).hls.xhr.beforeRequest`

But if i trying set it as:

```
player.hls.xhr.beforeRequest = function(options) {
  options.uri = options.uri.replace('example.com', 'foo.com');

  return options;
};
```

I see nothing.
My app loading video programmatically, application do next steps to play:

```
function _play(vObj, data) {

   // some magic about vObj - data about stream with source and some other params
   // _api is videojs api
  _api.src({ type: "application/x-mpegurl", src: url, withCredentials: true});
  _api.hls.xhr.beforeRequest = function(options) {
      console.log(options);
      return options;
   };
   api.load();
   api.play();
}
```

Maybe i do something wrong, but as i said above , was succeed only with Globally defined function or already when playback starts, assigning function to player before playback didnt call this functionality.

sdcoca:
Have similar issue. And see two additional unexpected behaviours:

- If videojs.Hls.xhr.beforeRequest is changed, then it is never called since a copy of it seems to be used instead of the actual. Thus, if not updated before it is copied (which I cannot manage to do) then it is ignored.
- When the player switches sources, the player.hls.xhr hook gets removed as if it was a new player instance.

whatvn:
with newest version I no longer can override xhr. beforeRequest by writing my own videojs.Hls.xhr. beforeRequest function. 

The only way to do this is override that function at runtime : `player.hls.xhr.beforeRequest`. 
And as described in contribute hls doc, we also can override hls.xhr, although it's not recommended, but it's possible. With previous version of contrib-hls, i can override xhr by writing my own videojs.Hls.xhr, but now it's not possible too. 
Any idea?

apatinoyu:
Hello. Di you fix it? I am trying to set it as a global variable videojs.Hls.xhr.beforeRequest = function (options) {
//here I add my headers
}
but no success

kazazes:
I'm also experiencing this regression. My `videojs.hls.xhr` overridden function used to be called, now it isn't.

kazazes:
[...]

PR #983, opened 2017-01-26
If beforeRequest is set, reuse it on source changes

gesinger:
## Description
handleSource would replace any preexisting beforeRequest. 
Reported in https://github.com/videojs/videojs-contrib-hls/issues/665
Example of bug: http://jsbin.com/qijehan/2/edit?html,console,output

## Specific Changes proposed
Now handleSource checks for a preexisting beforeRequest, and reuses it if it is there.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [X] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -691,7 +691,7 @@ const HlsSourceHandler = function(mode) {
 
       let previousBeforeRequest;
 
-      if (tech.hls && tech.hls.xhr && tech.hls.xhr.beforeRequest) {
+      if (tech.hls && tech.hls.xhr) {
         previousBeforeRequest = tech.hls.xhr.beforeRequest;
       }
 
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -703,10 +703,6 @@ const HlsSourceHandler = function(mode) {
         // The player had a beforeRequest set prior to the source change.
         // Use it for the new source.
         tech.hls.xhr.beforeRequest = previousBeforeRequest;
-      } else if (videojs.Hls.xhr.beforeRequest) {
-        // Use a global `before` function if specified on videojs.Hls.xhr
-        // but still allow for a per-player override
-        tech.hls.xhr.beforeRequest = videojs.Hls.xhr.beforeRequest;
       }
 
       tech.hls.src(source.src);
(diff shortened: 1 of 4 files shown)


src/videojs-contrib-hls.js
```diff
@@ -674,12 +674,23 @@ const HlsSourceHandler = function(mode) {
 
       let settings = videojs.mergeOptions(options, {hls: {mode}});
 
+      let previousBeforeRequest;
+
+      if (tech.hls && tech.hls.xhr && tech.hls.xhr.beforeRequest) {
```
mjneil: You don't need to check for `tech.hls.xhr.beforeRequest` here. Even if `beforeRequest` is `undefined`, the next line won't cause an  exception.

src/videojs-contrib-hls.js
```diff
@@ -674,12 +674,23 @@ const HlsSourceHandler = function(mode) {
...
+        previousBeforeRequest = tech.hls.xhr.beforeRequest;
```
mjneil: Could change this to `previousBeforeRequest = tech.hls.xhr.beforeRequest || videojs.Hls.xhr.beforeRequest;` This could remove the need for the if statements below to instead just always set `tech.hls.xhr.beforeRequest = previousBeforeRequest;`. However, the code in its current state may be more understandable, so I'd only consider this comment if you think its an improvement there

mjneil:
Many of the people in the issue you referenced mention that using the global override `videojs.Hls.xhr.beforeRequest` isn't working and the only way to override the function is at runtime. Is that still an issue and does this address that?

gesinger:
Good call @mjneil , changed it so that they can be changed anytime.

src/videojs-contrib-hls.js
```diff
@@ -674,13 +674,20 @@ const HlsSourceHandler = function(mode) {
 
       let settings = videojs.mergeOptions(options, {hls: {mode}});
 
+      let previousBeforeRequest;
+
+      if (tech.hls && tech.hls.xhr) {
```
imbcmdth: Not sure how I feel about this. We should be cleaning up `tech.hls` on dispose anyway. That we don't is actually a bug.

gesinger: On dispose is fine, but what about on source changes? Since the `hls` object is per player, shouldn't it continue to exist and not be disposed?

imbcmdth: No, the source handler is disposed of on each source change.

imbcmdth: In fact, the HLS property on the player is for compatibility reasons. David and I were considering removing the object all together if HLS is used with Video.js 6+

PR #984, opened 2017-01-27
WIP: Vjs 6 tests

gkatsev:


gkatsev:
This fixes ~90% of the tests with videojs 6. The stuff that are left are unusual tests that have expectations that have changed, like logs and flash.
For whatever reason, tests aren't being generated correctly if `videojs-flash` is being imported via a linked dependency. Might be solved with a real release. Currently, most of them are now wrapped in a null check.

src/videojs-contrib-hls.js
```diff
@@ -666,6 +666,7 @@ const HlsSourceHandler = function(mode) {
 
       let settings = videojs.mergeOptions(options, {hls: {mode}});
 
+      console.log('foo');
```
BrandonOCasey: the foo is strong with this one :)

gkatsev: heh, oops

gkatsev:
Needs to be rebased against master. Also, the checking for whether `Flash` exists can be removed. Otherwise, all tests pass locally when run against vjs 6.0.0-RC.1.

BrandonOCasey:
this PR should also include videojs-flash as a dev dependency so that flash tests can always run

gesinger:
Thanks @gkatsev . Closing in favor of https://github.com/videojs/videojs-contrib-hls/pull/1151

gkatsev:
Re-opening this PR because #1151 is done. I'll rebase it against master and fix issues.

test/videojs-contrib-hls.test.js
```diff
@@ -1751,44 +1922,55 @@ QUnit.test('fires loadstart manually if Flash is used', function(assert) {
...
   assert.throws(function() {
     player.src({
       src: 'http://example.com/manifest/master.m3u8',
       type: 'application/x-mpegURL'
     });
+    this.clock.tick(1);
   }, 'errors if native tracks are enabled');
   player.dispose();
 
   player = createPlayer({html5: {hls: {overrideNative: true}}});
```
gesinger: Should this have the clock passed?

test/videojs-contrib-hls.test.js
```diff
@@ -1751,44 +1922,55 @@ QUnit.test('fires loadstart manually if Flash is used', function(assert) {
...
   assert.ok(!player.tech_.hls, 'did not load hls tech');
   player.dispose();
+  Html5.canPlaySource = oldHtml5CanPlaySource;
 });
 
 QUnit.test('loads if native HLS is available and override is set locally', function(assert) {
   let player;
 
   Hls.supportsNativeHls = true;
   player = createPlayer({html5: {hls: {overrideNative: true}}});
```
gesinger: Should this have the clock passed?

test/videojs-contrib-hls.test.js
```diff
@@ -107,6 +107,7 @@ QUnit.module('HLS', {
 
     // setup a player
     this.player = createPlayer();
```
gesinger: Should this have the clock passed?

gkatsev:
To answer @gesinger's question, there's no need to pass a clock into `createPlayer` in all cases. It's only necessary if `createPlayer` is with a particular source/tech and so we need to tick the clock before we call `mockTech` or else it blows up.

.travis.yml
```diff
@@ -11,12 +11,25 @@ notifications:
     channels:
       - "chat.freenode.net#videojs"
     use_notice: true
-# Set up a virtual screen for Chrome.
 before_script:
+  # check if the current version is equal to the version for the env
+  - 'export IS_INSTALLED="$(npm list video.js | grep "video.js@$VJS")"'
```
forbesjo: This will pick up `video.js@6` in `videojs-flash` and never install `video.js@6` at the top level. Perhaps instead use `npm list video.js --depth 0`

gkatsev: Maybe we should just always run `npm install "video.js@$VJS"`?

forbesjo: Sounds like the simplest course of action

forbesjo:
When I run the tests with vjs 6 many fail saying IS_FIREFOX is read only and cannot be reassigned

package.json
```diff
@@ -49,7 +49,7 @@
     "watch:js:switcher": "watchify utils/switcher/switcher.js -v -t babelify -g browserify-shim -o dist/switcher.js",
     "watch:test:browserify": "watchify test/browserify-test.js -o dist-test/browserify-test.js",
     "watch:test:webpack": "watchify test/webpack-test.js -o dist-test/webpack-test.js",
-    "watch:test": "npm-run-all -p watch:test:js watch:test:webpack watch:test:browserify",
+    "watch:test": "npm-run-all -p watch:test:manifest watch:test:js watch:test:webpack watch:test:browserify",
     "watch:test:js": "node scripts/watch-test.js",
     "watch:test:manifest": "node -e \"var b=require('./scripts/manifest-data.js'); b.watch();\"",
     "prepublish": "npm run build"
```
forbesjo: Might want to use [in-publish](https://www.npmjs.com/package/in-publish) so this doesn't run on `npm install`

forbesjo:
One suggestion but otherwise LGTM

PR #989, opened 2017-02-01
Add events for underflow and live resync

dmlap:
For QoS measurement purposes, it may be useful to know how often the playback watcher is activating. Add new events for when the player falls off the back of the live window or stalls due to a video buffer gap.

### DIFF
--- a/test/playback-watcher.test.js
+++ b/test/playback-watcher.test.js
@@ -173,6 +173,62 @@ QUnit.test('seek to live point if we fall off the end of a live playlist', funct
   assert.equal(seeks[0], 45, 'player seeked to live point');
 });
 
+QUnit.test('fires notifications when activated', function(assert) {
+  let buffered = [[]];
+  let seekable = [[]];
+  let currentTime = 0;
+  let liveresync = 0;
+  let videounderflow = 0;
+  let playbackWatcher;
+
+  this.player.src({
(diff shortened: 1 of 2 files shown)


test/playback-watcher.test.js
```diff
@@ -173,6 +173,62 @@ QUnit.test('seek to live point if we fall off the end of a live playlist', funct
...
+  this.player.tech_.on('videounderflow', () => videounderflow++);
+
+  currentTime = 19;
+  seekable[0] = [20, 30];
+  playbackWatcher.waiting_();
+  assert.equal(liveresync, 1, 'triggered a liveresync event');
+
+  currentTime = 12;
+  seekable[0] = [0, 100];
+  buffered[0] = [0, 9];
```
gesinger: Minor, but can do `buffered = [[0, 9], [10, 20]];`

test/playback-watcher.test.js
```diff
@@ -173,6 +173,62 @@ QUnit.test('seek to live point if we fall off the end of a live playlist', funct
...
+  buffered.push([10, 20]);
+  playbackWatcher.waiting_();
+  assert.equal(videounderflow, 1, 'triggered a videounderflow event');
```
gesinger: Might be worth verifying liveresync stays the same at the end.

PR #990, opened 2017-02-01
feature: Support for Akamai-style Redundant HLS

zhuangs:
## Description
Only play backup streams when the primary have failed to support for Akamai-style Redundant HLS

## Specific Changes proposed
Implement a stable sort for renditions
Change the way of select bandwidthBestVariants in selectPlaylist algorithm

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -136,6 +136,12 @@ Hls.STANDARD_PLAYLIST_SELECTOR = function() {
   let width;
   let height;
 
+  // In Chrome, the Array#sort function is not stable so add a
+  // presortIndex that we can use to ensure we get a stable-sort
+  sortedPlaylists.forEach(function(elem, idx) {
+    elem.presortIndex = idx;
+  });
+
   sortedPlaylists.sort(Hls.comparePlaylistBandwidth);
 
   // filter out any playlists that have been excluded due to
@@ -161,6 +167,10 @@ Hls.STANDARD_PLAYLIST_SELECTOR = function() {
       // since the playlists are sorted in ascending order by
       // bandwidth, the first viable variant is the best
       if (!bandwidthBestVariant) {
+        if (i !== 0 &&
+            sortedPlaylists[i - 1].attributes.BANDWIDTH === variant.attributes.BANDWIDTH) {
+          continue;
+        }
         bandwidthBestVariant = variant;
       }
     }
@@ -710,6 +720,9 @@ Hls.comparePlaylistBandwidth = function(left, right) {
   }
   rightBandwidth = rightBandwidth || window.Number.MAX_VALUE;
 
+  if (leftBandwidth === rightBandwidth) {
+    return left.presortIndex - right.presortIndex;
+  }
   return leftBandwidth - rightBandwidth;
 };
 
@@ -747,6 +760,9 @@ Hls.comparePlaylistResolution = function(left, right) {
   if (leftWidth === rightWidth &&
       left.attributes.BANDWIDTH &&
       right.attributes.BANDWIDTH) {
+    if (left.attributes.BANDWIDTH === right.attributes.BANDWIDTH) {
+      return right.presortIndex - left.presortIndex;
+    }
     return left.attributes.BANDWIDTH - right.attributes.BANDWIDTH;
   }
   return leftWidth - rightWidth;
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -167,6 +167,7 @@ Hls.STANDARD_PLAYLIST_SELECTOR = function() {
       // since the playlists are sorted in ascending order by
       // bandwidth, the first viable variant is the best
       if (!bandwidthBestVariant) {
+        // make sure the first playlist is chosen
         if (i !== 0 &&
             sortedPlaylists[i - 1].attributes.BANDWIDTH === variant.attributes.BANDWIDTH) {
           continue;
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -119,18 +119,18 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
 /**
  * Resuable stable sort function
  *
- * @param {Playlists} playlists
- * @param {Function} different comparators
+ * @param {Playlists} array
+ * @param {Function} sortFn Different comparators
  * @function stableSort
  */
 const stableSort = function(array, sortFn) {
-  let cmp;
+  let newArray = array.slice();
 
   array.sort(function(left, right) {
-    cmp = sortFn(left, right);
+    let cmp = sortFn(left, right);
 
     if (cmp === 0) {
-      return array.indexOf(left) - array.indexOf(right);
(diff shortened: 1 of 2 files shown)


src/videojs-contrib-hls.js
```diff
@@ -161,6 +167,10 @@ Hls.STANDARD_PLAYLIST_SELECTOR = function() {
       // since the playlists are sorted in ascending order by
       // bandwidth, the first viable variant is the best
       if (!bandwidthBestVariant) {
+        if (i !== 0 &&
+            sortedPlaylists[i - 1].attributes.BANDWIDTH === variant.attributes.BANDWIDTH) {
+          continue;
+        }
```
imbcmdth: This might be easier done as a `filter` to get all of the renditions with the same (highest) bandwidth and then taking the very first element.

src/videojs-contrib-hls.js
```diff
@@ -117,110 +117,107 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
 };
 
 /**
+ * Resuable stable sort function
+ *
+ * @param {Playlists} playlists
```
mjneil: The names here do not match the actual parameter names

src/videojs-contrib-hls.js
```diff
@@ -117,110 +117,107 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
...
+ * @function stableSort
+ */
+const stableSort = function(array, sortFn) {
+  let cmp;
+
+  array.sort(function(left, right) {
+    cmp = sortFn(left, right);
+
+    if (cmp === 0) {
+      return array.indexOf(left) - array.indexOf(right);
```
mjneil: `Array.sort` works on the array in place, so the indices of items will change as this function is sorting. You need to use the original indices of the items in the array for stability. Since we are trying to avoid adding properties to the original elements, I'd suggest creating a copy of the array before sorting to check `indexOf` on

src/videojs-contrib-hls.js
```diff
@@ -117,110 +117,107 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
...
-      resolutionPlusOneAttribute = resolutionPlusOne.attributes.RESOLUTION;
-    }
+  // if we have the exact resolution as the player use it
+  resolutionBestVariantList = haveResolution.filter(function(elem) {
+    return elem.attributes.RESOLUTION.width === width &&
+           elem.attributes.RESOLUTION.height === height;
+  });
+  // ensure that we pick the highest bandwidth variant that have exact resolution
+  resolutionBestVariant = resolutionBestVariantList.filter(function(elem) {
+    return elem.attributes.BANDWIDTH === resolutionBestVariantList.slice(-1)[0].attributes.BANDWIDTH;
```
mjneil: The slice call here is a lot of overhead just to get the last element of the array. You can just index into `length - 1` like you do in the bandwidthBestVariant filter above

src/videojs-contrib-hls.js
```diff
@@ -117,110 +117,107 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
...
+    });
+    // find all the variants have the same smallest resolution
+    resolutionPlusOneSmallest = resolutionPlusOneList.filter(function(elem) {
+      return elem.attributes.RESOLUTION.width === resolutionPlusOneList[0].attributes.RESOLUTION.width &&
+             elem.attributes.RESOLUTION.height === resolutionPlusOneList[0].attributes.RESOLUTION.height;
+    });
+    // ensure that we also pick the highest bandwidth variant that
+    // is just-larger-than the video player
+    resolutionPlusOne = resolutionPlusOneSmallest.filter(function(elem) {
+      return elem.attributes.BANDWIDTH === resolutionPlusOneSmallest.slice(-1)[0].attributes.BANDWIDTH;
```
mjneil: Same thing with this slice as the other comment

src/videojs-contrib-hls.js
```diff
@@ -117,110 +117,107 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
...
+  // find the smallest variant that is larger than the player
+  // if there is no match of exact resolution
+  if (!resolutionBestVariant) {
+    resolutionPlusOneList = haveResolution.filter(function(elem) {
+      return elem.attributes.RESOLUTION.width > width ||
+             elem.attributes.RESOLUTION.height > height;
+    });
+    // find all the variants have the same smallest resolution
+    resolutionPlusOneSmallest = resolutionPlusOneList.filter(function(elem) {
+      return elem.attributes.RESOLUTION.width === resolutionPlusOneList[0].attributes.RESOLUTION.width &&
```
mjneil: While I think in most cases, `resolutionPlusOneList[0]` will be the smallest resolution, the original list of playlists is only sorted by bandwidth. This leaves the possibility that a variant with lower bandwidth but higher resolution would come first in the list.

test/videojs-contrib-hls.test.js
```diff
@@ -652,6 +652,32 @@ QUnit.test('selects a playlist below the current bandwidth', function(assert) {
...
+    src: 'manifest/master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(this.player, this.clock);
+  standardXHRResponse(this.requests[0]);
+
+  this.player.tech_.hls.playlists.master.playlists[0].attributes.BANDWIDTH = 528;
+  this.player.tech_.hls.playlists.master.playlists[1].attributes.BANDWIDTH = 528;
+  this.player.tech_.hls.playlists.master.playlists[2].attributes.BANDWIDTH = 728;
+  this.player.tech_.hls.playlists.master.playlists[3].attributes.BANDWIDTH = 728;
```
mjneil: I think it would be good to test not only same bandwidth, but combinations of playlists that have the same resolution and bandwidths, playlists with same resolution and different bandwidths, and playlists with same bandwidths but different resolution.

src/videojs-contrib-hls.js
```diff
@@ -117,110 +117,107 @@ const handleHlsLoadedMetadata = function(qualityLevels, hls) {
...
 
 /**
+ * Resuable stable sort function
+ *
+ * @param {Playlists} playlists
+ * @param {Function} different comparators
+ * @function stableSort
+ */
+const stableSort = function(array, sortFn) {
+  let cmp;
```
imbcmdth: There is no reason to declare this variable outside of the sort function below.

src/videojs-contrib-hls.js
zhuangs: Not only sorted by bandwidth but also sorted by resolution in line 172

mjneil: You're right, I missed that. This should be fine then

test/videojs-contrib-hls.test.js
zhuangs: the unit test above and also 'selects the correct rendition by tech dimensions' one covered playlists with same resolution and different bandwidths, and playlists with same bandwidths but different resolution, added the playlists that have the same resolution and bandwidths.

mjneil:
LGTM

PR #996, opened 2017-02-06
Fix for EXT-X-MAP assumption of fragmented mp4s

gnujosh:
## Description
When using the EXT-X-MAP tag to specify the location of initialization metadata, it is currently assumed that a fragmented mp4 is being specified.  However, an MPEG-TS file can also utilize the EXT-X-MAP tag to specify the location of the first PAT and PMT packets in a byte stream.  This branch and pull request merely add a check in 3 different locations to ensure that a transport stream can be specified by EXT-X-MAP and still load.

## Testing performed
I've only been able to test this on the limited data that we have available.  To generate m3u8 files that index into our single .ts files, we used:

`fmpeg -i input.ts -codec copy -hls_flags single_file out.m3u8`

and then ran `tsinfo` on the original mpeg-ts file to get the packets of the first PAT and PMT packets, allowing us to fill in the EXT-X-MAP tag in the m3u8 file.

## Remaining question

I'm fairly new to javascript development and HTML MediaSources, so I was wondering if my change in `segment-loader.js` was appropriate.  If we append the ts fragment directly to the MediaSources buffer, video playing fails, because it isn't data in an mp4 container.

Is there other information in the PAT and PMT packets that isn't already extracted at that point?  Would the proper change be to append to the sourcebuffer either `initSegment.bytes` if using a fragmented mp4, or something like `mux_mpegts_to_mp4(initSegment.bytes)` if using a transport stream?

We haven't got a lot of variety of data for testing, so while the suggested change works for our data, I would be open to suggestions about a more proper conversion of mpeg-ts into mp4 that would be more future-proof.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -122,10 +122,9 @@ export const mimeTypesForPlaylist_ = function(master, media) {
   // An initialization segment means the media playlists is an iframe
   // playlist or is using the mp4 container. We don't currently
   // support iframe playlists, so assume this is signalling mp4
-  // fragments.
-  // the existence check for segments can be removed once
-  // https://github.com/videojs/m3u8-parser/issues/8 is closed
-  if (media.segments && media.segments.length && media.segments[0].map) {
+  // fragments if the uri doesn't end in .ts (which indicates mpeg2ts 
+  // segments).
+  if (media.segments.length && media.segments[0].map && !media.segments[0].uri.endsWith('.ts')) {
     container = 'mp4';
   }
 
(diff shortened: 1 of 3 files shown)


gnujosh:
I see a message above that "The Travis CI build failed".  I looked at the details and it is not clear to me why or even where it failed.  Perhaps someone else can give some advice on this.

gesinger:
Hey @gnujosh , thanks for the PR.

The Tavis CI build failed due to linting errors. One I noticed was an extra space at the end of line 125 of src/master-playlist-controller.js https://github.com/videojs/videojs-contrib-hls/pull/996/files#diff-3f8fe2d6cf175982cf8257063da92a81R125

As for the change, we may want to consider another option to test for transport streams. The only reason I mention that is that it is possible the URI will not end in .ts (for instance, there could also be GET data after the .ts). Do you have any other ideas offhand on methods we can use? Otherwise, we may have to consider inspecting the init segment to determine if it is mp4 or mpeg2ts.

forbesjo:

PR #998, opened 2017-02-08
update deps for 4.3.0 mediasources

mjneil:
## Description
contrib-media-sources includes a number of flash performance enhancements. This updates the version to 4.3.0

Also updated a few other dependencies so that the versions across projects are in line with each other. Also removed a few wildcards in dependencies so it is easier to know what version of media-sources and mux.js is included in the dist of contrib-hls

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,14 +89,13 @@
     "test/"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


imbcmdth:
LGTM

PR #1002, opened 2017-02-09
Resync on a very poor guess

imbcmdth:
## Description
The `SegmentLoader` is designed to generate a conservative guess that is always at or before the target time so that we can enter a mode where we fetch a contiguous set of segments and will be guaranteed to hit the "play head".

With very long playlists (> 2 hours) this conservative guess can be quite a bit off. For instance, a playlist with a duration of 5 hours can result in a guess that is a minute or more behind the playhead. The reason for this is that the only known segments will be early in the playlist and we attempt to make a guess some 5 hours and thousands of segments later often without very accurate timing info.

This PR addresses this by allowing the `SegmentLoader` to guess again but using the information learned from the original poor guess. In testing, this second guess is usually pretty close to perfect since the original guess was (relatively) fairly close to begin with.

To accomplish this some enhancements were necessary to `SyncController` as well. The `Segment` strategy did not take into account the `currentTime` and would always return the latest (in time) segment with timing information. Now we always return the segment closest to the time ensuring that guesses are more accurate.

Some minor reorganizing was done to `SegmentLoader` in the earlier commits so that function order better reflects the flow of execution and puts related functions near each other for easier development and understanding.

`SyncController` was improved in other small ways and no longer returns the very first sync-point found but tries *all* strategies and returns the one closest to the `currentTime` which is the most accurate sync-point to base calculations off of.

## Specific Changes proposed

## Requirements Checklist

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -205,6 +205,20 @@ export default class SegmentLoader extends videojs.EventTarget {
   }
 
   /**
+   * abort all pending xhr requests and null any pending segements
+   *
+   * @private
+   */
+  abort_() {
+    if (this.xhr_) {
+      this.xhr_.abort();
+    }
+
+    // clear out the segment being processed
+    this.pendingSegment_ = null;
+  }
+
+  /**
    * set an error on the segment loader and null out any pending segements
    *
    * @param {Error} error the error to set on the SegmentLoader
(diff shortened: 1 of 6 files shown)


imbcmdth:
I just realized this will not work well for Flash since it has a very different notion of the `buffered` value from MSE. Changes incoming!

imbcmdth:
Rebasing this for clarity...

PR #1004, opened 2017-02-10
isLowestEnabledRendition returns true when there are redundant streams

zhuangs:
## Description
PlaylistLoader#isLowestEnabledRendition_ never returns true when there are redundant streams

## Specific Changes proposed
return true when the lowest "rendition set" (all renditions sharing the same bandwidth) is selected.

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -245,7 +245,7 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
       return false;
     }
 
-    let currentBandwidth = loader.media().attributes.BANDWIDTH || 0;
+    let currentBandwidth = media.attributes.BANDWIDTH || 0;
 
     return !(loader.master.playlists.filter((playlist) => {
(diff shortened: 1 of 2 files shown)


src/playlist-loader.js
```diff
@@ -259,9 +259,9 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
       if (playlist && playlist.attributes) {
         bandwidth = playlist.attributes.BANDWIDTH;
       }
-      return bandwidth <= currentBandwidth;
+      return bandwidth < currentBandwidth;
 
-    }).length > 1);
+    }).length >= 1);
```
imbcmdth: I think this boolean logic could use some cleaning up. Right now it says `length not greater than or equal to one`  which is difficult to wrap my brain around. I think a better one is simply `length is zero`.

zhuangs: yes, that will be much clear

mjneil:
LGTM

PR #1006, opened 2017-02-13
Fix resuming live playback after long pauses

gesinger:
## Description
After a live video is paused long enough that the playlist no longer contains any segments previously requested, our seekable range is off and we often can't resume playback. This was for two primary reasons: (1) we were not updating seekable after the playlist updated enough that we no longer had any known segments and (2) if the segment times did not match up perfectly with the specified durations in the playlist, our seekable range would be off and we'd never try to adjust after seeking.

## Specific Changes proposed
This _seeks_ to resolve resuming live playback by keeping the seekable range updated, and checking to see if we are outside of a seekable range when we resume (after getting new sync info).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -79,7 +79,32 @@ export default class PlaybackWatcher {
    * @private
    */
   checkCurrentTime_() {
-    if (this.tech_.paused() || this.tech_.seeking()) {
+    if (this.tech_.seeking()) {
+      let seekable = this.seekable();
+
+      if (!seekable.length) {
+        return;
+      }
+
+      let seekableEnd = seekable.end(seekable.length - 1);
+      let currentTime = this.tech_.currentTime();
+
+      // handle any cases where we're trying to seek outside of the seekable range
+      // provide a buffer of .1 seconds to handle rounding/imprecise numbers
+      if (currentTime < seekable.start(0) - .1 ||
+          currentTime > seekableEnd + .1) {
+        // sync to live point (if VOD, our seekable was updated and we're
+        // simply adjusting)
+        this.logger_(`Trying to seek outside of seekable at time ${currentTime} with ` +
+                     `seekable range ${Ranges.printableRange(seekable)}. Seeking to ` +
+                     `${seekableEnd}.`);
+        this.tech_.setCurrentTime(seekableEnd);
+      }
+
+      return;
+    }
+
+    if (this.tech_.paused()) {
       return;
     }
 
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -79,28 +79,9 @@ export default class PlaybackWatcher {
    * @private
    */
   checkCurrentTime_() {
-    if (this.tech_.seeking()) {
-      let seekable = this.seekable();
-
-      if (!seekable.length) {
-        return;
-      }
-
-      let seekableEnd = seekable.end(seekable.length - 1);
-      let currentTime = this.tech_.currentTime();
-
-      // handle any cases where we're trying to seek outside of the seekable range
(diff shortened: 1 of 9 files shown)


src/ranges.js
```diff
@@ -318,11 +318,35 @@ const getSegmentBufferedPercent = function(startOfSegment,
...
+const printableRange = (range) => {
+  let str = '';
+
+  if (!range || !range.length) {
+    return str;
+  }
+
+  for (let i = 0; i < range.length; i++) {
+    if (i > 0) {
+      str += ', ';
```
mjneil: You can avoid this if you make `str` and array pushing each start+end and return `str.join(',')`. (It's also better for performance, not that these ranges will ever be large enough for that to matter)

test/playback-watcher.test.js
```diff
@@ -229,6 +229,146 @@ QUnit.test('fires notifications when activated', function(assert) {
...
+  this.player.tech_.trigger('waiting');
+  assert.equal(seeks.length, 2, 'did not seek');
+
+  currentTime = 50;
+  // if we're not seeking, the case shouldn't be handled here
+  seeking = false;
+  this.player.tech_.trigger('waiting');
+  assert.equal(seeks.length, 2, 'did not seek');
+
+  // no check for 0 with seekable false because that should be handled by live falloff
```
mjneil: seekable -> seeking

src/playback-watcher.js
```diff
@@ -79,6 +79,12 @@ export default class PlaybackWatcher {
    * @private
    */
   checkCurrentTime_() {
+    if (this.tech_.seeking() && this.fixBadSeeks_()) {
```
imbcmdth: Can we please, please rename `fixBadSeeks_` to `fixesBadSeeks_` because that would make me soooo happy.

src/playback-watcher.js
```diff
@@ -115,6 +121,31 @@ export default class PlaybackWatcher {
...
+   * Fixes situations where there's a bad seek
+   *
+   * @return {Boolean} whether an action was taken to fix the seek
+   * @private
+   */
+  fixBadSeeks_() {
+    let seekable = this.seekable();
+    let currentTime = this.tech_.currentTime();
+
+    if (this.tech_.seeking() &&
```
mjneil: It's probably safer to leave this check in here, just thought I'd note that I don't think we can get into this function with seeking is false since the check `this.tech_.seeking() && this.fixBadSeeks_()` won't call `fixBadSeeks_` if seeking returns false

src/playback-watcher.js
```diff
@@ -123,6 +154,10 @@ export default class PlaybackWatcher {
     let seekable = this.seekable();
     let currentTime = this.tech_.currentTime();
 
+    if (this.tech_.seeking() && this.fixBadSeeks_()) {
```
mjneil: Why is this check made in both `checkCurrentTime_` and `waiting_`?

gesinger: The waiting event can fire and cause `waiting_` to run independent of `checkCurrentTime_`

src/playback-watcher.js
imbcmdth: Should this function ever run in VOD?

It seems wise to check to see if the duration is "Infinity" before taking any other measures.

mjneil: I don't think its bad to have this for VOD. Seekable doesn't change in VOD anyway, so if someone tries seeking outside of the duration of the video, this would just adjust it to 0 or end of video depending on which direction they overseeked

gesinger: Yeah, it's a tough one to decide between.

imbcmdth:
LGTM!

PR #1014, opened 2017-02-15
Remove ad-hoc logging in favor of a config-enabled logging like playback-watcher

imbcmdth:
Pretty self-explanatory. The original logging wasn't meant to be a permanent solution.

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -468,7 +468,13 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
-    this.logger_('cB_', 'mediaIndex:', mediaIndex, 'hasPlayed:', hasPlayed, 'currentTime:', currentTime, 'syncPoint:', syncPoint, 'fetchAtBuffer:', this.fetchAtBuffer_, 'bufferedTime:', bufferedTime);
+    this.logger_('checkBuffer_',
+      'mediaIndex:', mediaIndex,
+      'hasPlayed:', hasPlayed,
+      'currentTime:', currentTime,
+      'syncPoint:', syncPoint,
+      'fetchAtBuffer:', this.fetchAtBuffer_,
+      'bufferedTime:', bufferedTime);
 
     // When the syncPoint is null, there is no way of determining a good
     // conservative segment index to fetch from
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -161,6 +151,10 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     // ...for determining the fetch location
     this.fetchAtBuffer_ = false;
+
+    if (settings.debug) {
+      this.logger_ = videojs.log.bind(videojs, 'segment-loader', this.loaderType_, '->');
```
mjneil: If `settings.debug` is `false`, won't all the calls to `this.logger_` throw errors since `this.logger_` will be `undefined`?

mjneil:
Changes need to be made in `MasterPlaylistController` so that the `debug` option is added onto the settings object passed to the segment loaders as well as a settings object needs to be passed to the sync controller

src/segment-loader.js
```diff
@@ -477,19 +468,21 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
+    this.logger_('cB_', 'mediaIndex:', mediaIndex, 'hasPlayed:', hasPlayed, 'currentTime:', currentTime, 'syncPoint:', syncPoint, 'fetchAtBuffer:', this.fetchAtBuffer_, 'bufferedTime:', bufferedTime);
```
gesinger: Really long line

gesinger: And it might be a good idea to just write out checkBuffer_

src/segment-loader.js
```diff
@@ -522,7 +515,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       mediaIndex = mediaSourceInfo.mediaIndex;
       startOfSegment = mediaSourceInfo.startTime;
     }
-    log('gMIFT', mediaIndex, 'sos', startOfSegment);
+    this.logger_('gMIFT', mediaIndex, 'sos', startOfSegment);
```
gesinger: While changing, would be good to write out getMediaIndexForTime

src/segment-loader.js
```diff
@@ -1104,4 +1097,12 @@ export default class SegmentLoader extends videojs.EventTarget {
...
       (segmentProcessingThroughput - rate) / (++this.throughput.count);
   }
+
+  /**
+   * A debugging logger noop that is set to console.log only if debugging
+   * is enabled globally
+   *
+   * @private
+   */
+  logger_() {}
```
gesinger: Instead of writing this function and the check for the debug setting in each module we want to log, we should just have the actual logger passed in, and check for the debug option at the highest level.

imbcmdth: I plan on doing this later by pulling in Pat's logger/history thing. This change was just to get rid of that ugly hack since we already have support for a less-ugly option in playback-watcher.

imbcmdth:
@mjneil Right now we are using global videojs HLS options for this. The settings object remains unchanged but the merger with the global options brings the debug flag. I decided to remove the options for sync-controller since it wasn't being used anyway.

PR #1015, opened 2017-02-15
Reorganized the functions in SegmentLoader to better follow the flow of execution from top-down

imbcmdth:
This is mostly a quality-of-life improvement. It was tedious to scroll around because execution didn't follow an orderly path through the file. Also helper functions and external functions were moved to the top for easier discovery.

PR #1016, opened 2017-02-15
Resync on poor initial segment choice

imbcmdth:
## Description
The `SegmentLoader` is designed to generate a conservative guess that is always at or before the target time so that we can enter a mode where we fetch a contiguous set of segments and will be guaranteed to hit the "play head".

With very long playlists (> 2 hours) this conservative guess can be quite a bit off. For instance, a playlist with a duration of 5 hours can result in a guess that is a minute or more behind the playhead. The reason for this is that the only known segments will be early in the playlist and we attempt to make a guess some 5 hours and thousands of segments later often without very accurate timing info.

This PR addresses this by allowing the `SegmentLoader` to guess again but using the information learned from the original poor guess. In testing, this second guess is usually pretty close to perfect since the original guess was (relatively) fairly close to begin with.

To accomplish this some enhancements were necessary to `SyncController` as well. The `Segment` strategy did not take into account the `currentTime` and would always return the latest (in time) segment with timing information. Now we always return the segment closest to the time ensuring that guesses are more accurate.

Some minor reorganizing was done to `SegmentLoader` in the earlier commits so that function order better reflects the flow of execution and puts related functions near each other for easier development and understanding.

`SyncController` was improved in other small ways and no longer returns the very first sync-point found but tries *all* strategies and returns the one closest to the `currentTime` which is the most accurate sync-point to base calculations off of.

## Specific Changes proposed

## Requirements Checklist

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -482,6 +482,18 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
+    let buffer = this.sourceUpdater_.buffered();
+
+    // If we have a buffer but it ends more than 3 targetDurations before the currentTime_
+    // that means that our conservative guess was too conservative. In that case, reset the
+    // loader state so that we try to use any information gained from the previous request
+    // to create a new, more accurate, sync-point.
+    if (buffer &&
+        buffer.length &&
+        this.currentTime_() - buffer.end(buffer.length - 1) > this.playlist_.targetDuration * 3) {
+      this.resetLoader();
+    }
+
     if (!this.syncPoint_) {
       this.syncPoint_ = this.syncController_.getSyncPoint(this.playlist_,
                                                           this.mediaSource_.duration,
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1070,8 +1070,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // the currentTime_ that means that our conservative guess was too conservative.
     // In that case, reset the loader state so that we try to use any information gained
     // from the previous request to create a new, more accurate, sync-point.
-    if (segment &&
-        segment.end &&
+    if (segment.end &&
         this.currentTime_() - segment.end > segmentInfo.playlist.targetDuration * 3) {
       this.resetLoader();
     } else {
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -482,22 +482,11 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    let buffer = this.sourceUpdater_.buffered();
-
-    // If we have a buffer but it ends more than 3 targetDurations before the currentTime_
-    // that means that our conservative guess was too conservative. In that case, reset the
-    // loader state so that we try to use any information gained from the previous request
-    // to create a new, more accurate, sync-point.
-    if (buffer &&
-        buffer.length &&
-        this.currentTime_() - buffer.end(buffer.length - 1) > this.playlist_.targetDuration * 3) {
-      this.resetLoader();
-    }
-
     if (!this.syncPoint_) {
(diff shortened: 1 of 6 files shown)


src/segment-loader.js
```diff
@@ -1047,13 +1049,31 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        this.currentTime_() - segment.end > segmentInfo.playlist.targetDuration * 3) {
+      this.resetLoader();
+    } else {
+      this.mediaIndex = segmentInfo.mediaIndex;
+      this.fetchAtBuffer_ = true;
+    }
+
+    // Don't do a rendition switch unless the SegmentLoader is already walking forward
+    if (isWalkingForward) {
+      this.trigger('progress');
```
mjneil: I understand this is to limit ping-ponging between renditions without at least appending another segment, but i worry this could cause unwanted timeout errors. I could imagine that in a live scenario where we are close to the live point, and we do a rendition switch, now if this is a new rendition, we have to wait for a sync point, then another load + append, and another load + append before we are considered walkingForward. If in this intermediate time, we reach the end of the buffer and waiting on buffering, we may not fire progress events for some time if bandwidth conditions cause those loads to take a long time. This could cause a timeout when we otherwise would be notifying the player that we are making some type of progress

mjneil:
Don't forget to migrate the change from [here](https://github.com/videojs/videojs-contrib-hls/pull/1015#discussion_r101586849) to this PR

test/sync-controller.test.js
```diff
@@ -58,34 +67,38 @@ QUnit.test('returns correct sync point for ProgramDateTime strategy', function(a
...
+  let strategy = getStrategy('Segment');
   let playlist = {
     segments: [
       { timeline: 0 },
       { timeline: 0 },
+      { timeline: 1, start: 10 },
+      { timeline: 1, start: 20 },
       { timeline: 1 },
       { timeline: 1 },
-      { timeline: 1, start: 30 },
```
gesinger: Would be good to have a test with exact start as currentTime as well

src/sync-controller.js
```diff
@@ -45,54 +45,83 @@ export const syncPointStrategies = [
...
             typeof segment.start !== 'undefined') {
-          let syncPoint = {
-            time: segment.start,
-            segmentIndex: i
-          };
+          let distance = Math.abs(currentTime - segment.start);
+
+          // Once the distance begins to increase, we have passed
+          // currentTime and can stop looking for better candidates
+          if (lastDistance && lastDistance < distance) {
```
imbcmdth: I'm not sure I understand. How can you ever have two segments with the same start-time?

gesinger: Not two segments with the same start time, but if `currentTime` is equal to `segment.start` then `distance` is 0 and the `syncPoint` would be set to the following segment (if one exists) because when `lastDistance` is 0 for the next iteration, this condition is false and the next condition (where it checks for `!lastDistance`) would result in true (leading to an overwrite of syncPoint).

imbcmdth: Oh, your complaint was that I need to do `lastDistance !== null` ... Why did you just say so! 😆

src/segment-loader.js
imbcmdth: I discussed this with Matt in person. Basically, there are trade-offs either way. With this in place we may not respond to bandwidth issues as quickly but we ensure that we are always in a stable playback mode before we do switch. This is less important for VOD but it's a different story for live. One problem is that `seekable` can become stalled because we are changing playlists without having sync. Another problem is that we can conservatively guess multiple times in a row and will end up fetching the same segment over and over at different renditions (worst-case being ping-ponging as Matt put it.) This change should always result in a sync point being established and some forward progress being made before a rendition is allowed to change. Timeout-caused rendition changes will still happen immediately so that the player can take emergency action.

PR #1019, opened 2017-02-17
Add support for description audio tracks in HLS

OwenEdwards:
## Description
An audio track can be included in an HLS mux which contains [description](https://en.wikipedia.org/wiki/Audio_description) for blind or low-vision users. This track is indicated by including a `CHARACTERISTICS` key, with value `public.accessibility.describes-video`.

This PR, and [m3u8-parser PR 14](https://github.com/videojs/m3u8-parser/pull/14), mark the `kind` of the audio track to indicate that it is `main-desc` (main and description), so that the player can specifically handle this track in a different way.

## Please describe the change as necessary.
Use CHARACTERISTICS of an audio track when determining its `kind`.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -288,7 +288,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
       let media = this.masterPlaylistLoader_.media();
       let requestTimeout = (this.masterPlaylistLoader_.targetDuration * 1.5) * 1000;
 
-      this.requestOptions_.timeout = requestTimeout;
+      // If we don't have any more available playlists, we don't want to
+      // timeout the request.
+      if (this.masterPlaylistLoader_.isLowestEnabledRendition_()) {
+        this.requestOptions_.timeout = 0;
+      } else {
+        this.requestOptions_.timeout = requestTimeout;
+      }
 
       // if this isn't a live video and preload permits, start
       // downloading segments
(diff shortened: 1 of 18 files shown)


OwenEdwards:
If/when PR 14 get's pulled into m3u8-parser and released, I'll update the `package.json` file to reflect the version of that package that is needed to support this.

src/master-playlist-controller.js
```diff
@@ -569,6 +569,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+   * Convert the properties of an HLS track into an audioTrackKind.
+   *
+   * @private
+   */
+  audioTrackKind_(properties) {
+    let kind = properties.default ? 'main' : 'alternative';
+
+    if (properties.characteristics) {
+      if (properties.characteristics.indexOf('public.accessibility.describes-video') >= 0) {
+        kind = 'main-desc';
```
OwenEdwards: In theory that's possible; I'd assume it would need to follow the same rules as, say, a CAPTIONS or SUBTITLES tracks:

https://tools.ietf.org/html/draft-pantos-http-live-streaming-20#section-4.3.4.1.1:
> All EXT-X-MEDIA tags in a Playlist MUST meet the following
>    constraints:
> 
>    o  All EXT-X-MEDIA tags in the same Group MUST have different NAME
>       attributes.
> 
>    o  A Group MUST NOT have more than one member with a DEFAULT
>       attribute of YES.
> 
>    o  All members of a Group whose AUTOSELECT attribute has a value of
>       YES MUST have LANGUAGE [RFC5646] attributes with unique values.
> 

So you could tell them apart by `name` (`label`) and `language`. But it wouldn't change the audio track's `kind`. Is that what you were wondering about?

src/master-playlist-controller.js
```diff
@@ -569,6 +569,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
   /**
+   * Convert the properties of an HLS track into an audioTrackKind.
+   *
+   * @private
+   */
+  audioTrackKind_(properties) {
+    let kind = properties.default ? 'main' : 'alternative';
+
+    if (properties.characteristics) {
+      if (properties.characteristics.indexOf('public.accessibility.describes-video') >= 0) {
```
OwenEdwards: Sure - I wasn't sure how to deal with the line getting too long! I'll change that.

OwenEdwards:
I'll take a look at how to set up some tests.

mjneil:
Thanks for the contribution @OwenEdwards

PR #1023, opened 2017-02-22
Fix a bug with the combination of seek-to-live and resync-on-a-poor-guess behaviors

imbcmdth:
## Description
1. When the seekable window gets far out of sync from being pretty far in the future from the last known sync-point, we can set currentTime to a value that is unreachable
2. Once the segment is loaded, we have enough information to adjust seekable to the correct value
3. The original segment fetched from the “very wrong” currentTime is actually the right segment (it was based on the same, poor information)
4. Doing a seek-to-live again after we adjust the seekable window will put the currentTime at the end of the segment we just fetched
5. If we perform these operations in the right order we can avoid any spurious segment requests (ie. not trigger the resync-on-a-poor-guess behavior)

## Specific Changes proposed
1. Stop emitting `syncinfoupdate` from SyncController after probing a segment
   * Do that manually later in SegmentLoader
2. In `handleUpdateEnd_` use `resetEverything` to delete any portion of the segment that might be buffered when we re-sync
3. Add a new event `seekablechanged` to PlaybackWatcher so that it can immediately perform a seek-to-live when the seekable window has changed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -957,6 +957,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
                                                        mainSeekable.end(0)
       ]]);
     }
+
+    this.tech_.trigger('seekablechanged');
   }
 
   /**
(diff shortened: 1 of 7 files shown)


forbesjo:
Looks good to me

PR #1026, opened 2017-02-22
Never blacklist the final rendition available

zhuangs:
## Specific Changes proposed
Stop blacklisting the final rendition


src/playlist-loader.js
```diff
@@ -264,6 +264,17 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     }).length > 1);
   };
 
+  /**
+   * Returns whether the current playlist is the final available rendition
+   *
+   * @return {Boolean} true if on final rendition
+   */
+  loader.isFinalRendition_ = function() {
+    return (loader.master.playlists.filter((playlist) => {
```
mjneil: You can just pass `loader.master.playlists.filter(isEnabled)` for the callback instead of defining an anonymous function that just calls it

test/videojs-contrib-hls.test.js
```diff
@@ -1031,13 +1031,18 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
   assert.ok(media.excludeUntil > 0, 'original media blacklisted for some time');
   assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
 
+  // request for the final available media
+  this.requests[2].respond(400);
+  url = this.requests[2].url.slice(this.requests[2].url.lastIndexOf('/') + 1);
+  media = this.player.tech_.hls.playlists.master.playlists[url];
+
+  assert.ok(!media.excludeUntil, 'dont blacklist the second media because its final media');
+  assert.equal(this.env.log.warn.calls, 0, 'warning logged for blacklist');
```
mjneil: Description doesn't match what is being asserted

src/playlist-loader.js
zhuangs: yes, nice call, changed it

test/videojs-contrib-hls.test.js
zhuangs: changed it to 'media not blacklisted after playlist 404'

src/master-playlist-controller.js
```diff
@@ -794,23 +794,20 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
 
       return this.masterPlaylistLoader_.media(nextPlaylist);
     }
-    videojs.log.warn('Problem encountered with the current ' +
-                     'HLS playlist. No suitable alternatives found.');
-    // We have no more playlists we can select so we must fail
-    this.error = error;
-    return this.mediaSource.endOfStream('network');
+    // Never blacklisting this playlist because it's final rendition
+    return this.masterPlaylistLoader_.media(currentPlaylist);
```
mjneil: if `currentPlaylist === this.masterPlaylistLoader_.media()` the call to `media()` on the playlist loader will exit early without starting another request for the playlist because it considers it not a change in media. Playlist loader should continue refreshing the live playlist until either it recovers, or the player timesout with a `PLAYER_ERR_TIMEOUT`

zhuangs: so if it's the final rendition, just return without call the media() function, is this correct?

mjneil: No I don't think that will work. When a playlist request errors, `PlaylistLoader` triggers an error and returns without setting up another request. At this point we want to kick `PlaylistLoader` back into gear to start requesting the final rendition again. Calling `media()` will normally do this as long as the media passed in is not the same as the current media. Calling `load()` will also kickstart the loader, but depending on the loader's `state` after an error occurred, a `load` may never actually get the loader started. This will probably require some changes to the way `PlaylistLoader` handles errors so after an error, it gets reset to a state that it can start loading again

zhuangs: changed this

zhuangs: changed it to set the loader's state when detect it's the final rendition and before calling ```load()```. Trying to do this because I think reset the state after an error happens may mess the normal error handle logic of those non-final renditions. But not sure if it's the right way to do this.

mjneil: This definitely has the effect we are going for, but changing the playlist loader's state outside of itself like this can make debugging difficult because now you have to know about this one line of code in a separate file that changes the state. I think we can safely restore the state in certain error conditions without introducing regressions. There are only a few places that errors trigger in `PlaylistLoader` and some places already use the `playlistRequestError` function which the optional third parameter takes a state value to reset that to.  I think you should be able to safely reset the state in each error condition using this.

imbcmdth:
We need to test the behavior when a segment request errors on the final available rendition. Do we re-request? Give up somehow?

mjneil:
@imbcmdth Right now, if a segment request fails on the last rendition, the segment loader is paused and the player will either stall at the end of the buffer until it times out or if its live, falls off the live window and playback watcher gets things moving again when seeking to live. Do you think we should unpause the segment loader in this case.

Also, there is nothing currently limiting the requests made. When the final rendition fails, it will keep trying to reload it indefinitely (or for at least 5 minutes until the other renditions are no longer blacklisted). Is this ok behavior?

imbcmdth:
@mjneil For the first paragraph: Sounds like we need some sort of retry logic for segment requests in this case. We are out of options so just keep retrying for 45 seconds.

For the second paragraph: What requests are you talking about here? Manifest refreshes?

mjneil:
Well the segment loader gets paused on the error and reset to `READY`, so it should just be as simple as a call to `load` to get the segment loader working again. 

Talking manifest refreshes, but in this case the retry happens right away instead of setting another timeout for refresh delay, so it repeated requests as fast as they respond/fail. (If we move to retry segments then we are talking segment requests too)

imbcmdth:
The manifest refreshing as fast as possible is a bug and a huge bandwidth hog for customers. That needs to be fixed!

src/playlist-loader.js
```diff
@@ -432,7 +432,14 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
    */
-  loader.load = () => {
+  loader.load = (isFinalRendition) => {
+    let refreshDelay;
+
+    if (isFinalRendition) {
+      refreshDelay = loader.media() ? loader.media().targetDuration * 1000 : 10 * 1000;
+      window.clearTimeout(mediaUpdateTimeout);
+      mediaUpdateTimeout = window.setTimeout(loader.load(false), refreshDelay);
+    }
```
mjneil: The `mediaUpdateTimeout` should be cleared immediately after this `if` block in case `load` is called from somewhere else and that caller doesn't want to wait the delay, we dont want the delayed call to run as well

zhuangs: changed! nice catch :)

src/playlist-loader.js
```diff
@@ -421,7 +432,15 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
    * start loading of the playlist
    */
-  loader.load = () => {
+  loader.load = (isFinalRendition) => {
+    let refreshDelay;
+
+    if (isFinalRendition) {
+      refreshDelay = loader.media() ? loader.media().targetDuration * 1000 : 10 * 1000;
+      window.clearTimeout(mediaUpdateTimeout);
+      mediaUpdateTimeout = window.setTimeout(loader.load(false), refreshDelay);
```
mjneil: `setTimeout` needs to be passed a callback that calls `loader.load(false)`. You could also bind a call to load with false passed, `loader.load.bind(null, false)`

src/playlist-loader.js
```diff
@@ -432,12 +452,12 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
       loader.start();
     }
   };
-
```
gesinger: Was this meant to be removed?

src/master-playlist-controller.js
```diff
@@ -360,6 +360,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.updateDuration();
 
       if (!updatedPlaylist.endList) {
+        // kick segmentloader back into gear if the playlist isn't
```
gesinger: Why is this only done for live playlists?

src/master-playlist-controller.js
```diff
@@ -820,23 +826,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return this.mediaSource.endOfStream('network');
     }
 
-    // Blacklist this playlist
-    currentPlaylist.excludeUntil = Date.now() + BLACKLIST_DURATION;
+    if (!this.masterPlaylistLoader_.isFinalRendition_()) {
```
gesinger: Really minor, and is mostly style preference, but if you change to use a check for it being the final rendition, and handling that case and returning, only one line needs to be indented, and the standard case (of blacklisting) can follow.

src/playlist-loader.js
```diff
@@ -392,7 +401,9 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
       }
 
       if (error) {
-        return playlistRequestError(request, loader.media().uri);
+        let startingState = 'HAVE_METADATA';
```
gesinger: Don't need the variable if only used in one place.

src/playlist-loader.js
```diff
@@ -464,6 +484,11 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
           // MEDIA_ERR_NETWORK
           code: 2
         };
+        if (loader.state === 'HAVE_NOTHING') {
+          loader.started = false;
+        } else {
+          loader.state = startingState;
```
gesinger: Is something else changing the state necessitating a reversion to the starting state?

test/playlist-loader.test.js
```diff
@@ -691,7 +691,7 @@ QUnit.test('aborts in-flight playlist refreshes when switching', function(assert
   assert.strictEqual(this.requests[0].aborted, true, 'aborted refresh request');
   assert.ok(!this.requests[0].onreadystatechange,
            'onreadystatechange handlers should be removed on abort');
-  assert.strictEqual(loader.state, 'SWITCHING_MEDIA', 'updated the state');
+  assert.strictEqual(loader.state, 'HAVE_METADATA', 'the state is set accoring to the startingState');
```
gesinger: Why is it that it no longer is set to SWITCHING_MEDIA? Is SWITCHING_MEDIA still needed/used as a state?

test/videojs-contrib-hls.test.js
```diff
@@ -1084,13 +1084,19 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
-
   assert.ok(media.excludeUntil > 0, 'original media blacklisted for some time');
   assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
 
+  // request for the final available media
+  this.requests[2].respond(400);
+  url = this.requests[2].url.slice(this.requests[2].url.lastIndexOf('/') + 1);
+  media = this.player.tech_.hls.playlists.master.playlists[url];
+
+  // media didn't be blacklisted because it's fianl rendition
```
gesinger: media wasn't blacklisted because it's the final rendition

test/videojs-contrib-hls.test.js
```diff
@@ -1084,13 +1084,19 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  assert.ok(!media.excludeUntil, 'media not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 0, 'warning logged for blacklist');
```
gesinger: Isn't this saying that no warning was logged?

test/videojs-contrib-hls.test.js
```diff
@@ -2769,6 +2775,13 @@ QUnit.test('blacklists playlist if key requests fail', function(assert) {
   hls.mediaSource.trigger('sourceopen');
   this.requests.shift()
     .respond(200, null,
+              '#EXTM3U\n' +
```
gesinger: Why is this now responding with a master playlist? The URL above (provided as `src`) seems to indicate that it is a media playlist.

test/videojs-contrib-hls.test.js
```diff
@@ -2802,6 +2815,13 @@ QUnit.test('treats invalid keys as a key request failure and blacklists playlist
...
```
gesinger: (As per above): Why is this now responding with a master playlist? The URL above (provided as `src`) seems to indicate that it is a media playlist.

src/master-playlist-controller.js
zhuangs: I think it's because we don't want the segment requests to go off too much and VOD triggers the ```loadedplaylist``` immediately. cc @mjneil

src/playlist-loader.js
zhuangs: seems unintentionally

test/videojs-contrib-hls.test.js
zhuangs: Because it has to have multiple playlists then the playlist which key requests fail can be blacklisted. It seems I forgot to change the media.m3u8 to master.m3u8.

test/playlist-loader.test.js
zhuangs: Because when the refresh request is aborted, ```playlistRequestError``` is called and we reset the loader.state to startingState(HAVE_METADATA). so that no longer is set to SWITCHING_MEDIA. I think if switching happens without error, the loader.state as SWITCHING_MEDIA won't be affected.

src/master-playlist-controller.js
mjneil: I believe that is why, but now since the playlist loader has a delay, it should probably be done for both live and vod

src/master-playlist-controller.js
```diff
@@ -359,6 +359,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.mainSegmentLoader_.playlist(updatedPlaylist, this.requestOptions_);
       this.updateDuration();
 
+      // kick segmentloader back into gear if the playlist isn't paused
```
gesinger: playlist => player

gesinger: Though I think it could make sense to provide a bit more context around why we're calling load if it isn't paused (what exactly is kicking it back into gear). Maybe something like `If the player isn't paused, ensure that the segment loader is running, as it is possible that it was temporarily stopped while waiting for a playlist (e.g., in case the playlist errored and we re-requested it).`.

src/master-playlist-controller.js
```diff
@@ -806,6 +811,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
   blacklistCurrentPlaylist(error = {}) {
     let currentPlaylist;
     let nextPlaylist;
+    let isFinalRendition = this.masterPlaylistLoader_.isFinalRendition_();
```
gesinger: Probably better to keep this closer to where it is used.

src/master-playlist-controller.js
```diff
@@ -820,23 +826,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
 
-      return this.masterPlaylistLoader_.media(nextPlaylist);
-    }
-    videojs.log.warn('Problem encountered with the current ' +
-                     'HLS playlist. No suitable alternatives found.');
-    // We have no more playlists we can select so we must fail
-    this.error = error;
-    return this.mediaSource.endOfStream('network');
+    return this.masterPlaylistLoader_.media(nextPlaylist);
+
```
gesinger: Extra line

src/master-playlist-controller.js
gesinger: I think if you move it closer you can also remove the check for `loader.master` in `loader.isFinalRendition_`

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,54 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
   assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
 
+  // request for the final available media
+  this.requests[2].respond(404);
+  url = this.requests[2].url.slice(this.requests[2].url.lastIndexOf('/') + 1);
+  media = this.player.tech_.hls.playlists.master.playlists[url];
+
+  // media wasn't blacklisted because it's final rendition
+  assert.ok(!media.excludeUntil, 'media not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 0, 'no warning logged for blacklist');
```
gesinger: We probably still want some kind of warning log.

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,54 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  assert.strictEqual(4, this.requests.length, 'one more request was made');
+  assert.strictEqual(this.requests[3].url,
+                     absoluteUrl('manifest/media1.m3u8'),
+                     'media playlist requested');
+
   // verify stats
   assert.equal(this.player.tech_.hls.stats.bandwidth, 1e10, 'bandwidth set above');
 });
 
+QUnit.test('never blacklist the playlist if its the only playlist', function(assert) {
```
gesinger: it's

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,54 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  let media;
+
+  this.player.src({
+    src: 'manifest/media.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(this.player, this.clock);
+
+  this.requests[0].respond(404);
+  media = this.player.tech_.hls.mediaSource;
```
gesinger: `this.player.tech_.hls.playlists.media()`

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,54 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  // request for the final available media
+  this.requests[2].respond(404);
+  url = this.requests[2].url.slice(this.requests[2].url.lastIndexOf('/') + 1);
+  media = this.player.tech_.hls.playlists.master.playlists[url];
+
+  // media wasn't blacklisted because it's final rendition
+  assert.ok(!media.excludeUntil, 'media not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 0, 'no warning logged for blacklist');
+
+  this.clock.tick(10 * 1000);
```
gesinger: Should this be half the target duration / 5 seconds? Might be worth checking with two clock ticks that add up to verify that it uses the target duration / 5 seconds, and doesn't happen before.

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,73 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+
+  // media wasn't blacklisted because it's final rendition
+  assert.ok(!media.excludeUntil, 'media not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
+  assert.equal(this.env.log.warn.args[1],
+              'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
+              'log specific error message for final playlist');
+
+  this.clock.tick(2 * 1000);
+  // no more request was made since it hasn't been half the segment duration
```
gesinger: ~~more~~

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,73 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  assert.strictEqual(3, this.requests.length, 'no more request was made');
```
gesinger: ~~more~~

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,73 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  assert.strictEqual(4, this.requests.length, 'one more request was made');
+  assert.strictEqual(this.requests[3].url,
+                     absoluteUrl('manifest/media1.m3u8'),
+                     'media playlist requested');
 
   // verify stats
   assert.equal(this.player.tech_.hls.stats.bandwidth, 1e10, 'bandwidth set above');
 });
 
+QUnit.test('never blacklist the playlist if it is the only playlist', function(assert) {
```
gesinger: I think we need one more test for the behavior of a playlist error (404 or 500) on the first request for that playlist (that it should fail fast when there is only that one media playlist and no master). It looks like this was changed such that it gets a response first and is never blacklisted, which is a good test to have, but we should have the other case too.

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,87 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+
+  // media wasn't blacklisted because it's final rendition
+  assert.ok(!media.excludeUntil, 'media was not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
+  assert.equal(this.env.log.warn.args[0],
+              'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
+              'log specific error message for final playlist');
+});
+
+QUnit.test('playlist error on the first request makes playlist fails fast when there is only that one media playlist', function(assert) {
```
gesinger: Long line, and maybe `error on the first playlist request triggers a media source error when there is only a media playlist`

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1077,87 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  this.player.src({
+    src: 'manifest/media.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(this.player, this.clock);
+
+  this.requests.shift().respond(404);
+
+  assert.equal(this.player.tech_.hls.mediaSource.error_,
```
gesinger: For good measure, maybe have an assert before the request that ensures it does not exist (maybe something like `assert.notOk(this.player.tech_.hls.mediaSource.error_, 'no media source error');`

zhuangs: added :)

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1080,100 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+
+  // media wasn't blacklisted because it's final rendition
+  assert.ok(!media.excludeUntil, 'media was not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
+  assert.equal(this.env.log.warn.args[0],
+              'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
+              'log specific error message for final playlist');
+});
+
+QUnit.test('error on the first playlist request triggers a media source error ' +
```
gesinger: I think this should be `error on the first playlist request doesn't error `

test/videojs-contrib-hls.test.js
```diff
@@ -1077,24 +1080,100 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+
+  // media wasn't blacklisted because it's final rendition
+  assert.ok(!media.excludeUntil, 'media was not blacklisted after playlist 404');
+  assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
+  assert.equal(this.env.log.warn.args[0],
+              'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
+              'log specific error message for final playlist');
+});
+
+QUnit.test('error on the first playlist request does not error ' +
```
gesinger: This was entirely my fault, but reading it again, I think it should be `error on the first playlist request does not trigger an error`

zhuangs: I thought we were using error as verb :)

mjneil:
Closing as this is now contained within #1039

Issue #1022, opened 2017-02-21
VideoJS not playing high bitrate HLS video


arshanawais:
I am trying to play an m3u8 file served by Wowza steaming engine at following address:

http://54.218.7.156:1935/vod/mp4:sample360.mp4/playlist.m3u8

The video has a high bitrate i.e. 13 mb/s. When I try to play it in VideoJS, network tab of the browser shows that the player keeps downloading the .ts file chunks again and again at regular intervals but never plays the file. I have tried this in Chrome & Firefox.

Same issue occurs with live stream instead of VOD. But in that case, instead of .ts file chunks, the m3u8 file keeps downloading again and again.

Low bitrate videos are working well in both live and VOD.

mjneil:
This looks to be a bug. If the last rendition available in the playlist is playing, requests shouldn't timeout, but these requests are. It looks like the check for last rendition is only called on mediachange events, which should also happen for the initial rendition

vpowers:
I've tested this fix with PR https://github.com/videojs/videojs-contrib-hls/pull/1006 and playback is working as expected.  Test was done with a long running live HLS stream using Wowza.  I can't confirm in the master branch (after merge) since playback of HLS is not longer working.  Possibly related https://github.com/videojs/videojs-contrib-hls/issues/1022.

mjneil:
This should be fixed by #1032

PR #1032, opened 2017-02-24
do not timeout segment requests for non-master playlist source

mjneil:
## Description
If the source loaded into the player is just a media playlist and not a master playlist, segments will still timeout after `1.5 * targetDuration` even though it is playing the lowest enabled rendition (seeing how its the only rendition)

Addresses #1022 

## Requirements Checklist

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -239,21 +239,13 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
    * @return {Boolean} true if on lowest rendition
    */
   loader.isLowestEnabledRendition_ = function() {
-    let media = loader.media();
-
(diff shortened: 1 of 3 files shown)


src/playlist-loader.js
```diff
@@ -241,10 +241,18 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
   loader.isLowestEnabledRendition_ = function() {
     let media = loader.media();
 
-    if (!media || !media.attributes) {
+    if (!media) {
       return false;
     }
 
+    if (!media.attributes) {
```
imbcmdth: I think a better way would be to just always return `true` if there is only a single variant.

PR #1039, opened 2017-03-06
blacklist the playlist that has stopped being updated and never blacklist the final playlist available

zhuangs:
## Description
Detect that playlist has stopped being updated in a timely manner and blacklist the playlist
Never blacklist the final playlist available #1026


src/master-playlist-controller.js
```diff
@@ -361,6 +370,32 @@ export class MasterPlaylistController extends videojs.EventTarget {
         } else {
           addSeekableRange();
         }
+
+        buffered = this.tech_.buffered();
```
mjneil: I think it would be good to pull out this section of calculations to a separate function as this event handler is getting quite gnarly

src/master-playlist-controller.js
```diff
@@ -361,6 +370,32 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+        if (this.seekable().length) {
+          seekableEnd = this.seekable().end(0);
+        }
+
+        // blacklist this playlist if the playhead is at the end of the buffer
+        // and the buffered data ends at the end of the last segment in the playlist
+        if (bufferedTime <= Ranges.TIME_FUDGE_FACTOR &&
+            endTime <= Ranges.TIME_FUDGE_FACTOR &&
+            currentTime > seekableEnd) {
```
mjneil: `seekableEnd`, `lastBufferedEnd` and `endTime` have the potential of being NaN at this point since `seekableEnd` and `lastBufferedEnd` may never be assigned a value if the conditionals for the if statements fail, and `playEnd` may be null from `this.playlistEnd()` affecting the calculation of `endTime`

src/master-playlist-controller.js
```diff
@@ -233,6 +234,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.seekable_ = videojs.createTimeRanges();
     this.hasPlayed_ = () => false;
+    this.playlistEnd = playlistEnd;
```
mjneil: It isn't necessary to attach this function to `MasterPlaylistController`, you can just call `playlistEnd` directly where you need it. On top of that, if you attach `playlistEnd` to the default export of `playlist.js`, the default export is already included in the `Hls` object available in this file. If you go that route, you do not need to import this function in a seperate import call, and instead could just call `Hls.Playlist.playlistEnd()`

src/playlist.js
```diff
@@ -308,6 +308,40 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
 
 /**
+ * Returns the end point the the playlist
+ *
+ * @param {Object} playlist a media playlist object
+ * @returns {Object} an object containing the time length of the playlist
+ * @function playlistEnd
+ */
+
+export const playlistEnd = function(playlist) {
```
mjneil: this function is very similar to `seekable` defined just below. Is there any way we can reduce the amount of repeat code here?

mjneil:
Would be good to have a some tests for `playlistEnd`

src/master-playlist-controller.js
zhuangs: I attached this function to ```MasterPlaylistController``` because I was trying to call this function in the ```videojs-contrib-hls.test.js``` file. but maybe there is better way to do this

src/master-playlist-controller.js
zhuangs: I thought ```seekableEnd```, ```lastBufferedEnd```, ```endTime``` being NaN may only happen at the very beginning of a playlist and that isn't the case we should ```blacklistCurrentPlaylist```, is this correct? In that case, it doesn't affect the ```if``` statement. Did you mean we should check ```seekableEnd```, ```lastBufferedEnd```, ```endTime``` isNaN() inside the ```if```?

src/master-playlist-controller.js
mjneil: Ah didn't notice that. Hmmmm, its not a big deal to attach the function, but if you come up with another solution that doesnt require it, thatd be great

zhuangs: yes, I found directly override ```Hls.Playlist.playlistEnd()``` in the test file is better :)

src/master-playlist-controller.js
mjneil: With `lastBufferedEnd` for example, you declare `lastBufferedEnd` at the start of this function, so it is `undefined`. The only location `lastBufferedEnd` is assigned a value is on line 378, which is inside the `if (buffered.length)` check. If `buffered.length === 0` then `lastBufferedEnd` will still be `undefined` and any calculation involving `lastBufferedEnd` e.g. `endTime = playEnd - lastBufferedEnd` will result in `NaN`.

With that said, looking at this again, it may be ok for this to happen. Since any comparison with `NaN` will return false, `seekable.length === 0` or `buffered.length === 0` probably means we don't have enough information to know if the playlist stopped updating and probably shouldn't blacklist it. This will require testing to see if we can create a scenario in which those values result in `NaN` but we still want to blacklist

src/master-playlist-controller.js
zhuangs: changed this

src/playlist.js
zhuangs: changed this

src/playlist.js
```diff
@@ -305,6 +307,43 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
 
     return segmentSync.time - sumDurations(playlist, syncIndex, 0);
   }
+  return 0;
```
mjneil: This should return `null` by default isntead of `0`. If there isn't enough information to determine expired time, returning `0` would be sending false information if there actually is expired time.

src/playlist.js
```diff
@@ -305,6 +307,43 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+};
+
+const calculatePlaylistEnd_ = function(playlist, useSafeLiveEnd) {
+  if (!playlist || !playlist.segments) {
+    return null;
+  }
+  if (playlist.endList) {
+    return duration(playlist);
+  }
+  let { expiredSync, segmentSync } = getPlaylistSyncPoints(playlist);
```
mjneil: Since you moved `getPlaylistSyncPoints` to `calculateExpiredTime` calling it again here just to check for null on `expiredSync` `segmentSync` is unnecessary work. If `calculateExpiredTime` starts returning null when it can't calculate expired, you can just check for `null` on `expired` instead of checking the sync points

src/master-playlist-controller.js
```diff
@@ -793,6 +798,46 @@ export class MasterPlaylistController extends videojs.EventTarget {
   }
 
   /**
+   * Check if a playlist has stopped being updated
+   * @param {Object} playlist the media playlist object
+   * @return {boolean} whether the playlist has stopped being updated or not
+   */
+  stopUpdateCheck(playlist) {
```
mjneil: minor, but I think we can come up with a better name for the method. Maybe `isPlaylistOutdated_`, `isPlaylistStagnant_`, `isPlaylistUpdating` or something along these lines that is a bit more descriptive of what the method does. (note the return value should reflect the name logically). This method should also be private, indicated by the `_` appended to the end of the name

src/playlist.js
zhuangs: Did this in the first place but then I found there is a case that the playlist contains the endList, so the seekableEnd will return the duration of playlist but the seekableStart returns null. I think that's why I change here to return 0 and check for null on the sync points instead of checking expired.

mjneil: That is a good point, but I don't think returning 0 is the right call here since there are live scenarios where we may not have any sync points, but there has been expired time, we just can't calculate it. Since `endList` indicates a VOD asset, we actually already have a known sync point then just off knowing `endList` alone, which would be `{ time: 0, mediaSequence: 0 }` since VOD always starts at time 0 segment 0. I'd say it would make most sense to update `getPlaylistSyncPoints` to set `expiredSync` to `{ time: 0, mediaSequence: 0 }` if  `playlist.endList`, otherwise set to `playlist.syncInfo`, and finally resorting to `null`. That way `calculateExpiredTime` can properly calculate for VOD case, while still returning `null` in the live case where we don't have enough information

src/master-playlist-controller.js
zhuangs: fixed this

src/playlist.js
zhuangs: fixed this

src/playlist.js
zhuangs: fixed this

src/master-playlist-controller.js
mjneil: So I was giving this more thought and I think we can avoid worrying about it all together by calling this method and blacklisting the playlist outside of the `loadedplaylist` listener. Playlist Loader already knows when a playlist it has loaded [didn't actually update the live window](https://github.com/videojs/videojs-contrib-hls/blob/master/src/playlist-loader.js#L188). The `isPlaylistOutdated_` method more so checks if the player is stuck at the end despite loading "fresh" playlists. If instead playlist loader triggered an event at that moment it knows the playlist isnt a fresh update, then `MasterPlaylistController` can listen to that and run the checks for playlistOutdated and if those checks pass, then we blacklist. This way we wont run these checks every time the playlist is refreshing correctly and run the risk of accidentally blacklisting a playlist we don't want to

zhuangs: changed as suggested

src/master-playlist-controller.js
```diff
@@ -408,6 +408,15 @@ export class MasterPlaylistController extends videojs.EventTarget {
         bubbles: true
       });
     });
+
+    this.masterPlaylistLoader_.on('playlistnotupdate', () => {
```
mjneil: maybe name the event `playlistunchanged` ?

src/master-playlist-controller.js
```diff
@@ -793,6 +802,46 @@ export class MasterPlaylistController extends videojs.EventTarget {
   }
 
   /**
+   * Check if a playlist has stopped being updated
+   * @param {Object} playlist the media playlist object
+   * @return {boolean} whether the playlist has stopped being updated or not
+   */
+  isPlaylistOutdated_(playlist) {
```
mjneil: I know I've suggested renaming this already, but now that we have an event that signifies the playlist is outdated/not refreshing, and this method more checks if we are stuck at the end of the playlist, so maybe somthing like `stuckAtPlaylistEnd_`or `isAtPlaylistEnd_`. Then our logic sort of semantically reads like if the refreshed playlist is unchanged and the player is at the end of the playlist, blacklist this playlist.

src/master-playlist-controller.js
zhuangs: renamed!

src/master-playlist-controller.js
zhuangs: makes sense! renamed!

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,58 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+    return 170;
+  };
+  this.requests[1].respond(200, null,
+                           '#EXTM3U\n' +
+                           '#EXT-X-MEDIA-SEQUENCE:16\n' +
+                           '#EXTINF:10,\n' +
+                           '16.ts\n');
+  // trigger a refresh
+  this.clock.tick(10 * 1000);
+
```
mjneil: I think it would be good to add a check between the two refreshes to make sure the playlist wasnt blacklisted early. (you may have to move the `play` and `playing` triggers up, not sure though)

src/playlist.js
```diff
@@ -305,6 +305,40 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+    return duration(playlist);
+  }
+  let expired = calculateExpiredTime(playlist);
+  let endSequence;
+
+  if (expired === null) {
+    return null;
+  }
+  endSequence = useSafeLiveEnd ? Math.max(0, playlist.segments.length - Playlist.UNSAFE_LIVE_SEGMENTS) : Math.max(0, playlist.segments.length);
+  let end = intervalDuration(playlist,
```
gesinger: Can just return directly instead of creating a variable for it.

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,64 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  let url;
+
+  this.player.src({
+    src: 'master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+  openMediaSource(this.player, this.clock);
+  this.player.tech_.triggerReady();
+
+  this.requests[0].respond(200, null,
```
gesinger: We might be able to use `standardXHRResponse` to save some lines. We can also use `this.requests.shift` to make it so that the latest outstanding request is always at the beginning of the array (no need to keep track of them via indexing).

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,64 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  this.requests[1].respond(200, null,
+                           '#EXTM3U\n' +
+                           '#EXT-X-MEDIA-SEQUENCE:16\n' +
+                           '#EXTINF:10,\n' +
+                           '16.ts\n');
+
+  url = this.requests[1].url.slice(this.requests[1].url.lastIndexOf('/') + 1);
+  media = this.player.tech_.hls.playlists.master.playlists[url];
+
+  assert.ok(!media.excludeUntil, 'playlist didnt be blacklisted');
```
gesinger: playlist wasn't blacklisted

src/master-playlist-controller.js
```diff
@@ -408,6 +408,19 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    this.masterPlaylistLoader_.on('playlistunchanged', () => {
+      let updatedPlaylist = this.masterPlaylistLoader_.media();
+      let playlistOutdated = this.stuckAtPlaylistEnd_(updatedPlaylist);
+
+      // when the playlist is unchanged and the playhead is stuck at the edge
+      // of playlist, we will blacklist the outdated playlist if it's not the final
+      // rendition to give the player a chance to re-adjust to the proper live point
+      // on a different playlist
+      if (playlistOutdated) {
+        this.blacklistCurrentPlaylist();
```
gesinger: Thinking again about the comment, I think it might make sense to put it in here as something like:

```js
// Playlist has stopped updating and we're stuck at its end. Try to blacklist it and switch to another
// playlist in the hope that that one is updating (and give the player a chance to re-adjust to the
// safe live point).
```

src/master-playlist-controller.js
```diff
@@ -793,6 +806,37 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    }
+
+    // does not use the safe live end to calculate playlist end, since we
+    // don't want to say we are stuck while there is still content
+    let absolutePlaylistEnd = Hls.Playlist.playlistEnd(playlist);
+    let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
+
+    if (!buffered.length) {
+      // playhead reached absolute end of playlist
```
gesinger: My fault, but maybe add a clarification (like you did below in the return comment 👍 ) to something like: `// return true if the playhead reached the absolute end of the playlist`

src/playlist.js
```diff
@@ -305,6 +305,39 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
 
     return segmentSync.time - sumDurations(playlist, syncIndex, 0);
   }
+  return null;
+};
+
+/**
+ * Calculates the playlist end time or seekable end time depends on using
```
gesinger: I think this description can just be left as `Calculates the playlist end time` since the parameter describes its use

src/playlist.js
```diff
@@ -305,6 +305,39 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+ * safe live end or not
+ *
+ * @param {Object} playlist a media playlist object
+ * @param {Object|null} useSafeLiveEnd a boolean value if true then live playlists
```
gesinger: If it is a boolean maybe change the param from `Object|null` to `Boolean|false`

And the description could probably be something like `Whether or not the playlist end calculation should consider the safe live end (truncate the playlist end by three segments). This is normally used for calculating the end of the playlist's seekable range.`

src/playlist.js
```diff
@@ -305,6 +305,39 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+ *                      should not expose three segment durations worth of content
+ *                      from the end of the playlist
+ * @returns {Number} the end time of playlist or the end point of seekable range
```
gesinger: Can remove the "seekable range" part

src/playlist.js
```diff
@@ -305,6 +305,39 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+ * @function calculatePlaylistEnd
```
gesinger: playlistEnd

src/playlist.js
```diff
@@ -305,6 +305,39 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+ */
+export const playlistEnd = function(playlist, useSafeLiveEnd) {
+  if (!playlist || !playlist.segments) {
+    return null;
+  }
+  if (playlist.endList) {
+    return duration(playlist);
+  }
+  let expired = calculateExpiredTime(playlist);
+  let endSequence;
```
gesinger: Can move this down to where it is first set.

test/playlist.test.js
```diff
@@ -271,9 +271,9 @@ QUnit.test('a non-positive length interval has zero duration', function(assert)
   assert.equal(Playlist.duration(playlist, -1), 0, 'negative length duration is zero');
 });
 
-QUnit.module('Playlist Seekable');
+QUnit.module('Playlist Seekable and PlaylistEnd');
```
gesinger: Maybe just "Playlist Seekable and End"

test/playlist.test.js
```diff
@@ -333,8 +352,8 @@ function(assert) {
   assert.equal(seekable.end(0), 7, 'ends three target durations from the last segment');
 });
 
-QUnit.test('seekable end accounts for non-standard target durations', function(assert) {
-  let seekable = Playlist.seekable({
+QUnit.test('seekable end and playlist end accounts for non-standard target durations', function(assert) {
```
gesinger: account

test/playlist.test.js
```diff
@@ -382,16 +404,17 @@ QUnit.test('playlist with no sync points has empty seekable range', function(ass
...
+  };
+  let seekable = Playlist.seekable(playlist);
+  let playlistEnd = Playlist.playlistEnd(playlist);
 
   assert.equal(seekable.length, 0, 'no seekable range for playlist with no sync points');
+  assert.equal(playlistEnd, null, 'no playlist end for playlist with no sync points');
 });
 
-QUnit.test('seekable uses available sync points for calculating seekable range',
+QUnit.test('seekable and playlistEnd uses available sync points for calculating',
```
gesinger: use

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
   assert.equal(this.player.tech_.hls.stats.bandwidth, 1e10, 'bandwidth set above');
 });
 
+QUnit.test('detect if the playlist stuck at the end', function(assert) {
```
gesinger: Probably something like "detects if we are stuck at the playlist end" or "detects if the player is stuck at the playlist end"

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  // not stuck at playlist end when no seekable, even if empty buffer
+  // and positive currentTime
+  hls.masterPlaylistController_.seekable = function() {
+    return videojs.createTimeRange();
+  };
+  this.player.tech_.buffered = function() {
+    return videojs.createTimeRange();
+  };
+  this.player.tech_.setCurrentTime(170);
+  assert.ok(!hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
```
gesinger: Since these are testing MasterPlaylistController's stuckAtPlaylistEnd_ function, they'd probably be better in test/master-playlist-controller.test.js

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
   assert.equal(this.player.tech_.hls.stats.bandwidth, 1e10, 'bandwidth set above');
 });
 
+QUnit.test('detect if the playlist stuck at the end', function(assert) {
+  let playlist;
```
gesinger: Can be moved down to where it is used

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  };
+  this.player.tech_.setCurrentTime(170);
+  assert.ok(!hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
+
+  // not stuck at playlist end when no seekable, even if empty buffer
+  // and currentTime 0
+  this.player.tech_.setCurrentTime(0);
+  assert.ok(!hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
+
+  // not stuck at playlist end when no seekable but more buffer
```
gesinger: maybe  `// not stuck at playlist end when no seekable but current time is at the end of the buffered range`

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  this.player.tech_.buffered = function() {
+    return videojs.createTimeRange(0, 170);
+  };
+  assert.ok(!hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
+
+  // not stuck at playlist end when currentTime not at seekable end
+  // even if the buffer is empty
+  hls.masterPlaylistController_.seekable = function() {
```
gesinger: For all of these really short functions, when we don't need a specific context, it is sometimes nice to use the ES6 syntax: `hls.masterPlaylistController_.seekable = () => videojs.createTimeRange(0, 130);`

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+
+  // stuck at playlist end when there is no buffer and playhead
+  // reached absolute end of playlist
+  this.player.tech_.setCurrentTime(160);
+  Hls.Playlist.playlistEnd = function() {
+    return 160;
+  };
+  assert.ok(hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'stuck at playlist end');
+
+  // stuck at playlist end when current time reach to the buffer end
```
gesinger: current time reached the buffer end

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,121 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  };
+  this.player.tech_.buffered = function() {
+    return videojs.createTimeRange(0, 170);
+  };
+  this.player.tech_.setCurrentTime(170);
+  Hls.Playlist.playlistEnd = function() {
+    return 170;
+  };
+  assert.ok(hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'stuck at playlist end');
+});
```
gesinger: I think there might be one more interesting case, where seekable is a normal range, the buffer reaches the absolute end of the playlist, and current time is in the buffered range (but not yet at the absolute end, say 1 second before it). This case should return that we aren't yet stuck.

src/master-playlist-controller.js
```diff
@@ -408,6 +408,19 @@ export class MasterPlaylistController extends videojs.EventTarget {
         bubbles: true
       });
     });
+
+    this.masterPlaylistLoader_.on('playlistunchanged', () => {
+      let updatedPlaylist = this.masterPlaylistLoader_.media();
+      let playlistOutdated = this.stuckAtPlaylistEnd_(updatedPlaylist);
+
+      // Playlist has stopped updating and we're stuck at its end. Try to
```
gesinger: I'd just move this inside of the conditional (since the phrasing considers that we already checked `playlistOutdated`).

src/playlist.js
```diff
@@ -305,6 +305,38 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
     return segmentSync.time - sumDurations(playlist, syncIndex, 0);
   }
+  return null;
+};
+
+/**
+ * Calculates the playlist end time
+ *
+ * @param {Object} playlist a media playlist object
+ * @param {Boolean|false} useSafeLiveEnd a boolean value indicates whether or not the playlist
```
gesinger: indicating

src/playlist.js
```diff
@@ -305,6 +305,38 @@ const calculateExpiredTime = function(playlist, expiredSync, segmentSync) {
...
+  if (playlist.endList) {
+    return duration(playlist);
+  }
+  let expired = calculateExpiredTime(playlist);
+
+  if (expired === null) {
+    return null;
+  }
+  let endSequence = useSafeLiveEnd ? Math.max(0, playlist.segments.length - Playlist.UNSAFE_LIVE_SEGMENTS) :
+                                 Math.max(0, playlist.segments.length);
```
gesinger: Maybe indent to align?

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,106 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
+  this.player.tech_.buffered = () => videojs.createTimeRange(0, 160);
+  Hls.Playlist.playlistEnd = () => 160;
+  assert.ok(!hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
+
+  // stuck at playlist end when there is no buffer and playhead
+  // reached absolute end of playlist
+  this.player.tech_.setCurrentTime(160);
+  assert.ok(hls.masterPlaylistController_.stuckAtPlaylistEnd_(playlist), 'stuck at playlist end');
+
+  // stuck at playlist end when current time reached to the buffer end
```
gesinger: ~~to~~

test/videojs-contrib-hls.test.js
```diff
@@ -1096,6 +1096,106 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
   assert.equal(this.player.tech_.hls.stats.bandwidth, 1e10, 'bandwidth set above');
 });
 
+QUnit.test('detects if the player is stuck at the playlist end', function(assert) {
```
gesinger: I think this test can be moved to test/master-playlist-controller.test.js

test/playlist.test.js
```diff
@@ -271,9 +271,21 @@ QUnit.test('a non-positive length interval has zero duration', function(assert)
   assert.equal(Playlist.duration(playlist, -1), 0, 'negative length duration is zero');
 });
 
-QUnit.module('Playlist Seekable');
+QUnit.module('Playlist Seekable and End', {
+  beforeEach(assert) {
```
gesinger: Are these `beforeEach` and `afterEach` methods needed anymore?

zhuangs: removed! :)

src/master-playlist-controller.js
```diff
@@ -408,6 +408,19 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    this.masterPlaylistLoader_.on('playlistunchanged', () => {
+      let updatedPlaylist = this.masterPlaylistLoader_.media();
+      let playlistOutdated = this.stuckAtPlaylistEnd_(updatedPlaylist);
+
+      if (playlistOutdated) {
+        // Playlist has stopped updating and we're stuck at its end. Try to
+        // blacklist it and switch to another playlist in the hope that that
+        // one is updating (and give the player a chance to re-adjust to the
+        // safe live point).
+        this.blacklistCurrentPlaylist();
```
gesinger: Would be worthwhile to trigger an event just after with something like: 

```js
// useful for monitoring QoS
this.tech_.trigger('playliststuck');
```

gesinger: We can also pass some error info into `blacklistCurrentPlaylist` regarding the error itself, so we can log appropriately.

e.g.,
```js
this.blacklistCurrentPlaylist({
  message: 'Playlist no longer updating.'
});
```

Then in `blacklistCurrentPlaylist` add a check for an error message and include it in the middle of logs (where it makes sense).

src/master-playlist-controller.js
```diff
@@ -408,6 +408,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      let updatedPlaylist = this.masterPlaylistLoader_.media();
+      let playlistOutdated = this.stuckAtPlaylistEnd_(updatedPlaylist);
+
+      if (playlistOutdated) {
+        // Playlist has stopped updating and we're stuck at its end. Try to
+        // blacklist it and switch to another playlist in the hope that that
+        // one is updating (and give the player a chance to re-adjust to the
+        // safe live point).
+        this.blacklistCurrentPlaylist({
+          message: 'Playlsit no longer updating.'
```
gesinger: Playlist

src/master-playlist-controller.js
```diff
@@ -824,8 +870,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     nextPlaylist = this.selectPlaylist();
 
     if (nextPlaylist) {
-      videojs.log.warn('Problem encountered with the current ' +
-                       'HLS playlist. Switching to another playlist.');
+      let logMessage = error.message ? 'Problem encountered with the current HLS playlist. ' + error.message + ' Switching to another playlist.' :
```
gesinger: I think this may be cleaner as:

```js
videojs.log.warn('Problem encountered with the current HLS playlist.' +
                 (error.message ? ' ' + error.message : '') +
                 ' Switching to another playlist.');
```

src/master-playlist-controller.js
```diff
@@ -408,6 +408,21 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      let playlistOutdated = this.stuckAtPlaylistEnd_(updatedPlaylist);
+
+      if (playlistOutdated) {
+        // Playlist has stopped updating and we're stuck at its end. Try to
+        // blacklist it and switch to another playlist in the hope that that
+        // one is updating (and give the player a chance to re-adjust to the
+        // safe live point).
+        this.blacklistCurrentPlaylist({
+          message: 'Playlsit no longer updating.'
+        });
```
gesinger: We should also trigger an event with:

```js
// useful for monitoring QoS
this.tech_.trigger('playliststuck');
```

src/master-playlist-controller.js
```diff
@@ -824,8 +872,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     nextPlaylist = this.selectPlaylist();
 
     if (nextPlaylist) {
-      videojs.log.warn('Problem encountered with the current ' +
-                       'HLS playlist. Switching to another playlist.');
+      videojs.log.warn('Problem encountered with the current HLS playlist.' +
+                 (error.message ? ' ' + error.message : '') +
```
gesinger: Indentation

zhuangs: changed! sorry for keeping making this mistake

test/master-playlist-controller.test.js
```diff
@@ -368,6 +368,61 @@ QUnit.test('updates the enabled track when switching audio groups', function(ass
...
+  this.player.tech_.buffered = () => videojs.createTimeRange(0, 170);
+  assert.ok(!this.masterPlaylistController.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
+
+  // not stuck at playlist end when currentTime not at seekable end
+  // even if the buffer is empty
+  this.masterPlaylistController.seekable = () => videojs.createTimeRange(0, 130);
+  this.player.tech_.setCurrentTime(50);
+  this.player.tech_.buffered = () => videojs.createTimeRange();
+  Hls.Playlist.playlistEnd = () => 130;
+  assert.ok(!this.masterPlaylistController.stuckAtPlaylistEnd_(playlist), 'not stuck at playlist end');
```
mjneil: Since `stuckAtPlaylistEnd_` only uses seekable to check if we have a seekable range, but doesn't use `seekable.end(0)` for any of its checks, is there even a point to checking `not stuck at playlist end when currentTime not at seekable end even if the buffer is empty`?

If we still want to keep this part in, I'm not sure having `playlistEnd` and `seekable.end(0)` both be the same (130) is representative of an actual scenario since `seekable.end` is the time since playback started to minus 3 segment durations from the end of the window, where `playlistEnd` is time since playback started to minus 0 segment durations from the end of the window.

gesinger: Good point. I think that leaving in tests that make use of seekable values is good, even if `stuckAtPlaylistEnd_` doesn't use them, since the tests shouldn't know too much about the internal workings of the function. But you're right that the scenario of `playlistEnd` equaling `seekable`'s end shouldn't be a common case. However, in the event that a playlist changes from LIVE to VOD (at the end of a stream), and the only update was the addition of `#EXT-X-ENDLIST`, `seekable`'s end should equal `playlistEnd`, and I think it's possible that this case is a valid one.

src/playlist.js
```diff
@@ -235,9 +235,7 @@ const getPlaylistSyncPoints = function(playlist) {
   if (!playlist || !playlist.segments) {
     return [null, null];
   }
-
-  let expiredSync = playlist.syncInfo || null;
-
+  let expiredSync = playlist.endList ? { time: 0, mediaSequence: 0} : playlist.syncInfo || null;
```
mjneil: When a live stream is finished, usually an `EXT-X-ENDLIST` tag is applied to the end of the playlist even though it wasn't originally a VOD. This would set `expiredSync` to `{ time: 0, mediaSequence: 0 }` even if that isn't an actual sync point. We should prioritize `playlist.syncInfo` if it exists, otherwise `{ time: 0, mediaSequence: 0 }` if `playlist.endList` is true, otherwise `null`.

PR #1042, opened 2017-03-07
Fix #1030

### LINKED ISSUE #1030: A momentary loss of connection while downloading chunks causes an exception.
## Description
While streaming an HLS if connectivity is lost even for a moment the player throws an exception and locks up. A couple of retries with exponential backoff would be ideal. We are seeing lots of small drops in connection during playback. 

## Sources
Can be triggered on any HLS video.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.

1. Begin streaming any HLS.
2. Disable network connection.
3. Re-enable network connection.

### Expected
I would expect this work in the same fashion that a progressive mp4 download works. I can kill a connection with a progressive download for a few seconds and re-enable it without issue.

### Error output
`{code: 2, message: "A network error caused the media download to fail part-way."}`

### videojs-contrib-hls version
what version of videojs-contrib-hls does this occur with?
videojs-contrib-hls 5.2.1

### videojs version
what version of videojs does this occur with?
video.js 5.16.0

### Browsers
what browsers are affected? please include browser and version for each
* Chrome

### Platforms
what platforms are affected? please include operating system and version or device and version for each
* Windows 7

leonardo-hotmart:
## Description
If the connection is lost, videojs-contrib-hls will try to load the media every five seconds until the connection is back.

## Specific Changes proposed
Change the error handling of mainSegmentLoader.
Case the error is a connection lost error it will try to load the segment again from five to five seconds.
Case the error is any other else it will blacklist the playlist 

## Requirements Checklist
- [X] Bug fixed

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -426,7 +426,24 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('error', () => {
-      this.blacklistCurrentPlaylist(this.mainSegmentLoader_.error());
+      var segmentLoader = this;
+      var error = this.mainSegmentLoader_.error();
+      if(error.code == 2){//If lost connection, try again later
+          //pause the media
+          segmentLoader.mainSegmentLoader_.pause();
+          if (segmentLoader.audioPlaylistLoader_) {
+              segmentLoader.audioSegmentLoader_.pause();
+          }
+          //setTimeout to try again in 5 seconds
+          setTimeout(function(){
+              segmentLoader.mainSegmentLoader_.load();
+              if (segmentLoader.audioPlaylistLoader_) {
+                  segmentLoader.audioSegmentLoader_.load();
+              }
+          }, 5000);
+      }else{
+          segmentLoader.blacklistCurrentPlaylist(error);
+      }
     });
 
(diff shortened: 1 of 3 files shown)


leonardo-hotmart:
I can't see why the build is failing.
I've ran the tests locally and everything went fine.
Would it be the chrome version? Mine is 55

image

gesinger:
Hey @leonardo-hotmart , thanks for the contribution!

This is an interesting problem, and you're right (as well as the posts in #1030 ), in that we shouldn't blacklist the playlist for a momentary loss of connectivity. This PR definitely can solve it, but it might introduce some other scenarios where the behavior is not quite optimal. For instance, if the connection keeps dropping for the same playlist, we may want to try a different one. Or if someone tries to pause the video during the 5 second delay, then it'll automatically resume after those 5 seconds (though that can be resolved with more state management).

It's also hard to detect the reason for the error. While an XMLHttpRequest status of 0 can be because of a momentary connectivity loss, it can also be because of CORS, firewalls, or another reason. In these cases we do want to blacklist and try another playlist.

It may be worth it for us to open up the discussion in the issue to see what people think would be the most optimal approach for error handling in these cases, but I'm interested to hear your thoughts.

leonardo-hotmart:
Hi @gesinger. Thanks for your answer. I'll give my thoughts on #1030 Ok?
Cheers

PR #1044, opened 2017-03-08
Make SegmentLoader great again

imbcmdth:
## Description
Segment XHR-related state is strewn across large parts of `SegmentLoader`. The actual problem of fetching a segment and one or two other requests that the segment might depend on is fairly self-contained. This PR aims to refactor the xhr code almost completely out of `SegmentLoader` so that it can focus on deciding what to get, not how to get them.

## Specific Changes proposed

1. All code related to xhr is completely removed from `SegmentLoader` into `mediaSegmentRequest` - now the only thing it has to do is handle a nicely packaged response that has either failed or is ready for transmuxing
2. All decryption code has also moved to `mediaSegmentRequest` including the callback management that was in `MasterPlaylistController`
   * The only thing MPC does with the decrypter is life-cycle management
3. Added better tracking for `mediaRequests*` stats
   * `mediaRequests` now properly tracks the count of ALL media requests even failed or aborted requests
   * `mediaRequestsErrored` has the number of requests that resulted in a failure (40x, 50x, etc.)
   * `mediaRequestsTimedout` has the number of requests that timed-out
   * `mediaRequestsAborted` has the number of requests explicitly aborted by the `SegmentLoader`
4. Added `mediaSegmentRequest` unit-tests

## Requirements Checklist

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -122,7 +122,7 @@ const handleErrors = (error, request) => {
     };
   }
 
-  if (!response) {
+  if (request.aborted || !response || response.byteLength === 0) {
     return {
       status: request.status,
       message: 'HLS request aborted at URL: ' + request.uri,
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -242,7 +242,7 @@ const handleProgress = (segment, callback) => (event) => {
  *
  * @private
  */
-export const segmentRequest = (xhr, xhrOptions, decryptionWorker, segment, progressFn, nextFn) => {
+export const mediaSegmentRequest = (xhr, xhrOptions, decryptionWorker, segment, progressFn, nextFn) => {
   const activeXhrs = {};
   const abortAll = aborter(activeXhrs);
   const processSegmentCallback = waitForCompletion(activeXhrs,
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -34,7 +34,7 @@ const segmentXhrHeaders = function(segment) {
   return headers;
 };
 
-const aborter = (activeXhrs) => () => {
+const abortAll = (activeXhrs) => {
   Object.keys(activeXhrs).forEach((xhrKey) => {
     const xhr = activeXhrs[xhrKey];
 
@@ -72,48 +72,49 @@ const getProgressStats = (progressEvent) => {
   return stats;
 };
 
-const handleErrors = (callback) => (error, request) => {
+const handleErrors = (error, request) => {
   const response = request.response;
 
   if (!request.aborted && error) {
-    return callback({
+    return {
       status: request.status,
       message: 'HLS request errored at URL: ' + request.uri,
       code: REQUEST_ERRORS.FAILURE,
       xhr: request
-    }, request);
+    };
   }
 
   if (request.timedout) {
-    return callback({
+    return {
       status: request.status,
       message: 'HLS request timed-out at URL: ' + request.uri,
       code: REQUEST_ERRORS.TIMEOUT,
       xhr: request
-    }, request);
+    };
   }
 
   if (!response) {
-    return callback({
+    return {
       status: request.status,
       message: 'HLS request aborted at URL: ' + request.uri,
       code: REQUEST_ERRORS.ABORTED,
       xhr: request
-    }, request);
+    };
   }
 
-  return callback(error, request);
+  return null;
 };
 
-const handleKeyResponse = (segment, callback) => (error, request) => {
+const handleKeyResponse = (segment, finishProcessingFn) => (error, request) => {
   const response = request.response;
+  const errorObj = handleErrors(error, request);
 
-  if (error) {
-    return callback(error, segment);
+  if (errorObj) {
+    return finishProcessingFn(errorObj, segment);
   }
 
   if (response.byteLength !== 16) {
-    return callback({
+    return finishProcessingFn({
       status: request.status,
       message: 'Invalid HLS key at URL: ' + request.uri,
       code: REQUEST_ERRORS.FAILURE,
@@ -129,13 +130,16 @@ const handleKeyResponse = (segment, callback) => (error, request) => {
     view.getUint32(8),
     view.getUint32(12)
   ]);
-  return callback(null, segment);
+  return finishProcessingFn(null, segment);
 };
 
-const handleInitSegmentResponse = (segment, callback) => (error, request) => {
-  if (error) {
-    return callback(error, segment);
+const handleInitSegmentResponse = (segment, finishProcessingFn) => (error, request) => {
+  const errorObj = handleErrors(error, request);
+
+  if (errorObj) {
+    return finishProcessingFn(errorObj, segment);
   }
+
   segment.map.bytes = new Uint8Array(request.response);
 
   if (!Array.isArray(segment.bandwidth)) {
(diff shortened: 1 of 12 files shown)


src/media-segment-request.js
```diff
@@ -0,0 +1,281 @@
...
+  const roundTripTime = Date.now() - request.requestTime;
+  const stats = {
+    bandwidth: Infinity,
+    bytesReceived: 0,
+    roundTripTime: Math.max(roundTripTime, 1)
+  };
+
+  if (progressEvent.lengthComputable) {
+    stats.bytesReceived = progressEvent.loaded;
+    stats.bandwidth = (stats.bytesReceived / roundTripTime) * 8 * 1000;
```
mjneil: I think this should use `stats.roundTripTime` or use `Math.max` when initializing `roundTripTime` to avoid division by 0 (I assume the use of `Math.max` above was to not report a round trip of 0)

src/media-segment-request.js
```diff
@@ -0,0 +1,281 @@
...
+  let count = 0;
+
+  return (error, segment) => {
+    // errors have to be unshifted to make sure the original error - the one
+    // that resulted in several aborts - ends up at the start of the array for
+    // ease of use downstream
+    if (error) {
+      // If there are errors, we have to abort any outstanding requests
+      abortAll(activeXhrs);
+      errors.unshift(error);
```
mjneil: Wouldn't unshifting cause the original error that caused all the aborts to be at the end of the array? You abort all the other requests, unshift the error, and then when the aborted requests get handled, they would have an error here and unshift themselves in front of the original?

src/media-segment-request.js
```diff
@@ -0,0 +1,281 @@
...
+  }
+  return headers;
+};
+
+const abortAll = (activeXhrs) => {
+  Object.keys(activeXhrs).forEach((xhrKey) => {
+    const xhr = activeXhrs[xhrKey];
+
+    // only abort xhrs that haven't had a response
+    if (!xhr.responseTime) {
```
mjneil: Does it hurt to call `abort` on a request that has already been aborted? If so should probably add a check for that here as well. If one of the requests causes an error, the rest of the requests are aborted which in turn would cause an error on the aborted request, which would loop through all the requests again, even the aborted one

src/media-segment-request.js
```diff
@@ -0,0 +1,281 @@
...
+const handleInitSegmentResponse = (segment, finishProcessingFn) => (error, request) => {
+  const errorObj = handleErrors(error, request);
+
+  if (errorObj) {
+    return finishProcessingFn(errorObj, segment);
+  }
+
+  segment.map.bytes = new Uint8Array(request.response);
+
+  if (!Array.isArray(segment.bandwidth)) {
```
mjneil: This bandwidth array isn't used anywhere. What is it for?

imbcmdth: Removed...

src/media-segment-request.js
```diff
@@ -72,6 +91,13 @@ const getProgressStats = (progressEvent) => {
   return stats;
 };
 
+/**
+ * Handle all error conditions in one place and return an object
+ * with all the information
+ *
+ * @param {Anything} error - if non-null signals an error occured with the XHR
```
mjneil: I think `{Object|null}` or `{Error|null}` would be acceptable here

src/media-segment-request.js
```diff
@@ -220,27 +272,80 @@ const waitForCompletion = (activeXhrs, decrypter, callback) => {
...
-      return maybeDecrypt(decrypter, segment, callback);
+      // Otherwise, everything is ready just continue
+      return doneFn(null, segment);
     }
   };
 };
 
-const handleProgress = (segment, callback) => (event) => {
+/**
+ * Simple progress event callback handler that gathers some states before
```
mjneil: gathers some ~~states~~ stats

src/media-segment-request.js
imbcmdth: Done

src/media-segment-request.js
imbcmdth: I haven't seen any issues. Aborting an already aborted xhr seems to be a no-op and doesn't trigger any sort of readystatechange.

src/media-segment-request.js
imbcmdth: You might be right. It looks like I was seeing sinon's mock xhr results (which are synchronous) and was confusing it with actual behavior.

src/media-segment-request.js
imbcmdth: Done

src/media-segment-request.js
imbcmdth: Fixed

src/segment-loader.js
```diff
@@ -805,185 +720,157 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      return;
     }
 
-    if (!this.xhr_.segmentXhr && !this.xhr_.keyXhr && !this.xhr_.initSegmentXhr) {
-      this.xhr_ = null;
-      this.processResponse_();
+    // the request was aborted and the SegmentLoader has already started
+    // another request. this can happen when the timeout for an aborted
+    // request triggers due to a limitation in the XHR library
+    // do not count this as any sort of request or we risk double-counting
```
mjneil: This is a bit confusing. Can you explain the limitation? (Not requesting you add more detail to the actual code comment)

mjneil:
I'm not sure what detail I'm missing here since Travis obviously seems to think the tests should pass, but I notice that all the `mediaSegmentRequest` unit tests have `assert.expect(n)` but only have `n - 2` `assert`s, which would lead me to believe that the tests should fail.

mjneil:
I think it would be good to have a test that makes sure `doneFn` isn't called until segment, key, and init segment requests have responded

test/segment-loader.test.js
```diff
@@ -249,7 +249,6 @@ QUnit.test('calculates bandwidth after downloading a segment', function(assert)
   // verify stats
   assert.equal(loader.mediaBytesTransferred, 10, '10 bytes');
   assert.equal(loader.mediaTransferDuration, 100, '100 ms (clock above)');
-  assert.equal(loader.mediaRequests, 1, '1 request');
```
mjneil: why remove this?

imbcmdth: Because we have this same test dozens of times and it's very redundant. I actually tried to remove more of them but lost hope.

src/segment-loader.js
```diff
@@ -669,6 +644,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     let segment = playlist.segments[mediaIndex];
 
     return {
+      requestId: 'segment-loader-' + Math.random(),
```
mjneil: I think it would be better to do `Math.random().toString(36)` like you do with `decryptionId` in `mediaSegmentRequest`.

src/media-segment-request.js
```diff
@@ -0,0 +1,403 @@
...
+
+/**
+ * Decrypt the segment via the decryption web worker
+ *
+ * @param {WebWorker} decrypter - a WebWorker interface to AES-128 decryption routines
+ * @param {Object} segment - a simplified copy of the segmentInfo object from SegmentLoader
+ * @param {Function} doneFn - a callback that is executed after decryption has completed
+ */
+const decryptSegment = (decrypter, segment, doneFn) => {
+  const decryptionId = 'segment-request-' + Math.random().toString(36);
```
mjneil: Why not use `segment.requestId` here instead of generating another random string?

PR #1046, opened 2017-03-08
Fix fast quality change for alternate audio

gesinger:


src/master-playlist-controller.js
```diff
@@ -698,7 +698,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.masterPlaylistLoader_.media(media);
 
       this.mainSegmentLoader_.resetLoader();
-      if (this.audiosegmentloader_) {
```
mjneil: not only was this not checking for the playlist loader, it's not cased properly for the segment loader lol

src/master-playlist-controller.js
```diff
@@ -698,7 +698,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      if (this.audioPlaylistLoader_) {
```
mjneil: Why should this be `audioPlaylistLoader_` vs `audioSegmentLoader_`?

gesinger: audioSegmentLoader_ always exists, even if not being used. audioPlaylistLoader_ only exists if we are using alt audio

mjneil:
Is it even necessary to reset the audio loader on fast quality change? If `fastQualityChange_` causes a rendition switch, the `mediachange` event listener will call `setUpAudio` if the active audio group has changed. `setUpAudio` will tear down the playlist loader and reset the segment loader

imbcmdth:
We decided to remove the code if it is redundant and to add a test to prove that it was not necessary.

PR #1043, opened 2017-03-07
In-Manifest WebVTT Support (WIP)

mjneil:
## Description
This adds support for in-manifest VTT

## Specific Changes proposed
* uses `PlaylistLoader` to continuously refresh subtitle playlist in live playback
* Adds `VTTSegmentLoader` for loading vtt segments
* updates m3u8-parser to v2.1.0
* uses vtt.js to parse X-TIMESTAMP-MAP header from vtt segments (Requires [gkatsev/vtt.js#4](https://github.com/gkatsev/vtt.js/pull/4))
* supports using an init segment via `EXT-X-MAP` tag in the playlist

## Requirements Checklist
- [ ] Feature implemented / Bug fixed
- [ ] Unit Tests updated or fixed
- [ ] Docs/guides updated
- [ ] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))
- [ ] Reviewed by Two Core Contributors

PR #1057, opened 2017-03-14
In-Manifest WebVTT Support

mjneil:
## Description
This adds support for in-manifest VTT

## Specific Changes proposed
* uses `PlaylistLoader` to continuously refresh subtitle playlist in live playback
* Adds `VTTSegmentLoader` for loading vtt segments
* updates m3u8-parser to v2.1.0
* uses vtt.js to parse X-TIMESTAMP-MAP header from vtt segments (Requires [gkatsev/vtt.js#4](https://github.com/gkatsev/vtt.js/pull/4))
* supports using an init segment via `EXT-X-MAP` tag in the playlist

## Requirements Checklist
- [ ] Docs/guides updated


src/master-playlist-controller.js
```diff
@@ -272,17 +274,22 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
         this.audioSegmentLoader_.handleDecrypted_(event.data);
+      } else if (event.data.source === 'vtt') {
+        this.subtitleSegmentLoader_.handleDecrypted_(event.data);
       }
     };
 
     this.setupSegmentLoaderListeners_();
 
-    this.masterPlaylistLoader_.start();
+    this.masterPlaylistLoader_.load();
```
mjneil: From #1043 
@imbcmdth: Just curious: why this was renamed?

src/master-playlist-controller.js
```diff
@@ -599,13 +612,55 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+  /**
    * Call load on our SegmentLoaders
    */
   load() {
     this.mainSegmentLoader_.load();
     if (this.audioPlaylistLoader_) {
       this.audioSegmentLoader_.load();
     }
+    if (this.subtitlePlaylistLoader_) {
+      this.subtitleSegmentLoader_.load();
```
mjneil: from #1043 
@imbcmdth: At some point I feel like it might just be nice to have an array of active segment loaders. Then we can just do this.segmentLoaders_.forEach((l) => l.load());... not necessary now but maybe a good idea to think about soon.

src/master-playlist-controller.js
```diff
@@ -624,6 +679,50 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+    this.subtitleSegmentLoader_.abort();
+
+    let track = this.activeSubtitleTrack_();
+
+    if (track) {
+      track.mode = 'disabled';
+    }
+
+    this.setupSubtitles();
```
mjneil: from #1043 
@imbcmdth: If `handleSubtitleError_` calls `setupSubtitles` is there a possibility that an infinite loop of failure will result?

src/master-playlist-controller.js
```diff
@@ -708,6 +807,91 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+        // easiest to simply recreate the playlist loader
+        !this.subtitlePlaylistLoader_.media() ||
+        this.subtitlePlaylistLoader_.media().resolvedUri !== properties.resolvedUri) {
+
+      if (this.subtitlePlaylistLoader_) {
+        this.subtitlePlaylistLoader_.dispose();
+      }
+
+      // reset the segment loader
+      this.subtitleSegmentLoader_.resetEverything();
```
mjneil: from #1043 
@imbcmdth: Why not move this to where the `SegmentLoader` is paused above. It's easier to reason about the state of the loader if it is manipulated in one place. Also, shouldn't it be reset even if there is no track?

src/vtt-segment-loader.js
```diff
@@ -235,6 +203,25 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+
+    const cues = this.subtitlesTrack_.cues;
+    let start = cues[0].startTime;
+    let end = cues[cues.length - 1].startTime;
+
+    return videojs.createTimeRanges([[start, end]]);
+  }
+
+  timestampOffset() {
+    return this.syncController_.timestampOffsetForTimeline(this.currentTimeline_);
```
mjneil: from #1043 
@imbcmdth: Is this function used anywhere? If so, it needs a comment.

src/vtt-segment-loader.js
```diff
@@ -1006,44 +972,166 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      // of the file
+      const initId = initSegmentId(segment.map);
+      const initSegment = this.initSegments_[initId];
+      const vttLineTerminators = new Uint8Array('\n\n'.split('').map(char => char.charCodeAt(0)));
+      const combinedByteLength = vttLineTerminators.byteLength + initSegment.bytes.byteLength;
+      const combinedSegment = new Uint8Array(combinedByteLength);
+
+      combinedSegment.set(initSegment.bytes);
+      combinedSegment.set(vttLineTerminators, initSegment.bytes.byteLength);
+      segment.map.bytes = combinedSegment;
```
mjneil: from #1043
@imbcmdth: Couple of thoughts on this whole block:

1. We should pre-calculate the `vttLineTerminators` value because it's a typed array that never changes.
2. I think we should do the entire append process when we first get the response for the `initSegment` so that we the object stored in `this.initSegments_[initId]` already has the value that would be in `combinedSegment`.

src/vtt-segment-loader.js
```diff
@@ -1006,44 +972,166 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    }
+
+    try {
+      this.parseVTTCues_(segmentInfo);
+    } catch (e) {
+      this.error({
+        message: e.message
```
mjneil: from #1043
@imbcmdth: If error is fatal it should have some code set in addition to the message. The code is important for reporting reasons.

src/vtt-segment-loader.js
```diff
@@ -1006,44 +972,166 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      cue.startTime += diff;
+      cue.endTime += diff;
+    });
+
+    const firstStart = segmentInfo.cues[0].startTime;
+    const lastStart = segmentInfo.cues[segmentInfo.cues.length - 1].startTime;
+    const midPoint = (firstStart + lastStart) / 2;
+
+    segment.start = midPoint - (segment.duration / 2);
+    segment.end = midPoint + (segment.duration / 2);
```
mjneil: from #1043
@imbcmdth: Everything about this feels wrong and dirty! 😱

src/master-playlist-controller.js
mjneil: @gesinger and I noticed that `load` calls `start` if start hasn't been called yet and is safe to call multiple times, but `start` should only be called once. Since this call is in the `MasterPlaylistController` constructor, theres no risk of running into that scenario, but we decided to just change any calls to `start` into `load` so we didn't have to concern ourselves with it. We can change it back if you think `start` makes more sense semantically

src/master-playlist-controller.js
mjneil: It shouldn't since just above we set the current active tracks `mode` to `'disabled'`. This has the same effect as turning off subtitles through the UI.

src/master-playlist-controller.js
mjneil: `setupSubtitles` gets called everytime the player's `TextTrackList` fires a `change` event. When changing subtitle tracks through the player ui, depending on which track was selected and which track you are changing to, this event will fire multiple times. (it also fires a change event if you select the already active track in the UI) Only one of these calls will actually reset the playlist loader and segment loader to start loading from a new track. By only resetting when we are actually changing tracks, we avoid reseting the segment loader after its already started from the previous call, or if the user selected the already active track, then all that happens is the segment loader is paused and unpaused. Also, yes it should be reset even if there is no track, but since we always reset before starting a new track, it will get reset when the user turns subtitles back on.

src/vtt-segment-loader.js
mjneil: Agree with both, good ideas

src/master-playlist-controller.js
imbcmdth: Sounds good to me. 👍

src/vtt-segment-loader.js
imbcmdth: We decided not to do this now as we plan on revisiting how and when we emit errors at some point in the near future.

src/vtt-segment-loader.js
imbcmdth: We decided to try not storing segment start and end times and letting the conservative guess make it work since web vtt segments should be much smaller than video or audio segments.

One thing to be careful about is that the "resync" functionality needs to be removed since it will continuously reset the segment loader on poor guess instead of just walking forward.

src/vtt-segment-loader.js
imbcmdth: Do it! 😃

src/master-playlist-controller.js
imbcmdth: I feel like a short version of this explanation needs to be in this function somewhere!

src/vtt-segment-loader.js
imbcmdth: Well, is it? 😆

utils/stats/index.html
```diff
@@ -56,7 +56,7 @@
     <br>
     <label>
       Player Options:
-      <input id="player-options" type="text" value='{}'>
+      <input id="player-options" type="text" value='{"vtt.js": "PATH/TO/vtt.js"}'>
```
mjneil: reminder to self to revert back before releasing

src/vtt-segment-loader.js
mjneil: the resync functionality actually checks for the existence of `segment.end` so we are safe leaving it here if we never set it

src/vtt-segment-loader.js
```diff
@@ -506,34 +476,30 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          // if not paused, queue a buffer check as soon as possible
+          this.monitorBuffer_();
+        }
+      };
 
-    if (segmentInfo.mediaIndex === this.playlist_.segments.length - 1 &&
-        this.mediaSource_.readyState === 'ended' &&
-        !this.seeking_()) {
+      this.syncController_.on('timestampoffset', checkTimestampOffset);
+      this.state = 'WAITING_ON_TIMELINE';
```
imbcmdth: What happens if we switch playlists *or seek* while the state is 'WAITING_ON_TIMELINE'?

src/vtt-segment-loader.js
```diff
@@ -1006,44 +972,166 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    if (typeof window.WebVTT !== 'function' &&
+        this.subtitlesTrack_ &&
+        this.subtitlesTrack_.tech_) {
+
+      const loadHandler = () => {
+        this.subtitlesTrack_.tech_.off('vttjsloaded', loadHandler);
+        this.handleSegment_();
+      };
+
+      this.state = 'WAITING_ON_VTTJS';
```
imbcmdth: What happens if we switch playlists *or seek* while the state is 'WAITING_ON_VTTJS'?

src/vtt-segment-loader.js
mjneil: For the vtt-loader, no it isnt

src/vtt-segment-loader.js
```diff
@@ -1006,44 +972,166 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      this.subtitlesTrack_.tech_.on('vttjsloaded', loadHandler);
```
imbcmdth: This seems like the perfect case to use `this.subtitlesTrack_.tech_.one('vttjsloaded', ...)`

src/vtt-segment-loader.js
```diff
@@ -506,34 +476,30 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        if (!this.paused()) {
+          // if not paused, queue a buffer check as soon as possible
+          this.monitorBuffer_();
+        }
+      };
 
-    if (segmentInfo.mediaIndex === this.playlist_.segments.length - 1 &&
-        this.mediaSource_.readyState === 'ended' &&
-        !this.seeking_()) {
+      this.syncController_.on('timestampoffset', checkTimestampOffset);
```
imbcmdth: This seems like the perfect case to use `this.syncController_.one('timestampoffset', ...)`

test/segment-loader.test.js
```diff
@@ -403,7 +403,7 @@ QUnit.test('only appends one segment at a time', function(assert) {
   assert.equal(loader.mediaRequests, 1, '1 request');
 });
 
-QUnit.test('adjusts the playlist offset if no buffering progress is made', function(assert) {
+QUnit.skip('adjusts the playlist offset if no buffering progress is made', function(assert) {
```
imbcmdth: Why was this test skipped?

mjneil: This test is actually outdated from before simple-fetcher times when there was the offset correction variable to handle incorrect guesses, (now we just walk forward), it just wasnt skipped/removed previously because it happened to pass after segment loader changes. If you notice the other playlist offset tests are also skipped. I was planning on cleaning up skipped tests when the loaders get refactored

src/vtt-segment-loader.js
mjneil: While the state is not `WAITING` any aborts will just clear pending segment (which at this point there isn't one yet) and does nothing else. The callback will set the state to ready and call `monitorBuffer_` to restart the lifecycle. Since it starts back from the beginning, any seeks or subtitle changes that happened in the waiting time shouldn't have adverse effects

src/vtt-segment-loader.js
mjneil: This is basically the same as aborting while waiting in the `DECRYPTING` stage, the callback will call `handleSegment_` again and if it was aborted while we were waiting, the first `if` block will reset the loader to `READY`

src/vtt-segment-loader.js
mjneil: I think it will be better to do these changes when we integrate `mediaSegmentRequest` since that changes when the maps are stored in `initSegments_`

PR #1060, opened 2017-03-22
Network Simulator Enhancements

imbcmdth:
Fixed the network simulator since it was out of date and didn't work with the changes that had been added over the last year. In addition, added the ability to provide real network traces to simulate real-world conditions. Added some configuration options so that you can specify the  "throughput" value to simulate how slow MSE or Flash backends effect ABR behavior.

Created a very barebones node.js server to simulate ABR bandwidth from the same network traces so that segments can be served to a real player with the same real-world conditions used by the simulator. 

Network traces:
http://skuld.cs.umass.edu/traces/mmsys/2013/pathbandwidth/


imbcmdth:
I will add and attribution for the dataset and paper before merging.

Issue #575, opened 2016-03-01
InvalidStateError calling endOfStream for broken source urls in chrome


t-n:
When using a video source that is returning a 404 or connection refused I get no error event but instead get an uncaught InvalidStateError.

```
Uncaught InvalidStateError: Failed to execute 'endOfStream' on 'MediaSource': The MediaSource's readyState is not 'open'.
videojs.HlsHandler.blacklistCurrentPlaylist_ @ videojs-hls.js:1129
(anonymous function) @ videojs-hls.js:189
trigger @ stream.js:60
(anonymous function) @ playlist-loader.js:326
(anonymous function) @ xhr.js:42
(anonymous function) @ index.js:8
loadFunc @ index.js:145
readystatechange @ index.js:75
```

Steps to reproduce:
1. Put videojs-contrib-hls(1.3.9) on a web server.
2. Open example.html in Chrome(48).
3. Change the video url to a vaild non-existing url, e.g. changing just the domain name to the same domain name/ip and port as the webserver. (with valid CORS headers)

Expected result:
- An error event (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) with the corresponding message in the player.

Actual result:
- Uncaught InvalidStateError: Failed to execute 'endOfStream' on 'MediaSource': The MediaSource's readyState is not 'open'.
  (If you try to load the video url a second time the correct event is triggered if it's a 404 but not if it's a connection refused)

sashee:
This just bit me too. Could you find a workaround?

t-n:
No, not yet.

ZacharySato:
Having the same issue. 
I try to bind player rebuilding on 'error' event, but in this case 'error' isn't actually triggered. 
Strange thing is that 'InvalidStateError' appears randomly across correct 'CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED'.

mrmckeb:
We're also seeing this issue intermittently, but it doesn't appear to be causing any actual problems in the video streams.

askaliuk:
I confirm this issue

gesinger:
I see a lot of people have seen this issue as well. I tried the listed steps with the latest player, but was unable to reproduce the problem. Would someone who is seeing the issue be able to provide a reduced test case using [this template](http://jsbin.com/vokipos/edit?html,output)?

Thank you all for reporting the issue.

aquu:
[...]

PR #1061, opened 2017-03-24
Trigger player error on initial request error instead of media source error

gesinger:
## Description
See https://github.com/videojs/videojs-contrib-hls/issues/575
This is to resolve an issue where we try to trigger an error on a media source that isn't yet ready.

## Specific Changes proposed
Since we still want to trigger an error on the player, and can't use the media source (since it isn't ready yet), this fires the error directly on the player.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -872,7 +872,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // trying to load the master OR while we were disposing of the tech
     if (!currentPlaylist) {
       this.error = error;
-      return this.mediaSource.endOfStream('network');
+
+      try {
+        return this.mediaSource.endOfStream('network');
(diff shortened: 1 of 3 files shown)


src/master-playlist-controller.js
```diff
@@ -817,7 +817,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // trying to load the master OR while we were disposing of the tech
     if (!currentPlaylist) {
       this.error = error;
-      return this.mediaSource.endOfStream('network');
+      return this.trigger('error');
```
mjneil: What about putting the `endOfStream` call in a try catch and if it fails trigger the error on MPC? I feel like we probably want to still call `endOfStream` in the common case of not hitting an `InvalidStateError`

PR #1065, opened 2017-03-27
Minor loader fixes

imbcmdth:
## Description
This PR has two minor fixes that were discovered after testing some new tools:

* Fix for request timeouts being interpreted as errors
   * The tests not catching this problem because Sinon’s FakeXHR differs from real XHRs the way aborts are handled
* Fix to ensure `MediaSource#endOfStream` is still called when the final segment request causes a rendition switch

## Specific Changes proposed
The `xhrFactory` used by videojs-contrib-hls now ensures that the `aborted` property is set on aborted requests so that other parts of the project don't have to worry about it (and simplify the `handleErrors` function in media-segment-request).

In `handleUpdateEnd` we always use the `segmentInfo` object's `mediaIndex` so that it doesn't matter if the `SegmentLoader` is reset before we perform the check for end-of-stream condition.

Fixed, Sinon's  FakeXHR's `abort` function so that it doesn't call any error handlers when abort is called on the request (this was fixed by later versions of Sinon but we can't use them for other reasons).

## Requirements Checklist
- [x] Bug fixed

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -40,19 +40,13 @@ const segmentXhrHeaders = function(segment) {
 };
 
 /**
- * Abort all requests that have not already returned data
+ * Abort all requests
  *
  * @param {Object} activeXhrs - an object that tracks all XHR requests
  */
 const abortAll = (activeXhrs) => {
   activeXhrs.forEach((xhr) => {
-    // only abort xhrs that haven't had a response
-    if (!xhr.responseTime) {
-      // set an aborted property so that we can correctly
-      // track the request and not treat it as an error
-      xhr.aborted = true;
-      xhr.abort();
-    }
+    xhr.abort();
   });
 };
 
@@ -102,17 +96,6 @@ const getProgressStats = (progressEvent) => {
  * @param {Object} request -  the XHR request that possibly generated the error
  */
 const handleErrors = (error, request) => {
-  const response = request.response;
-
-  if (!request.aborted && error) {
-    return {
-      status: request.status,
-      message: 'HLS request errored at URL: ' + request.uri,
-      code: REQUEST_ERRORS.FAILURE,
-      xhr: request
-    };
-  }
-
   if (request.timedout) {
     return {
       status: request.status,
@@ -122,11 +105,20 @@ const handleErrors = (error, request) => {
     };
   }
 
-  if (request.aborted || !response || response.byteLength === 0) {
+  if (request.aborted) {
     return {
       status: request.status,
       message: 'HLS request aborted at URL: ' + request.uri,
       code: REQUEST_ERRORS.ABORTED,
+      xhr: request
+    };
+  }
+
+  if (error) {
+    return {
+      status: request.status,
+      message: 'HLS request errored at URL: ' + request.uri,
+      code: REQUEST_ERRORS.FAILURE,
       xhr: request
     };
   }
--- a/test/media-segment-request.test.js
+++ b/test/media-segment-request.test.js
@@ -177,60 +177,6 @@ QUnit.test('cancels outstanding key requests on timeout', function(assert) {
   this.clock.tick(2000);
 });
 
-QUnit.test('ensure the aborted property is set on aborted requests', function(assert) {
-  let keyReq;
-  const done = assert.async();
-  let oldXHRPrototype = XMLHttpRequest.prototype;
-
-  assert.expect(7);
-  // Sinon automatically adds the aborted property so let's stop it from
-  // doing that to make sure we add it under the expected condition when
-  // running on a real, native XHR object
-  XMLHttpRequest.prototype = Object.create(XMLHttpRequest.prototype);
-  XMLHttpRequest.prototype.abort = function abort() {
-    this.response = this.responseText = '';
-    this.errorFlag = true;
-    this.requestHeaders = {};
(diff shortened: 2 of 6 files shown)


src/media-segment-request.js
```diff
@@ -102,31 +96,29 @@ const getProgressStats = (progressEvent) => {
  * @param {Object} request -  the XHR request that possibly generated the error
  */
 const handleErrors = (error, request) => {
-  const response = request.response;
-
-  if (!request.aborted && error) {
+  if (request.timedout) {
```
mjneil: Can you go into detail of what the issue was and why changing the order of checking for timeout then errors solves it?

test/media-segment-request.test.js
```diff
@@ -173,11 +173,64 @@ QUnit.test('cancels outstanding key requests on timeout', function(assert) {
...
+QUnit.test('ensure the aborted property is set on aborted requests', function(assert) {
+  let keyReq;
+  const done = assert.async();
+  let oldXHRPrototype = XMLHttpRequest.prototype;
+
+  assert.expect(7);
+  // Sinon automatically adds the aborted property so let's stop it from
+  // doing that to make sure we add it under the expected condition when
+  // running on a real, native XHR object
+  XMLHttpRequest.prototype = Object.create(XMLHttpRequest.prototype);
```
mjneil: Should this be moved to `test-helpers.js` so all our tests involving XHR act more like real, native XHR?

test/test-helpers.js
```diff
@@ -168,6 +168,26 @@ export const useFakeEnvironment = function(assert) {
   });
   fakeEnvironment.clock = sinon.useFakeTimers();
   fakeEnvironment.xhr = sinon.useFakeXMLHttpRequest();
+
+  // Sinon 1.10.2 handles abort incorrectly (triggering the error event)
+  // Later versions fixed this but broke the ability to set the response
+  // to an arbitrary object (in our case, a typed array).
+  XMLHttpRequest.prototype = Object.create(XMLHttpRequest.prototype);
```
mjneil: Shouldn't this be `fakeEnvironment.xhr.prototype`?

test/media-segment-request.test.js
imbcmdth: There is a small (1-line) difference between them.

Basically, the one in `test-helpers.js` removes a line that triggered the `error` callback (since that is clearly wrong) but kept everything else the same.

This one additionally removes the setting of the `aborted` property to true since the goal of the test is to make sure that happens. Native browsers do not set that property but removing it from all FakeXHR instances would break other tests (namely, playlist loading tests) that rely on the property's presence to know if a request was aborted during the test.

mjneil: Now that `xhrFactory` adds the aborted property to aborted requests, shouldn't these tests still pass since there will be an aborted property?

src/media-segment-request.js
imbcmdth: The `error` parameter can be set for a variety of reasons. The only time we should consider it an actual error is when neither `timedout` nor `aborted` are set.

The issue was that a `timeout` was being interpreted as an error because the `error` parameter was set and the request wasn't aborted.

In the unit tests, Sinon's FakeXHR *was* setting the `aborted` property on timeouts avoiding this condition but the browser's native XHR does not set that property.

test/test-helpers.js
imbcmdth: Sinon globally replaces `XMLHttpRequest` constructor. The two are equivalent.

test/media-segment-request.test.js
imbcmdth: That is exactly the point - to make sure we are adding the property ourselves even when Sinon isn't (like in the native browser).

mjneil: Right, thats the point of this test. I mean that with the combination of `xhrFactory` also setting `aborted` to `true` and the reordering of checking for timeouts and aborts before errors sounds like all the other tests that would break by the removal of the line that sets `aborted` to `true` in sinons FakeXHR should not break, and if they are, I feel those tests should be updated to check for the correct error condition

test/test-helpers.js
```diff
@@ -168,6 +168,26 @@ export const useFakeEnvironment = function(assert) {
...
+  XMLHttpRequest.prototype.abort = function abort() {
+    this.aborted = true;
+    this.response = this.responseText = '';
+    this.errorFlag = true;
+    this.requestHeaders = {};
```
gesinger: Should this check for current request headers?

test/test-helpers.js
```diff
@@ -168,6 +168,26 @@ export const useFakeEnvironment = function(assert) {
...
+    this.responseHeaders = {};
+
+    if (this.readyState > 0 && this.sendFlag) {
+      this.readyStateChange(4);
```
gesinger: With XMLHttpRequest aborts, it looks like the `readystatechange` event is not fired (source: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort ).

test/test-helpers.js
imbcmdth: All this code comes from Sinon - the only thing I changed was to remove the dispatching of an `error` event at the end of the function.

test/test-helpers.js
imbcmdth: See above.

PR #1067, opened 2017-03-30
fix segment time mapping for fmp4 playback

mjneil:
## Description
Segment start and end times were being miscalculated for fragmented mp4s when `timestampOffset` was being set.

## Specific Changes proposed
Use `segmentInfo.startOfSegment` for setting the time mapping instead of `segmentInfo.timestampOffset`. These values are the same except in the fmp4 case  when `timestampOffset` is adjusted to account for offset start times since fmp4 playback uses native source buffers instead of our `VirtualSourceBuffer` and transmuxer for setting `baseMediaDecodeTime` properly.

## Requirements Checklist

### DIFF
--- a/src/sync-controller.js
+++ b/src/sync-controller.js
@@ -362,13 +362,13 @@ export default class SyncController extends videojs.EventTarget {
       this.logger_('tsO:', segmentInfo.timestampOffset);
 
       mappingObj = {
-        time: segmentInfo.timestampOffset,
-        mapping: segmentInfo.timestampOffset - timingInfo.start
+        time: segmentInfo.startOfSegment,
+        mapping: segmentInfo.startOfSegment - timingInfo.start
       };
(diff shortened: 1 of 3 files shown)


src/sync-controller.js
```diff
@@ -355,12 +355,12 @@ export default class SyncController extends videojs.EventTarget {
       this.logger_('tsO:', segmentInfo.timestampOffset);
 
       mappingObj = {
-        time: segmentInfo.timestampOffset,
-        mapping: segmentInfo.timestampOffset - timingInfo.start
+        time: segmentInfo.startOfSegment,
+        mapping: segmentInfo.startOfSegment - timingInfo.start
```
imbcmdth: I think to ensure we don't introduce a regression in the future, we need a unit test around modification of `timestampOffset` (which we do in mp4Probe) to make sure it doesn't change the start and end time of the segment.

PR #1072, opened 2017-04-05
Trigger error events when received empty response

stevennick:
Trigger error events when received empty response. For example, receive contents from outdated streaming server through CDN.

## Description
This is an segment loader enhancement for better response handles.
In some use cases(especially for LIVE streaming, plus through private/public CDN network), the streaming source may throw empty segments, cause XHR response 200 status code but actually get empty content. Trying to parse those problem segments will cause player programmatic errors and makes it unrecoverable. This enhancement detects those error and throws exception, make player able to handle those problems.

## Specific Changes proposed
After segment loader receive responses, detect `response.byteLength` not at zero size before any processing tasks.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -151,6 +151,16 @@ const handleKeyResponse = (segment, finishProcessingFn) => (error, request) => {
     }, segment);
   }
 
+  // stop processing if received empty content
+  if(response.byteLength === 0) {
+    return finishProcessingFn({
+      status: request.status,
+      message: 'Empty HLS content at URL: ' + request.uri,
+      code: REQUEST_ERRORS.FAILURE,
+      xhr: request
+    }, segment);
+  }
+
   const view = new DataView(response);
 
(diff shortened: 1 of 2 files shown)


src/media-segment-request.js
```diff
@@ -170,11 +170,23 @@ const handleKeyResponse = (segment, finishProcessingFn) => (error, request) => {
...
 const handleInitSegmentResponse = (segment, finishProcessingFn) => (error, request) => {
+  const response = request.response;
   const errorObj = handleErrors(error, request);
 
   if (errorObj) {
     return finishProcessingFn(errorObj, segment);
   }
+
+  // stop processing if received empty content
+  if (response.byteLength === 0) {
```
mjneil: Is there any reason we wouldn't want to return this error for key requests? I think it would make sense to put this check at the end of `handleErrors` and just return the `errorObj` instead of making the check in both segment and init responses.

PR #1075, opened 2017-04-05
update readme: when using beforeRequest, source should be added at runtime

mattwaggs:
## Description
Specifically in Edge, the `beforeRequest` function doesn't consistently get called if the video source was set with a source tag.  It seems to be because Edge starts loading the video segments before videojs initializes. This is a problem because if your using `beforeRequest` for anything like attaching authorization headers the video player will go bang.

You can see an example here: http://jsbin.com/cijiyidego/1/edit?html,console,output

Wrapping the video player initialization in a timeout exaggerates the issue. If you watch the network inspector in Edge segments start downloading before the player initializes.  Removing the source tag and using `player.src` when the player is ready fixes the issue.

## Specific Changes proposed
This PR just updates the readme so people don't get hung up on this as long as I did.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,9 @@ player.hls.xhr.beforeRequest = function(options) {
 
 The global `videojs.Hls` also exposes an `xhr` property. Specifying a
 `beforeRequest` function on that will allow you to intercept the options
-for *all* requests in every player on a page.
+for *all* requests in every player on a page. For consistency across
+browsers the video source should be set at runtime once the video player
+is ready. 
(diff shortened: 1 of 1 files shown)


mjneil:
Thanks for finding this issue and adding it to our docs. I think it would be good to move the code workaround example and add details about the issue in Edge to the `Known Issues` section and to make note in the `beforeRequest` section that there is more information in the known issues

Issue #1074, opened 2017-04-05
Trigger an event when the final playlist would've been blacklisted


gesinger:
## Description
Right now, with the latest changes to the player, we don't blacklist the final rendition, meaning that the player will keep trying that stream. This is normally fine, as the stream should be able to play within a player without trouble, and normally when errors happen it is due to the server being down temporarily, rather than a complete block due to permissions (in which case the stream isn't usually meant for that domain/user to play). However, there could be various reasons for the event, and it would be nice to allow developers to react outside of the player to such an error condition.

zhuangs:
#1080

mjneil:
#1080 has been merged into master

PR #1080, opened 2017-04-11
Trigger an event when the final playlist would have been blacklisted

zhuangs:
## Description
Trigger an event when the final playlist would've been blacklisted
Might be a solution for #1074


test/videojs-contrib-hls.test.js
```diff
@@ -1135,6 +1143,7 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
   assert.equal(this.env.log.warn.args[1],
               'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
               'log specific error message for final playlist');
+  assert.equal(retryplaylist, 1, 'retried final playlist for once');
```
mjneil: I think you should add a check here to make sure `blacklistplaylist` is still `1`

PR #1083, opened 2017-04-11
Clear the blacklist for other playlists if final rendition errors

zhuangs:
## Description
Before the behavior is we never blacklist the final request and wait up to 5 minutes for other playlists' blacklist duration to be cleared. Change the behavior as follow:
- If there is only one rendition and it errors, never blacklist it
- if there are multiple renditions and we are down to the last one (the previous renditions are all blacklisted) and it errors, we clear the blacklist duration for the other playlists, and retry the best available playlist after a delay.


src/master-playlist-controller.js
```diff
@@ -1064,14 +1064,22 @@ export class MasterPlaylistController extends videojs.EventTarget {
     }
 
     let isFinalRendition = this.masterPlaylistLoader_.isFinalRendition_();
+    let playlists = this.tech_.hls.playlists.master.playlists;
```
mjneil: This can just be `this.masterPlaylistLoader_.master.playlists`

src/playlist-loader.js
```diff
@@ -295,6 +295,15 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
+    window.clearTimeout(mediaUpdateTimeout);
+
+    if (isFinalRendition) {
+      let refreshDelay = (playlist.targetDuration / 2) * 1000 || 5 * 1000;
+
+      mediaUpdateTimeout = window.setTimeout(loader.media.bind(loader, playlist, false), refreshDelay);
+      return;
+    }
+
     // setter
```
gesinger: This should probably be moved up

test/videojs-contrib-hls.test.js
```diff
@@ -1129,22 +1129,21 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
-  // media wasn't blacklisted because it's final rendition
-  assert.ok(!media.excludeUntil, 'media not blacklisted after playlist 404');
+  assert.ok(media.excludeUntil, 'second media was blacklisted after playlist 404');
   assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
   assert.equal(this.env.log.warn.args[1],
-              'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
-              'log specific error message for final playlist');
-  assert.equal(retryplaylist, 1, 'retried final playlist for once');
-  assert.equal(blacklistplaylist, 1, 'there is one blacklisted playlist');
+              'Problem encountered with the current HLS playlist. HLS playlist request error at URL: media1.m3u8 Switching to another playlist.',
```
gesinger: Long line

gesinger: I also just realized that there is no period between `media1.m3u8 Switching` (from before, but might be nice to fix.

test/videojs-contrib-hls.test.js
```diff
@@ -1153,9 +1152,11 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
   this.clock.tick(3 * 1000);
   // continue loading the final remaining playlist after it wasn't blacklisted
```
gesinger: Technically, I think this comment might need to change

test/videojs-contrib-hls.test.js
```diff
@@ -1531,13 +1532,14 @@ QUnit.test('playlist blacklisting duration is set through options', function(ass
   assert.equal(this.env.log.warn.args[0],
               'Problem encountered with the current HLS playlist. HLS playlist request error at URL: media.m3u8 Switching to another playlist.',
               'log generic error message');
+  // this takes one millisecond
```
gesinger: Why did this new response need to be added?

src/master-playlist-controller.js
```diff
@@ -1064,15 +1064,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+                       'HLS playlist. Trying again since it is the only playlist.');
 
       this.tech_.trigger('retryplaylist');
       return this.masterPlaylistLoader_.load(isFinalRendition);
     }
+    if (isFinalRendition) {
+      playlists.forEach((playlist) => {
+        // clear the blacklist duration for the other playlists when
+        // the final playlist errors
+        playlist.excludeUntil = Date.now();
```
gesinger: It might be clearer to `delete playlist.excludeUntil` for each.

test/videojs-contrib-hls.test.js
zhuangs: If the new response didn't be added here, it will error for the second(final) rendition. Since we clear blacklist duration for the first rendition when final rendition errors, we can't test if the blacklist duration is set through option successfully.

test/videojs-contrib-hls.test.js
```diff
@@ -1529,15 +1532,16 @@ QUnit.test('playlist blacklisting duration is set through options', function(ass
   assert.ok(media.excludeUntil > 0, 'original media blacklisted for some time');
   assert.equal(this.env.log.warn.calls, 1, 'warning logged for blacklist');
   assert.equal(this.env.log.warn.args[0],
-              'Problem encountered with the current HLS playlist. HLS playlist request error at URL: media.m3u8 Switching to another playlist.',
+              'Problem encountered with the current HLS playlist. HLS playlist request error at URL: media.m3u8. Switching to another playlist.',
```
gesinger: Long line.

test/videojs-contrib-hls.test.js
```diff
@@ -1127,35 +1127,38 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
   assert.equal(this.env.log.warn.args[1],
-              'Problem encountered with the current HLS playlist. Trying again since it is the final playlist.',
-              'log specific error message for final playlist');
-  assert.equal(retryplaylist, 1, 'retried final playlist for once');
-  assert.equal(blacklistplaylist, 1, 'there is one blacklisted playlist');
+              'Problem encountered with the current HLS playlist. HLS playlist request error at URL: media1.m3u8. ' +
+              'Switching to another playlist.',
+              'log generic error message');
+  assert.equal(retryplaylist, 0, 'haven\'t retried any playlist');
+  assert.equal(blacklistplaylist, 2, 'media1 is blacklisted');
```
gesinger: Although this may be true, since it's only testing that it got a blacklist event, maybe it is worth having two different checks, one for getting the blacklist event, the other checking for the excludeUntil

test/videojs-contrib-hls.test.js
```diff
@@ -1127,35 +1127,38 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
 
   this.clock.tick(2 * 1000);
   // no new request was made since it hasn't been half the segment duration
   assert.strictEqual(3, this.requests.length, 'no new request was made');
 
   this.clock.tick(3 * 1000);
-  // continue loading the final remaining playlist after it wasn't blacklisted
+  // loading the first playlist since the blacklist duration is cleared
```
gesinger: was

test/videojs-contrib-hls.test.js
```diff
@@ -1127,35 +1127,38 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
   // when half the segment duaration passed
+
   assert.strictEqual(4, this.requests.length, 'one more request was made');
+  // the first media was unblacklisted after a refresh delay
```
gesinger: Would be good to check excludeUntil was deleted

test/videojs-contrib-hls.test.js
zhuangs: I think we have that excludeUntil check in line 1141?

src/master-playlist-controller.js
```diff
@@ -1064,26 +1064,34 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       videojs.log.warn('Problem encountered with the current ' +
-                       'HLS playlist. Trying again since it is the final playlist.');
+                       'HLS playlist. Trying again since it is the only playlist.');
 
       this.tech_.trigger('retryplaylist');
       return this.masterPlaylistLoader_.load(isFinalRendition);
     }
+    if (isFinalRendition) {
+      playlists.forEach((playlist) => {
+        // clear the blacklist duration for the other playlists when
```
gesinger: Rather than describing what the code is doing, it might make more sense to have a comment about the reason we are doing it. Something like `Since we're on the final non-blacklisted playlist, and we're about to blacklist it, instead of erring the player or retrying this playlist, we clear out the current blacklist so the other playlists may be attempted in case any have been fixed.`

test/videojs-contrib-hls.test.js
```diff
@@ -1127,35 +1127,42 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
...
-  // continue loading the final remaining playlist after it wasn't blacklisted
+  // loading the first playlist since the blacklist duration was cleared
   // when half the segment duaration passed
+
   assert.strictEqual(4, this.requests.length, 'one more request was made');
+  url = this.requests[3].url.slice(this.requests[3].url.lastIndexOf('/') + 1);
+  media = this.player.tech_.hls.playlists.master.playlists[url];
+
+  // the first media was unblacklisted after a refresh delay
+  assert.ok(!media.excludeUntil, 'excludeUntil was deleted after cleared the blacklist duration for the first media');
```
gesinger: Can just be `removed first media from blacklist`

src/master-playlist-controller.js
```diff
@@ -1064,26 +1064,34 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    if (playlists.length === 1) {
+      // Never blacklisting this playlist because it's the only playlist
       videojs.log.warn('Problem encountered with the current ' +
-                       'HLS playlist. Trying again since it is the final playlist.');
+                       'HLS playlist. Trying again since it is the only playlist.');
 
       this.tech_.trigger('retryplaylist');
       return this.masterPlaylistLoader_.load(isFinalRendition);
     }
+    if (isFinalRendition) {
```
gesinger: Might be worth adding a warning log here with something like `Clearing blacklist because last rendition is about to be blacklisted.`

src/master-playlist-controller.js
```diff
@@ -1064,26 +1064,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      playlists.forEach((playlist) => {
+        // Since we're on the final non-blacklisted playlist, and we're about to blacklist it,
+        // instead of erring the player or retrying this playlist, we clear out the current blacklist
+        // so the other playlists may be attempted in case any have been fixed.
+        videojs.log.warn('Clearing blacklist because last rendition is about to be blacklisted.');
```
gesinger: This will get logged for every playlist. May just want to log it once.

src/master-playlist-controller.js
```diff
@@ -1064,26 +1064,36 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       videojs.log.warn('Problem encountered with the current ' +
-                       'HLS playlist. Trying again since it is the final playlist.');
+                       'HLS playlist. Trying again since it is the only playlist.');
 
       this.tech_.trigger('retryplaylist');
       return this.masterPlaylistLoader_.load(isFinalRendition);
     }
+    if (isFinalRendition) {
+      playlists.forEach((playlist) => {
+        // Since we're on the final non-blacklisted playlist, and we're about to blacklist it,
```
gesinger: We probably want this comment outside of the loop as well.

src/master-playlist-controller.js
```diff
@@ -1064,26 +1064,37 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       return this.masterPlaylistLoader_.load(isFinalRendition);
     }
+    if (isFinalRendition) {
+      // Since we're on the final non-blacklisted playlist, and we're about to blacklist it,
+      // instead of erring the player or retrying this playlist, we clear out the current blacklist
+      // so the other playlists may be attempted in case any have been fixed.
+      playlists.forEach((playlist) => {
+        delete playlist.excludeUntil;
+      });
+      videojs.log.warn('Clearing blacklist for every playlist because last rendition ' +
```
gesinger: Minor, but putting it just before the loop might be better for a couple reasons: (1) if something breaks (throws an exception) during the loop, we know the last warning log and it can help us determine what broke and (2) we already cleared everything, so `Clearing` might be more accurate before

zhuangs: Makes sense! Changed as suggested

gesinger (approved):

AndyBarron:
Any updates on this PR? The blacklisting is actively affecting a project I'm working on, would love to see this fixed :)

gesinger (approved):

src/playlist-loader.js
```diff
@@ -319,6 +319,16 @@ export default class PlaylistLoader extends EventTarget {
       throw new Error('Cannot switch media playlist from ' + this.state);
     }
 
+    window.clearTimeout(this.mediaUpdateTimeout);
+
+    if (isFinalRendition) {
+      const delay = (playlist.targetDuration / 2) * 1000 || 5 * 1000;
```
mjneil: I don't think we will encounter this because of when this is called with `isFinalRendition === true`, `playlist` is an object, but we support `playlist` being a string as you can see just below this, so it may be better to move this block below the `string` check

gesinger: 👍 updated

mjneil (approved):

forbesjo:
Closing and opening against VHS in the future

PR #1089, opened 2017-04-18
Fix SourceUpdater updating to be true during removes

mjneil:
Because Firefox changes the `updating` property on source buffers and then queues the `updateend` event to fire at a later time, the `SourceUpdater` may think the source buffer is not updating and overwrite the pending callback to be a different one. A similar issue was fixed previously by adding a check for `SourceUpdater.pendingCallback_` before considering `updating` to be `false`. However, `SourceUpdater.remove` did not add a `done` callback to run after the remove, so `updating` was not properly considered `true`. 

This change adds a no-op function as the `done` callback for removes. It also does not queue any remove actions until after the first call to `appendBuffer`, otherwise the `SourceUpdater` will be waiting for an `updateend` event from a source buffer that does not exist before completing the remove action.

## Requirements Checklist

### DIFF
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -45,7 +45,7 @@ export default class SourceUpdater {
     this.pendingCallback_ = null;
     this.timestampOffset_ = 0;
     this.mediaSource = mediaSource;
-    this.firstAppend_ = false;
+    this.processedAppend_ = false;
 
     if (mediaSource.readyState === 'closed') {
       mediaSource.addEventListener('sourceopen', createSourceBuffer);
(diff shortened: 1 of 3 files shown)


zhuangs:
Tested and LGTM

src/source-updater.js
```diff
@@ -43,6 +45,7 @@ export default class SourceUpdater {
     this.pendingCallback_ = null;
     this.timestampOffset_ = 0;
     this.mediaSource = mediaSource;
+    this.firstAppend_ = false;
```
imbcmdth: I feel this name is a bit ambiguous. Not sure if it's worth fixing but something like `processedAppend_` might be more clear.

imbcmdth:
Minor nitpick .. LGMT otherwise!

PR #1097, opened 2017-04-24
use specified mediasequence for VOD expired sync instead of assuming 0

mjneil:
## Description
Seekable calculations assumed that VOD playlists have an `EXT-X-MEDIA-SEQUENCE` of `0`, however this is not always the case. ~~This fixes the calculations to use the media sequence specified in the playlist~~ This has been updated to now use the `SyncController` to pick a sync point and calculate expired time

### DIFF
--- a/test/sync-controller.test.js
+++ b/test/sync-controller.test.js
@@ -397,4 +397,72 @@ QUnit.test('Correctly calculates expired time', function(assert) {
   expired = this.syncController.getExpiredTime(playlist, Infinity);
 
   assert.equal(expired, 100.8, 'estimated expired time using segmentSync');
+
+  playlist = {
+    targetDuration: 10,
+    discontinuityStarts: [],
+    mediaSequence: 100,
+    endList: true,
+    segments: [
+      {
+        duration: 10,
+        uri: '0.ts'
+      },
+      {
+        duration: 10,
+        uri: '1.ts'
+      },
+      {
+        duration: 10,
+        uri: '2.ts'
+      },
+      {
+        duration: 10,
+        uri: '3.ts'
+      },
+      {
+        duration: 10,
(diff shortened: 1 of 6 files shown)


gesinger:
A test we should check for: if a live stream ends, and we switch to a playlist that we haven't seen before (don't have `syncInfo`), or, we tune into a live stream towards its end and rendition switch, then it is possible this may change that behavior.

mjneil:
That scenario is definitely a problem, but it is not a very common case. I think it will be ok to move forward with this and make an issue for that case to address in another PR

test/sync-controller.test.js
```diff
@@ -249,3 +249,151 @@ QUnit.test('Correctly updates time mapping and discontinuity info when probing s
...
+      {
+        duration: 10,
+        uri: '4.ts'
+      }
+    ]
+  };
+
+  let expired = this.syncController.getExpiredTime(playlist, Infinity);
+
+  assert.equal(expired, 100, 'seekable range calculated');
```
gesinger: Assertion would probably be clearer using your below format. Something like `estimated expired time using sync info`

test/sync-controller.test.js
```diff
@@ -249,3 +249,151 @@ QUnit.test('Correctly updates time mapping and discontinuity info when probing s
...
+      {
+        duration: 10,
+        uri: '4.ts'
+      }
+    ]
+  };
+
+  expired = this.syncController.getExpiredTime(playlist, Infinity);
+
+  assert.equal(expired, 98.5, 'estimated start time using segmentSync');
```
gesinger: Probably better to use `estimated expired time...` instead of start for these and others

zhuangs:
Tested and LGTM

PR #1100, opened 2017-04-28
Fix playback stalls when everything appears okay

mjneil:
## Description
Sometimes things go wrong when things look right. This change adds a last ditch effort correction in `PlaybackWatcher` to seek to `currentTime` if a playback stall occurs and there is at least 1 second of forward buffer.

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -218,8 +218,11 @@ export default class PlaybackWatcher {
 
     // Sometimes the player can stall for unknown reasons within a contiguous buffered
     // region with no indication that anything is amiss (seen in Firefox). Seeking to
-    // currentTime is usually enough to kickstart the player.
-    if (currentRange.length) {
+    // currentTime is usually enough to kickstart the player. This checks that the player
+    // is currently within a buffered region and there is at least half a second
+    // of forward buffer so that this isn't triggered when the player is just buffering
+    // due to slow connection.
+    if (currentRange.length && currentTime <= currentRange.end(0) - 0.5) {
       this.cancelTimer_();
       this.tech_.setCurrentTime(currentTime);
 
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -211,6 +211,19 @@ export default class PlaybackWatcher {
       this.timer_ = setTimeout(this.skipTheGap_.bind(this),
                                difference * 1000,
                                currentTime);
+      return;
+    }
+
+    // Sometimes the player can stall for unknown reasons within a contiguious buffered
+    // region with no indication that anything is amiss (Seen in Firefox). Seeking to
+    // currentTime is usually enough to kickstart the player
+    if (buffered.length && currentTime + 1 < buffered.end(buffered.length - 1)) {
+      this.cancelTimer_();
+      this.tech_.setCurrentTime(currentTime);
+
+      // unknown waiting corrections may be useful for monitoring QoS
+      this.tech_.trigger('unknownwaiting');
+      return;
     }
   }
 
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -101,6 +101,16 @@ export default class PlaybackWatcher {
     }
 
     let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
+
+    if (buffered.length && currentTime + 0.1 >= buffered.end(buffered.length - 1)) {
+      // If current time is at the end of the final buffered region, then any playback
+      // stall is most likely caused by buffering in a low bandwidth environment. This
+      // prevents playback watcher from updating consecutive updates until the player
+      // has some forward buffer to avoid miscategorizing waiting on a slow connection
+      // as a playback issue.
+      return;
+    }
 
     if (this.consecutiveUpdates >= 5 &&
         currentTime === this.lastRecordedTime) {
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -186,13 +186,15 @@ export default class PlaybackWatcher {
     // Sometimes the player can stall for unknown reasons within a contiguous buffered
     // region with no indication that anything is amiss (seen in Firefox). Seeking to
     // currentTime is usually enough to kickstart the player. This checks that the player
-    // is currently within a buffered region and there is at least half a second
-    // of forward buffer so that this isn't triggered when the player is just buffering
-    // due to slow connection.
+    // is currently within a buffered region before attempting a corrective seek.
+    // Chrome does not appear to continue `timeupdate` events after a `waiting` event
+    // until there is ~ 3 seconds of forward buffer available. PlaybackWatcher should also
+    // make sure there is ~3 seconds of forward buffer before taking any corrective action
+    // to avoid triggering an `unknownwaiting` event in when the network is slow.
     // Note: This is not done when the `waiting` event fired by the tech because the tech
     // also fires `waiting` when the player is buffering in low bandwidth scenarios, which
     // requires no action from playback watcher.
-    if (currentRange.length && currentTime <= currentRange.end(0) - 0.5) {
+    if (currentRange.length && currentTime + 3 <= currentRange.end(0)) {
       this.cancelTimer_();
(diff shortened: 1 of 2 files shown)


src/playback-watcher.js
```diff
@@ -211,6 +211,19 @@ export default class PlaybackWatcher {
       this.timer_ = setTimeout(this.skipTheGap_.bind(this),
                                difference * 1000,
                                currentTime);
+      return;
+    }
+
+    // Sometimes the player can stall for unknown reasons within a contiguious buffered
+    // region with no indication that anything is amiss (Seen in Firefox). Seeking to
```
gesinger: seen

src/playback-watcher.js
```diff
@@ -211,6 +211,19 @@ export default class PlaybackWatcher {
...
+    // currentTime is usually enough to kickstart the player
```
gesinger: .

src/playback-watcher.js
```diff
@@ -211,6 +211,19 @@ export default class PlaybackWatcher {
       this.timer_ = setTimeout(this.skipTheGap_.bind(this),
                                difference * 1000,
                                currentTime);
+      return;
+    }
+
+    // Sometimes the player can stall for unknown reasons within a contiguious buffered
```
gesinger: contiguous

src/playback-watcher.js
```diff
@@ -211,6 +211,19 @@ export default class PlaybackWatcher {
...
+    // region with no indication that anything is amiss (Seen in Firefox). Seeking to
+    // currentTime is usually enough to kickstart the player
+    if (buffered.length && currentTime + 1 < buffered.end(buffered.length - 1)) {
```
gesinger: With really short segments it's possible this can break. We may want to use a function that checks to see if we're within a buffered region (and if we want to be careful around edges, we can use a margin of some partial second, as 1 may be too long).

src/playback-watcher.js
```diff
@@ -211,6 +211,19 @@ export default class PlaybackWatcher {
...
+      this.cancelTimer_();
+      this.tech_.setCurrentTime(currentTime);
+
+      // unknown waiting corrections may be useful for monitoring QoS
```
gesinger: We should log here.

src/playback-watcher.js
```diff
@@ -211,6 +211,24 @@ export default class PlaybackWatcher {
...
                                currentTime);
+      return;
+    }
+
+    let currentRange = Ranges.findRange(buffered, currentTime);
+
+    // Sometimes the player can stall for unknown reasons within a contiguous buffered
+    // region with no indication that anything is amiss (seen in Firefox). Seeking to
+    // currentTime is usually enough to kickstart the player.
+    if (currentRange.length) {
```
gesinger: Do we want to guard against cases where we're at the end of the buffered region and just waiting for data?

mjneil: That was why I originally had the 1 second safety net. The `Ranges.findRange` function uses `TIME_FUDGE_FACTOR` which I thought was more inline with your previous comment. I'll make sure to test this exact scenario and see if it triggers the event, and if so, we can discuss an appropriate fudge factor to protect against this

gesinger: Although shorter segments aren't all too common, the original one could break because it wasn't checking to see if the current time was within a buffered range. This one does check that the current time is within a buffered range. The TIME_FUDGE_FACTOR may be enough of a safety margin from the edge of a buffered region, but it could be worth being a bit safer with how far it is from the edge of its containing region.

src/playback-watcher.js
```diff
@@ -161,15 +161,55 @@ export default class PlaybackWatcher {
...
+        `playback by seeking to the current time.`);
+
+      // unknown waiting corrections may be useful for monitoring QoS
+      this.tech_.trigger('unknownwaiting');
+      return;
+    }
+  }
+
+  /**
+   * Handler for situations when the tech fires a `waiting` event
```
gesinger: Would be good to add a short note about how we can't predict when these events will occur for the browser.

src/playback-watcher.js
```diff
@@ -156,20 +167,63 @@ export default class PlaybackWatcher {
...
+    // Sometimes the player can stall for unknown reasons within a contiguous buffered
+    // region with no indication that anything is amiss (seen in Firefox). Seeking to
+    // currentTime is usually enough to kickstart the player. This checks that the player
+    // is currently within a buffered region and there is at least half a second
+    // of forward buffer so that this isn't triggered when the player is just buffering
+    // due to slow connection.
+    // Note: This is not done when the `waiting` event fired by the tech because the tech
+    // also fires `waiting` when the player is buffering in low bandwidth scenarios, which
+    // requires no action from playback watcher.
+    if (currentRange.length && currentTime <= currentRange.end(0) - 0.5) {
```
mjneil: I'm not sure how necessary the 0.5 safety offset is now that in `checkCurrentTime_` we check to see if we are at the end of the buffer. Thoughts?

src/playback-watcher.js
```diff
@@ -100,6 +101,16 @@ export default class PlaybackWatcher {
     }
 
     let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
+
+    if (buffered.length && currentTime + 0.1 >= buffered.end(buffered.length - 1)) {
```
gesinger: What do you think of putting this in `waiting_` after the `techWaiting_` checks? It might make sense for either, but I'm inclined to keep it closer to the other logic that cancels out of "waiting" type events and keep this method simple.

mjneil: It has to go here actually. If we put it in `waiting_`, it'll be too late and we get the same problem of `consecutiveUpdates` surpassing 5 while waiting on slow connection, then when the buffer is asynchronously appended to, everything will look fine in `techWaiting_`, so `waiting_` will trigger the `unknownwaiting` event. Putting the buffer check here before we increment `consecutiveUpdates` allows us to not consider waiting on a slow connection as a possible stall. That way when we actually run the checks in `waiting_` and `techWaiting_,`, we know those checks are attempting to correct an actual stall, and not just slow connection.

src/playback-watcher.js
```diff
@@ -100,6 +101,18 @@ export default class PlaybackWatcher {
     }
 
     let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
+
+    if (!buffered.length || currentTime + 0.1 >= buffered.end(buffered.length - 1)) {
```
gesinger: I think that `!buffered.length` will prevent us from checking for `fellOutOfLiveWindow_` and running `fixesBadSeeks_ ` on environments where the `waiting` event might not fire and we rely on our implementation.

src/playback-watcher.js
```diff
@@ -156,20 +169,65 @@ export default class PlaybackWatcher {
...
+    let currentTime = this.tech_.currentTime();
+    let buffered = this.tech_.buffered();
+    let currentRange = Ranges.findRange(buffered, currentTime);
+
+    // Sometimes the player can stall for unknown reasons within a contiguous buffered
+    // region with no indication that anything is amiss (seen in Firefox). Seeking to
+    // currentTime is usually enough to kickstart the player. This checks that the player
+    // is currently within a buffered region before attempting a corrective seek.
+    // Chrome does not appear to continue `timeupdate` events after a `waiting` event
+    // until there is ~ 3 seconds of forward buffer available. PlaybackWatcher should also
```
gesinger: are (and below)

src/playback-watcher.js
```diff
@@ -156,20 +169,65 @@ export default class PlaybackWatcher {
...
+    // make sure there is ~3 seconds of forward buffer before taking any corrective action
+    // to avoid triggering an `unknownwaiting` event in when the network is slow.
```
gesinger: ~~in~~

src/playback-watcher.js
```diff
@@ -156,20 +169,65 @@ export default class PlaybackWatcher {
...
+    // Note: This is not done when the `waiting` event fired by the tech because the tech
```
gesinger: is fired

PR #1118, opened 2017-05-17
Throttler: Minor changes to make throttler work with native safari

mrocajr:
## Description
Please describe the change as necessary.
Minor changes to make throttler work with native safari

## Specific Changes proposed
ts file headers set to application/vnd.apple.mpegurl


utils/switcher/throttle-server.js
```diff
@@ -95,7 +95,7 @@ const writeSlowly = (req, res, data) => {
 // maps file extention to MIME type
 // we only care about these two types
 const map = {
-  '.ts': 'application/octet-stream',
+  '.ts': 'application/vnd.apple.mpegurl',
```
gesinger: `vnd.apple.mpegurl` denotes an m3u8 playlist

mrocajr: Perhaps video/MP2T would be more appropriate?

gesinger: That would be good

utils/switcher/throttle-server.js
```diff
@@ -95,8 +95,8 @@ const writeSlowly = (req, res, data) => {
 // maps file extention to MIME type
 // we only care about these two types
 const map = {
-  '.ts': 'application/octet-stream',
-  '.m3u8': 'application/x-mpegurl'
+  '.ts': 'application/vnd.apple.mpegurl',
```
gesinger: I think these two are reversed. And did these have to be changed to get Safari to work?

mrocajr: Yep, just noticed that 👍

gesinger:
Tested, will merge once build passes.

PR #1119, opened 2017-05-17
Add module field to support ES6 modules.

misteroneill:
Tiny update to `package.json` so that Rollup (and other ES6 module systems) can `import` this library easily.

## Requirements Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "5.5.3",
   "description": "Play back HLS with video.js, even where it's not natively supported",
   "main": "es5/videojs-contrib-hls.js",
+  "module": "src/videojs-contrib-hls.js",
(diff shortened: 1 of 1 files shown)


gkatsev:
Unfortunately, because of the usage of webworkify, I'm not sure this is sufficient to cover all bases.

misteroneill:
Well, it'll be needed anyway. Another question: if webworkify is used, why isn't it a dependency of the project?

gkatsev:
It is: https://github.com/videojs/videojs-contrib-hls/blob/master/package.json#L99

gesinger:
Thanks @misteroneill . Closing in favor of https://github.com/videojs/videojs-contrib-hls/pull/1151

PR #1122, opened 2017-05-18
Trigger bandwidthupdate events on the tech

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/videojs-contrib-hls.js
+++ b/src/videojs-contrib-hls.js
@@ -587,10 +587,6 @@ class HlsHandler extends Component {
       this.tech_.trigger('progress');
     });
 
-    this.on(this.masterPlaylistController_, 'bandwidthupdate', function() {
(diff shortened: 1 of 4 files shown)


src/videojs-contrib-hls.js
```diff
@@ -587,6 +587,10 @@ class HlsHandler extends Component {
       this.tech_.trigger('progress');
     });
 
+    this.on(this.masterPlaylistController_, 'bandwidthupdate', function() {
+      this.tech_.trigger('bandwidthupdate');
```
mjneil: any reason to not trigger straight from the tech within MPC if the HLS source handler isn't doing anything other than forwarding the event to the tech?

gesinger: Good point, will just trigger from MPC on the tech.

PR #1134, opened 2017-05-25
Abr low waterline

mjneil:


src/config.js
```diff
@@ -2,5 +2,7 @@ export default {
   GOAL_BUFFER_LENGTH: 60,
   // A fudge factor to apply to advertised playlist bitrates to account for
   // temporary flucations in client bandwidth
-  BANDWIDTH_VARIANCE: 1.2
+  BANDWIDTH_VARIANCE: 1.2,
+  // How much of the buffer must be filled before we consider upswitching
```
gesinger: We should just note for VOD.

src/master-playlist-controller.js
```diff
@@ -544,7 +545,23 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       // figure out what stream the next segment should be downloaded from
       // with the updated bandwidth information
-      this.masterPlaylistLoader_.media(this.selectPlaylist());
+      const nextPlaylist = this.selectPlaylist();
+      const currentPlaylist = this.masterPlaylistLoader_.media();
+      const buffered = this.tech_.buffered();
+      const forwardBuffer = buffered.length ?
+        buffered.end(buffered.length - 1) - this.tech_.currentTime() : 0;
+
+      // we want to switch down to lower resolutions quickly to continue playback, but
```
gesinger: We can just add "For VOD" here, and below we can say "For live, we want to avoid using a low water line, as we only have 3 segment durations worth of content to buffer."

PR #1135, opened 2017-05-30
[ABR] Use AVERAGE-BANDWIDTH when available and we've crossed the buffer low…

gesinger:
… water line

## Requirements Checklist
  - [ ] Docs/guides updated


src/playlist-selectors.js
```diff
@@ -124,27 +146,32 @@ const simpleSelector = function(master, bandwidth, width, height) {
...
   // filter out any playlists that have been excluded due to
   // incompatible configurations or playback errors
   sortedPlaylists = sortedPlaylists.filter(Playlist.isEnabled);
   // filter out any variant that has greater effective bitrate
   // than the current estimated bandwidth
   bandwidthPlaylists = sortedPlaylists.filter(function(elem) {
-    return elem.attributes &&
-           elem.attributes.BANDWIDTH &&
-           elem.attributes.BANDWIDTH * Config.BANDWIDTH_VARIANCE < bandwidth;
+    const elemBandwidth = getBandwidth(elem, useAverageBandwidth);
```
mjneil: I think that this is the only place that using average bandwidth vs bandwidth makes a difference. Is there a way we can contain the changes to here?

gesinger: I believe there should be more places that it makes a difference. For instance, `comparePlaylistResolution` will use BANDWIDTH vs AVERAGE-BANDWIDTH if there is a matching resolution. In addition, it keeps consistency with what bandwidth we are using for comparisons (versus overwriting a playlist's attribute/maintaining a pseudo playlist with altered values), and also combines a lot of the standard guards against missing/empting `attributes` and `BANDWIDTH` properties.

mjneil: We use bandwidth comparisons for two different operations, sorting and filtering.

For sorting, if a playlist has a higher bandwidth, then it should have a higher average-bandwidth as well, so it should be safe to use `BANDWIDTH` instead of the average for sorting.

For filtering, (ignoring the filter I commented on saying the average makes a difference), we compare the bandwidths to check that they are equal for the purpose of identifying redundant streams. The attributes of redundant streams should be the same, so it should be safe to use either here. Apple also doesn't specify if the `AVERAGE-BANDWIDTH` attribute needs to be the same across redundant streams, so I may even argue not using the average is safer here to make sure primary is chosen over backup.

I'd say that it actually is less consistent with which bandwidth we are using for comparisons and guards less against missing/empty attributes. Since `AVERAGE-BANDWIDTH` is an optional property, if we opt to use it in either a sorting or filtering scenario, all it takes is for one stream in the list to not include the attribute and all the sort/filters will be incorrect.

The filter at this line however, is comparing bandwidths to the `HlsHandler`s computed bandwidth numbers in order to filter out playlists that have too high bitrate for the current bandwidth. Using the `AVERAGE-BANDWIDTH` here lets us consider a playlist we would otherwise throw out because its max bandwidth was too high, but on average, is lower bandwidth.

(made a separate comment regarding `comparePlaylistResolution`)

gesinger: `For sorting, if a playlist has a higher bandwidth, then it should have a higher average-bandwidth as well`
That is actually part of the reason why AVERAGE-BANDWIDTH can sometimes make a difference, is because it is possible for a playlist to have a higher BANDWIDTH but a lower AVERAGE-BANDWIDTH than another playlist (and vice versa).

`Since AVERAGE-BANDWIDTH is an optional property, if we opt to use it in either a sorting or filtering scenario, all it takes is for one stream in the list to not include the attribute and all the sort/filters will be incorrect.`
In practice, most m3u8s should exhibit all or nothing, either AVERAGE-BANDWIDTH is present on all playlists, or it is present on none. However, in the event that it is only available on some, AVERAGE-BANDWIDTH will be used unless it is not available, at which point the compared to playlist will use BANDWIDTH. While that isn't an "apples-to-apples" comparison,  it should be using a "best available information" approach.

src/playlist-selectors.js
```diff
@@ -103,17 +120,22 @@ export const comparePlaylistResolution = function(left, right) {
...
+  const rightBandwidth = getBandwidth(right, useAverageBandwidth);
+
   // NOTE - Fallback to bandwidth sort as appropriate in cases where multiple renditions
-  // have the same media dimensions/ resolution
-  if (leftWidth === rightWidth &&
-      left.attributes.BANDWIDTH &&
-      right.attributes.BANDWIDTH) {
-    return left.attributes.BANDWIDTH - right.attributes.BANDWIDTH;
+  // have the same media dimensions/ resolution.
+  if (leftWidth === rightWidth && leftBandwidth && rightBandwidth) {
```
gesinger: That's a good point. We should be able to remove that comparison. The only reason we might be inclined to keep it is for backwards compatibility since `comparePlaylistResolution` is exposed.

src/playlist-selectors.js
mjneil: `In practice, most m3u8s should exhibit all or nothing, either AVERAGE-BANDWIDTH is present on all playlists, or it is present on none.`
I agree with this, and a "best available information" approach for this case seems like it will still give desired results.

`it is possible for a playlist to have a higher BANDWIDTH but a lower AVERAGE-BANDWIDTH than another playlist`
I'm not entirely convinced this means we want to use the average when sorting. If I'm not mistaken, the goal of the playlist selector is to select the highest bandwidth+resolution we can support with the current system bandwidth. If a playlist with a lower BANDWIDTH has a higher AVERAGE than others, then it will be considered to have a higher bandwidth in sorting. This could make the player select the playlist with a lower overall bandwidth but higher average. Maybe this is desired, but from my understanding it is not. 

Redundant streams with differing averages still concerns me, but I haven't seen any real world examples of redundant streams with average-bandwidth to know whether they are likely to differ or not.

src/playlist-selectors.js
mjneil: I don't think we need to keep it for that reason. `comparePlaylistResolution` may be exposed, but its exposure is not documented anywhere as something that can/should be changed, so I think its reasonable to change it. This is the same reasoning as the call to not worry about this method (and the bandwidth comparator) being moved from `videojs-contrib-hls.js` to `playlist-selectors.js` such that `STANDARD_PLAYLIST_SELECTOR` no longer uses the comparators attached to the `Hls` object.

gesinger: That's definitely true, though if someone was using `comparePlaylistResolution` as part of their own `selectPlaylist` function (or for some other reason), the behavior difference could be noticeable (albeit, that is probably not too common of a case).

src/playlist-selectors.js
gesinger: I believe that is the intention. The idea of using AVERAGE-BANDWIDTH is that when we've crossed a certain threshold, we're no longer worried about what a max bandwidth is on an individual segment (since we have buffer shielding us), but are more concerned with what the segments average out to be.

src/playlist-selectors.js
```diff
@@ -103,91 +120,118 @@ export const comparePlaylistResolution = function(left, right) {
...
-  let resolutionPlusOneList = [];
-  let resolutionPlusOneSmallest = [];
-  let resolutionBestVariantList = [];
 
-  stableSort(sortedPlaylists, comparePlaylistBandwidth);
+  stableSort(sortedPlaylists, comparePlaylistBandwidth.bind(null, useAverageBandwidth));
+
+  // convert the playlists to an intermediary representation to make comparisons easier
+  // and prevent us from re-determining bandwidth each time
+  let sortedPlaylistReps = sortedPlaylists.map((playlist) => {
```
mjneil: Do this first, and then use `stableSort` on this list for both `comparePlaylistBandwidth` and `comparePlaylistResolution`. This way both those functions do not need to add a new parameter `useAverageBandwidth` and you can use `comparePlaylistResolution` again instead of the anonymous function.  Then you just need to check `rep.bandwidth` in the compare functions (this makes `comparePlaylistBandwidth` a one-liner!). 

Id say the main argument against this approach which I predict will concern you, is that it changes the compare functions internally such that anyone using the exposed compare functions (even though we've never said you should do that) will break. Again I still believe this is ok. We've already broken overriding the compare functions but not the simple selector by moving these functions to their own module.

src/playlist-selectors.js
```diff
@@ -1,8 +1,35 @@
...
+
+  if (useAverageBandwidth) {
+    bandwidth = playlist.attributes['AVERAGE-BANDWIDTH'];
+  }
+
+  if (!bandwidth) {
+    bandwidth = playlist.attributes.BANDWIDTH;
+  }
+
+  return bandwidth || void 0;
```
mjneil: the `|| void 0` isn't needed here since if `playlist.attributes.BANDWIDTH` doesn't exist, `bandwidth` will already be `undefined`. Though I'm ok with leaving it in for the sake of making it clear the function returns `undefined` in some cases

src/playlist-selectors.js
```diff
@@ -103,91 +125,118 @@ export const comparePlaylistResolution = function(left, right) {
...
-  let resolutionBestVariant;
-  let haveResolution;
-  let resolutionPlusOneList = [];
-  let resolutionPlusOneSmallest = [];
-  let resolutionBestVariantList = [];
 
-  stableSort(sortedPlaylists, comparePlaylistBandwidth);
+  // convert the playlists to an intermediary representation to make comparisons easier
+  // and prevent us from re-determining bandwidth each time
+  let sortedPlaylistReps = sortedPlaylists.map((playlist) => {
```
mjneil: since we are no longer sorting the `sortedPlaylists` itself, this can just be `master.playlists.map(...` without affecting the sorting of the original list`

src/playlist-selectors.js
```diff
@@ -103,91 +125,118 @@ export const comparePlaylistResolution = function(left, right) {
...
+
   // filter out any variant that has greater effective bitrate
   // than the current estimated bandwidth
-  bandwidthPlaylists = sortedPlaylists.filter(function(elem) {
-    return elem.attributes &&
-           elem.attributes.BANDWIDTH &&
-           elem.attributes.BANDWIDTH * Config.BANDWIDTH_VARIANCE < bandwidth;
-  });
+  let bandwidthPlaylistReps = sortedPlaylistReps.filter(
+    (rep) => rep.bandwidth && rep.bandwidth * Config.BANDWIDTH_VARIANCE < bandwidth
```
mjneil: don't need `rep.bandwidth &&` since you default to `MAX_VALUE` when creating the reps

src/playlist-selectors.js
```diff
@@ -103,91 +125,118 @@ export const comparePlaylistResolution = function(left, right) {
...
-  haveResolution = bandwidthPlaylists.filter(function(elem) {
-    return elem.attributes &&
-           elem.attributes.RESOLUTION &&
-           elem.attributes.RESOLUTION.width &&
-           elem.attributes.RESOLUTION.height;
+    return leftWidth - rightWidth;
   });
 
+  // filter out playlists without resolution information
+  let haveResolution = bandwidthPlaylistReps.filter((rep) => rep.width && rep.height);
```
mjneil: If you do this before the sort above, and then stable sort `haveResolution` you can remove the local constants and make it a one-line like the bandwidth sort

src/playlist-selectors.js
```diff
@@ -59,17 +86,10 @@ const stableSort = function(array, sortFn) {
...
 
-  if (left.attributes && left.attributes.BANDWIDTH) {
-    leftBandwidth = left.attributes.BANDWIDTH;
-  }
-  leftBandwidth = leftBandwidth || window.Number.MAX_VALUE;
-  if (right.attributes && right.attributes.BANDWIDTH) {
-    rightBandwidth = right.attributes.BANDWIDTH;
-  }
-  rightBandwidth = rightBandwidth || window.Number.MAX_VALUE;
+  videojs.log.warn('comparePlaylistBandwidth is deprecated');
```
mjneil: Should we add a suggestion on what alternative the user has if they get this message? e.g. one of our other deprecation messages says `player.hls is deprecated. Use player.tech_.hls instead.` I guess the alternative here is, write your own compare function, so maybe an alternative doesn't really apply here.

test/playlist-selectors.test.js
```diff
@@ -51,3 +61,301 @@ test('Exponential moving average has a configurable decay parameter', function(a
...
+
+  playlist.attributes['AVERAGE-BANDWIDTH'] = 3;
+
+  assert.equal(getBandwidth(playlist, false),
+               9,
+               'uses BANDWIDTH when useAverageBandwidth is false');
+  assert.equal(getBandwidth(playlist, true),
+               3,
+               'uses AVERAGE-BANDWIDTH when useAverageBandwidth is true');
+});
```
mjneil: Maybe add a test that if `useAverageBandwidth` is true but the avg isn't available, it falls back to the max

test/playlist-selectors.test.js
```diff
@@ -51,3 +61,301 @@ test('Exponential moving average has a configurable decay parameter', function(a
...
+
+test('crossedLowWaterLine detects when we have filled enough buffer', function(assert) {
+  assert.notOk(crossedLowWaterLine(0, videojs.createTimeRanges([])),
```
mjneil: For this test (and any others that use any of the constants in `Config`, I think it would be a good idea to override the config option with a set value for the test, and then restore is at the end. That way down the line if we decide that there is a better default value for that constant, we don't have to change all the tests that are written in a way that assume the old default value

mjneil: Or use it when creating the time ranges like you did below `videojs.createTimeRanges([[10, 10 + Config.BUFFER_LOW_WATER_LINE]])`

src/playlist-selectors.js
```diff
@@ -103,91 +125,118 @@ export const comparePlaylistResolution = function(left, right) {
...
+    };
+  });
+
+  stableSort(sortedPlaylistReps, (left, right) => left.bandwidth - right.bandwidth);
 
   // filter out any playlists that have been excluded due to
   // incompatible configurations or playback errors
-  sortedPlaylists = sortedPlaylists.filter(Playlist.isEnabled);
+  sortedPlaylistReps = sortedPlaylistReps.filter(
+    (rep) => Playlist.isEnabled(rep.playlist));
```
mjneil: In the name of style consistency, can we put `);` on the next line like you do in line 222

test/playlist-selectors.test.js
gesinger: I believe that should be a few lines up (`uses BANDWIDTH when only one available`)

test/playlist-selectors.test.js
```diff
@@ -51,3 +61,317 @@ test('Exponential moving average has a configurable decay parameter', function(a
...
+               void 0,
+               'undefined when playlist has no info');
+
+  playlist = {
+    attributes: {}
+  };
+
+  assert.equal(getBandwidth(playlist, false),
+               void 0,
+               'bandwidth 0 when playlist has no info');
```
mjneil: `bandwdith ~~0~~ undefined`

Issue #1108, opened 2017-05-10
Abort requests we have no hope of completing in time


imbcmdth:
We have progress events coming from `mediaSegmentRequest`  that is quite handy. We should leverage that data in that event to decide when to abort a segment request and trigger a rendition switch.

## Acceptance Criteria
* Implement the feature
* Test it using bandwidth simulator
* Do a managed release

## How
* Improves ABR -
   * Start up time should improve in low bandwidth scenarios
   * Ability to respond to catastrophic bandwidth drops should improve
      * Should lower rebuffering stats (time and count)
* Listen for progress events and judge the ability for the player to complete the request in a reasonable time
* We currently have a hard-timeout. That *should* be unnecessary in this new world. It might make sense to keep it in the event that progress events are not sent or if they don't contain the kind of data that would allow us to make decisions.
* We need to update the `bandwidth` property and trigger the `bandwidthchange` if we decide to abort because otherwise playlist selection doesn't run

## Notes
* Bandwidth starts off REALLY low on requests
* The bandwidth simulator DOES simulate progress events (you are welcome)

mjneil:
this was added in #1176

PR #1136, opened 2017-05-31
[ABR] Abort on progress events during a segment request if there is no hope…

mjneil:
… of downloading the segment in a reasonable time
#1108

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -79,13 +79,12 @@ const getProgressStats = (progressEvent) => {
     roundTripTime: roundTripTime || 0
   };
 
-  if (progressEvent.lengthComputable) {
-    stats.bytesReceived = progressEvent.loaded;
-    // This can result in Infinity if stats.roundTripTime is 0 but that is ok
-    // because we should only use bandwidth stats on progress to determine when
-    // abort a request early due to insufficient bandwidth
-    stats.bandwidth = Math.floor((stats.bytesReceived / stats.roundTripTime) * 8 * 1000);
-  }
+  stats.bytesReceived = progressEvent.loaded;
+  // This can result in Infinity if stats.roundTripTime is 0 but that is ok
+  // because we should only use bandwidth stats on progress to determine when
+  // abort a request early due to insufficient bandwidth
+  stats.bandwidth = Math.floor((stats.bytesReceived / stats.roundTripTime) * 8 * 1000);
+
   return stats;
 };
 
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -332,7 +332,13 @@ const waitForCompletion = (activeXhrs, decrypter, doneFn) => {
  * @param {Event} event - the progress event object from XMLHttpRequest
  */
 const handleProgress = (segment, progressFn) => (event) => {
-  segment.stats = getProgressStats(event);
+  segment.stats = videojs.mergeOptions(segment.stats, getProgressStats(event));
+
+  // record the time that we receive the byte of data
+  if (!segment.stats.firstByteReceived && segment.stats.bytesReceived) {
+    segment.stats.firstByteReceived = Date.now();
+  }
+
   return progressFn(event, segment);
 };
 
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -717,6 +717,26 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  updateHandler_(event, segment) {
+    if (!this.pendingSegment_ || segment.requestId !== this.pendingSegment_.requestId) {
+      return;
+    }
+
+    const bandwidthAdjustment = Math.min(0.8, segment.stats.roundTripTime / 5000);
+
+    // TODO: Replace timeout with a boolean indicating whether this playlist is the lowestEnabledRendition
+    if (this.xhrOptions_.timeout &&
+        segment.stats.roundTripTime > 1000 &&
+        segment.stats.bandwidth < this.playlist_.attributes.BANDWIDTH * bandwidthAdjustment) {
+      this.bandwidth = segment.stats.bandwidth;
+      this.trigger('bandwidthupdate');
+      this.abort();
+      return;
+    }
+
+    this.trigger('progress');
+  }
+
   /**
    * load a specific segment from a request into the buffer
    *
@@ -732,14 +752,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.decrypter_,
       this.createSimplifiedSegmentObj_(segmentInfo),
       // progress callback
-      (event, segment) => {
-        if (!this.pendingSegment_ ||
-            segment.requestId !== this.pendingSegment_.requestId) {
-          return;
-        }
-        // TODO: Use progress-based bandwidth to early abort low-bandwidth situations
-        this.trigger('progress');
-      },
+      this.updateHandler_.bind(this),
       this.segmentRequestFinished_.bind(this));
   }
 
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -722,19 +722,80 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    // the amount of time since the first byte of data was received
-    const firstByteReceived = Date.now() - segment.stats.firstByteReceived;
-    const adjustedBandwidth = segment.stats.bandwidth * Config.BANDWIDTH_VARIANCE;
-
+    // If the player is not paused and the current playlist is not the
+    // lowestEnabledRendition, consider the possibility of aborting the current request
+    // early for an emergency downswitch
     // TODO: Replace timeout with a boolean indicating whether this playlist is the
     //       lowestEnabledRendition
-    if (this.xhrOptions_.timeout &&
-        firstByteReceived > 1000 &&
-        adjustedBandwidth < this.playlist_.attributes.BANDWIDTH) {
-      this.bandwidth = segment.stats.bandwidth;
-      this.trigger('bandwidthupdate');
-      this.abort();
-      return;
+    if (!this.hls_.tech_.paused() && this.xhrOptions_.timeout) {
+      const firstByteReceived = Date.now() - segment.stats.firstByteReceived;
+
+      // Wait at least 1 second before using the calculated bandwidth from the
+      // progress event to allow the bitrate to stableize
+      if (firstByteReceived > 1000) {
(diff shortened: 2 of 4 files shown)


src/media-segment-request.js
```diff
@@ -79,13 +79,12 @@ const getProgressStats = (progressEvent) => {
...
   };
 
-  if (progressEvent.lengthComputable) {
-    stats.bytesReceived = progressEvent.loaded;
-    // This can result in Infinity if stats.roundTripTime is 0 but that is ok
-    // because we should only use bandwidth stats on progress to determine when
-    // abort a request early due to insufficient bandwidth
-    stats.bandwidth = Math.floor((stats.bytesReceived / stats.roundTripTime) * 8 * 1000);
-  }
+  stats.bytesReceived = progressEvent.loaded;
```
gesinger: `loaded` only seems to be supported in Firefox and Edge. Is there any alternative for other browsers? Otherwise, do we want to handle the case where it is undefined?

mjneil: hmm I do see MDN says that, but it is working in chrome, so the MDN pages for progress events must be very out of date

src/segment-loader.js
```diff
@@ -717,6 +717,43 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  abortRequestEarly_(stats) {
+    const firstByteReceived = Date.now() - stats.firstByteReceived;
+    // Since bandwidth values reported by progress events stabilize as time goes on,
+    // we adjust bandwidth by a conservative ammount, increasing as more time has passed
```
gesinger: amount

src/segment-loader.js
```diff
@@ -717,6 +717,43 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // maxing at 0.8 to prevent aborting a request we actually do have enough bandwidth
+    // for.
+    const playlistBandwidthAdjustment = Math.min(0.8, firstByteReceived / 5000);
+    const playlistBandwidth = this.playlist_.attributes.BANDWIDTH;
```
gesinger: Should we check that this is available (e.g., media playlist cases)

src/segment-loader.js
```diff
@@ -717,6 +717,43 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  abortRequestEarly_(stats) {
+    const firstByteReceived = Date.now() - stats.firstByteReceived;
+    // Since bandwidth values reported by progress events stabilize as time goes on,
+    // we adjust bandwidth by a conservative ammount, increasing as more time has passed
+    // maxing at 0.8 to prevent aborting a request we actually do have enough bandwidth
+    // for.
+    const playlistBandwidthAdjustment = Math.min(0.8, firstByteReceived / 5000);
```
gesinger: Why 5 seconds? Would it make more sense around target duration or the segment's specified duration?

src/segment-loader.js
```diff
@@ -717,6 +717,43 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  abortRequestEarly_(stats) {
+    const firstByteReceived = Date.now() - stats.firstByteReceived;
```
gesinger: Might make more sense as `millisUntilFirstByteReceived`

src/segment-loader.js
```diff
@@ -717,6 +717,43 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // for.
+    const playlistBandwidthAdjustment = Math.min(0.8, firstByteReceived / 5000);
+    const playlistBandwidth = this.playlist_.attributes.BANDWIDTH;
+
+    // Wait at least 1 second before using the calculated bandwidth from the
+    // progress event to allow the bitrate to stabilize. Bandwidth information
+    // during the first second is highly variable and inaccurate.
+    // TODO: Replace timeout with a boolean indicating whether this playlist is the
+    // lowestEnabledRendition
+    if (this.xhrOptions_.timeout &&
```
gesinger: We should have comments above each part explaining it. For instance

```
// checks timeout property as is indicates whether this playlist is on lowestEnabledRendition
// TODO replace with a boolean indicating lowestEnabledRendition
if (this.xhrOptions_.timeout &&
    // Wait at least 1 second before using the calculated bandwidth from the
    // progress event to allow the bitrate to stabilize. Bandwidth information
    // during the first second is highly variable and inaccurate.
    firstByteReceived > 1000 &&
    stats.bandwidth < playlistBandwidth * playlistBandwidthAdjustment) {
  ...
}

File: src/media-segment-request.js (hunk 1 above)
Comment (gesinger) [inline src/media-segment-request.js]: If it is undefined, should we handle the case differently?

File: test/loader-common.js [hunk 7]
```diff
@@ -277,6 +277,54 @@ export const LoaderCommonFactory = (LoaderConstructor,
       assert.equal(progressEvents, 1, 'triggered progress');
     });
 
+    QUnit.test('aborts request at progress events if bandwidth is too low',
+    function(assert) {
+      const playlist = playlistWithDuration(10);
+      const xhrOpotions = {
```
gesinger: Options

src/media-segment-request.js
mjneil: I'll do a double check in our platforms to see whats in `event.loaded` I think at this spot we dont have to do anything different, but in the abort early section when we look at the `stats` we should check that its a valid value

src/segment-loader.js
mjneil: 5 seconds is arbitrary. I like the idea of using target duration or the segments specified duration. Maybe half the segment duration? If we've already spent half the duration requesting the segment and bandwidth is still low, it might be a good time to switch. Another idea is what if we incorporate forward buffer so that the closer we get to the end of the buffer, the higher the adjustment ramps up?

src/media-segment-request.js
mjneil: IE11 and Chrome both are giving values in `loaded`, so we should be covered but have a guard in case is a good idea

src/segment-loader.js
mjneil: well here `const firstByteReceived` is the time that has passed since we received the first byte of data. `msSinceFirstByteReceived` maybe

src/media-segment-request.js
mjneil: Actually, we don't need any checks because `NaN < Number` will always return false, so we won't actually act upon `undefined` values

src/segment-loader.js
gesinger: Maybe a hard number is OK. Since it is dependent on ramp up time for requests, that should be consistent regardless of segment size. Have we seen a max amount of time for requests to ramp up to full bandwidth?

src/segment-loader.js
gesinger: Makes sense 👍

src/media-segment-request.js
gesinger: Can we guarantee we will always get NaN? Maybe it would be better to do something like `progressEvent.loaded || 0` (or some other value that we can guarantee)

mjneil: If you do a math operation with `undefined` you'll get nan, so `stats.bandwidth = Math.floor((stats.bytesReceived / stats.roundTripTime) * 8 * 1000);` will be `NaN` if `bytesReceived` is `NaN`. Even if `NaN` wasn't guaranteed, compare operations with `undefined` will be false as well. I'm skeptical of setting it to any other numerical number because then we may act upon that value when really we only want to act when we know we have actual info

src/segment-loader.js
```diff
@@ -717,6 +717,53 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  abortRequestEarly_(segment) {
+    const stats = segment.stats;
+    const msSinceFirstByte = Date.now() - stats.firstByteReceived;
+    const RTT = stats.roundTripTime;
```
gesinger: This variable loses some clarity of what it actually is, and `stats.roundTripTime` is only used in one place. Probably better to not use a variable for it.

src/media-segment-request.js
gesinger: Using `undefined` or `NaN` in formulas still concerns me. Maybe instead we only set `stats.bandwidth` if `progressEvent.loaded` is a valid number, and if it isn't, then `abortRequestEarly_` can simply return false if `stats.bandwidth` is `undefined`?

src/segment-loader.js
```diff
@@ -717,6 +717,53 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // Since bandwidth values reported by progress events stabilize as time goes on,
+    // we adjust bandwidth by a conservative amount, increasing as more time has passed
+    // maxing at 0.8 to prevent aborting a request we actually do have enough bandwidth
+    // for.
+    const playlistBandwidthAdjustment = Math.min(0.8, RTT / segment.duration);
+    const playlistBandwidth = this.playlist_.attributes &&
```
gesinger: Might be easier to above just do:

```
if (!this.playlist_.playlist || !this.playlist_.attributes.BANDWIDTH) {
  return false;
}
```

src/segment-loader.js
```diff
@@ -717,6 +717,53 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // TODO: Replace timeout with a boolean indicating whether this playlist is the
+    // lowestEnabledRendition
+    if (this.xhrOptions_.timeout &&
+        // Wait at least 1 second before using the calculated bandwidth from the
+        // progress event to allow the bitrate to stabilize. Bandwidth information
+        // during the first second can be highly variable and inaccurate.
+        msSinceFirstByte > 1000 &&
+        // Lastly, if the calculated bandwidth for the current request is less than
+        // the bitrate of the playlist (adjusted downwards so that we don't jump the gun)
+        // then we want to abort this request and try a smaller rendition that we could
```
gesinger: can

src/segment-loader.js
mjneil: Well, I feel like this adjustment number is doing 2 things. 
1. is accounting for bandwidth ramp up, as time goes on, bandwidth will have ramped up and become more stable, so we can use a value closer to the real bandwidth of the playlist. This is also why below we make sure its been about 1 second since first byte seen. I've run some (granted not very exhaustive) experiments that seemed to show around 1 second is when bandwidth from the progress events stabilizes.
2. is that increasing the adjustment as time goes on, allows us to be more likely to emergency abort when the request is taking too long. I like the idea of using segment duration here because 5 seconds may not be too long for 10 second segments, but for 2 second segments it is definitely way too long, and we probably should've been using the full `08.` well before 4 seconds since request

Issue #1113, opened 2017-05-10
Try saving bandwidth and throughput values into localStorage and use them if no other information is available at start of playback


imbcmdth:

forbesjo:
Closing since there's a PR open

PR #1142, opened 2017-06-01
[ABR] Store bandwidth and throughput values in local storage and use them o…

gesinger:
…n startup

## Description
https://github.com/videojs/videojs-contrib-hls/issues/1113

## Requirements Checklist
  - [ ] Docs/guides updated


mjneil:
Didn't we want to turn on the storing of local storage but leave the using of it off for now?

Also do you think this would be a good thing to allow turning off and on through the player options?

README.md
```diff
@@ -306,6 +308,19 @@ When the `bandwidth` property is set (bits per second), it will be used in
 the calculation for initial playlist selection, before more bandwidth
 information is seen by the player.
 
+##### storeBandwidthInLocalStorage
```
mjneil: Is this option needed? Any reason to not just have storage always enabled, and allow the setting of whether its actually used?

gesinger: Since local storage is limited in size (before access to more space must be granted by the site visitor), it is possible that a user of videojs-contrib-hls doesn't want the project using some of the space if they've carefully accounted for the amount allotted to them.

forbesjo: Can't `useBandwidthFromLocalStorage` do both?

mjneil: It's only storing 2 ints though, so I don't think it will be a big issue. Either way, I agree with @forbesjo , we should just use one flag to turn off both storing and using. I know previously we've discussed turning on the storing but not the using, but I don't think that matters much, especially since we are wanting to go the usage event route

gesinger: Sounds good 👍 . Will change to just one.

mjneil:
We should add the usage tracking event for when localstorage values have been used

src/videojs-contrib-hls.js
```diff
@@ -280,11 +280,28 @@ class HlsHandler extends Component {
...
+        if (storedThroughput) {
+          this.options_.throughput = parseInt(storedThroughput, 10);
+        }
+      }
+
+      // start playlist selection at a reasonable bandwidth for
+      // broadband internet
+      // 0.5 MB/s
+      if (typeof this.options_.bandwidth !== 'number') {
+        this.options_.bandwidth = 4194304;
```
OshinKaramian: Might make sense to put this magic number into a const somewhere.

test/videojs-contrib-hls.test.js
```diff
@@ -2687,6 +2689,108 @@ QUnit.test('populates quality levels list when available', function(assert) {
   assert.ok(this.player.tech_.hls.qualityLevels_, 'added quality levels from video with source');
 });
 
+QUnit.test('stores bandwidth and throughput in localStorage', function(assert) {
+  const origHlsOptions = videojs.options.hls;
```
OshinKaramian: This is a big test... it might make sense (if possible) to break it down into a few smaller tests.  Tiny tests make debugging and finding issues a little bit easier.

test/videojs-contrib-hls.test.js
```diff
@@ -2687,6 +2689,108 @@ QUnit.test('populates quality levels list when available', function(assert) {
...
+
+  // values must be stored before player is created, otherwise defaults are provided
+  this.player = createPlayer();
+  this.player.src({
+    src: 'manifest/master.m3u8',
+    type: 'application/vnd.apple.mpegurl'
+  });
+
+  assert.equal(this.player.tech_.hls.bandwidth,
+               4194304,
```
OshinKaramian: If you export the const you might be able to reference it here.

src/videojs-contrib-hls.js
```diff
@@ -247,8 +247,25 @@ class HlsHandler extends Component {
       this.options_.blacklistDuration = 5 * 60;
     }
 
-    // start playlist selection at a reasonable bandwidth for
-    // broadband internet (0.5 MB/s) or mobile (0.0625 MB/s)
+    if (typeof this.options_.bandwidth !== 'number') {
+      if (this.options_.useBandwidthFromLocalStorage && window.localStorage) {
+        let storedBandwidth =
```
OshinKaramian: Nit: Is there any reason this can't be a `const`?

src/videojs-contrib-hls.js
```diff
@@ -247,8 +247,25 @@ class HlsHandler extends Component {
...
+
+        if (storedThroughput) {
+          this.options_.throughput = parseInt(storedThroughput, 10);
+        }
+      }
+    }
+
+    // if bandwidth was not set by options or pulled from local storage, start playlist
+    // selection at a reasonable bandwidth
     if (typeof this.options_.bandwidth !== 'number') {
```
squarebracket: It's semantically equivalent, but I'm wondering why you didn't include this as an `else` in the clause above. i.e. if use local storage, grab it, otherwise set the bandwidth from a constant.

gesinger: It's possible that the user did not specify a bandwidth value and set `useBandwidthFromLocalStorage` to true, but there was nothing in local storage, leaving bandwidth undefined.

squarebracket: ah good call

test/videojs-contrib-hls.test.js
```diff
@@ -2932,6 +2934,108 @@ QUnit.test('populates quality levels list when available', function(assert) {
   assert.ok(this.player.tech_.hls.qualityLevels_, 'added quality levels from video with source');
 });
 
+QUnit.test('stores bandwidth and throughput in localStorage when option is true',
+function(assert) {
+  videojs.options.hls = {
```
gkatsev: should there be a test that sets it in the player directly?

README.md
```diff
@@ -309,6 +310,13 @@ When the `bandwidth` property is set (bits per second), it will be used in
 the calculation for initial playlist selection, before more bandwidth
 information is seen by the player.
 
+##### useBandwidthFromLocalStorage
+* Type: `boolean`
+* can be used as an initialization option
+
+If true, `bandwidth` and `throughput` values are stored in and retrieved from local
```
gkatsev: would be good to add explanation of default behavior

forbesjo:
Let's open this against VHS

forbesjo:
@gkatsev what are your thoughts on plugins using local storage? Should there be a common API provided by video.js or should the plugins do this directly?

gkatsev:
I think we shouldn't store things in local storage by default but have an option to enable it. This is because of the EU cookie laws where users have to be informed if things are stored locally. While local storage isn't specifically called out, it's best to play it safe. In Video.js, we have an option to persist caption settings for the user but it is disabled by default.
Then, when a site that has the cookie notification uses Video.js, they could enable the option to persist this information.

With ABR, we'd have two things that are stored in local storage, optionally, I don't think it quite warrants a combined API, but we should keep our eyes out, maybe we'll want to combine both of these into a common videojs api and a common `allowPersistence` option that enables or disabled all persistence via that new API.

gesinger:
Closing in favor of https://github.com/videojs/http-streaming/pull/275

PR #1147, opened 2017-06-02
[ABR] Use dynamic changed low-water-line and GOAL_BUFFER_LENGTH

zhuangs:
## Requirements Checklist
  - [ ] Docs/guides updated


src/config.js
```diff
@@ -1,8 +1,12 @@
...
+  MAX_GOAL_BUFFER_LENGTH: 60,
+  GOAL_BUFFER_RATE: 1,
   // A fudge factor to apply to advertised playlist bitrates to account for
   // temporary flucations in client bandwidth
   BANDWIDTH_VARIANCE: 1.2,
   // How much of the buffer must be filled before we consider upswitching
-  BUFFER_LOW_WATER_LINE: 30
+  BUFFER_LOW_WATER_LINE: 0,
+  MAX_BUFFER_LOW_WATER_LINE: 30,
+  BUFFER_LOW_WATER_RATE: 1
```
mjneil: Do we want separate rates for low waterline and goal buffer length? Or a single rate that they both increase at?

Another note, maybe instead of defining a rate so that we scale current time with and add to the min, what about defining a max, and a time at which the max should be reached. and just interpolate. e.g. with wanting to reach max at 30 seconds, `let t = Math.min(currentTime / 30, 1); GBL = (1 - t) * 30 + t * 60;`

Thoughts @gesinger ?

gesinger: I like separating low waterline and goal buffer length variables for finer control.

As for time vs rate, I think it would be worth testing both.

mjneil: Sticking w/ Rate for this PR

test/master-playlist-controller.test.js
```diff
@@ -812,23 +812,46 @@ function(assert) {
...
 
+  currentTime = 10;
+  currentPlaylistBandwidth = 1000;
+  nextPlaylistBandwidth = 1001;
+  buffered = [];
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.equal(mediaChanges.length,
+               0,
+               'did not change media when no buffer and and higher bandwidth playlist');
+  buffered = [[0, 10], [10, 19]];
```
mjneil: You can just make this a single region `[[0, 19]]` and same with the `buffered` a few lines below

src/videojs-contrib-hls.js
```diff
@@ -77,6 +113,43 @@ Object.defineProperty(Hls, 'BUFFER_LOW_WATER_LINE', {
...
+  set(v) {
+    videojs.log.warn('using Hls.MAX_BUFFER_LOW_WATER_LINE is UNSAFE be sure ' +
+                     'you know what you are doing');
+    if (typeof v !== 'number' || v < 0 || v > Config.GOAL_BUFFER_LENGTH) {
+      videojs.log.warn('value passed to Hls.BUFFER_LOW_WATER_LINE ' +
+                       'must be a number and greater than or equal to 0 and less than' +
+                       'Hls.GOAL_BUFFER_LENGTH');
+      return;
+    }
+    Config.BUFFER_LOW_WATER_LINE = v;
```
mjneil: `Config.MAX_BUFFER_LOW_WATER_LINE`

mjneil:
LGTM. @gesinger thoughts on the defaults for the constants? 0-30 low waterline, 30-60 goal buffer length at a rate of 1 will reach the max at 30 seconds. Are these the values we want to test out?

gesinger:
Those numbers sound good to me @mjneil

PR #1148, opened 2017-06-02
[ABR] use EWMA playlist selector with decay of 0.5

mjneil:


src/playlist-selectors.js
```diff
@@ -235,7 +235,8 @@ export const movingAverageBandwidthSelector = function(decay) {
       average = this.systemBandwidth;
     }
 
-    average = decay * this.systemBandwidth + (1 - decay) * average;
+    average =
+      (Config.EWMA_DECAY) * this.systemBandwidth + (1 - (Config.EWMA_DECAY)) * average;
```
gesinger: Don't need parens around Config.EWMA_DECAY (both cases)

src/videojs-contrib-hls.js
```diff
@@ -76,6 +75,24 @@ Object.defineProperty(Hls, 'BANDWIDTH_VARIANCE', {
...
+Object.defineProperty(Hls, 'EWMA_DECAY', {
+  get() {
+    videojs.log.warn('using Hls.EWMA_DECAY is UNSAFE be sure ' +
+                     'you know what you are doing');
+    return Config.EWMA_DECAY;
+  },
+  set(v) {
+    videojs.log.warn('using Hls.EWMA_DECAY is UNSAFE be sure ' +
+                     'you know what you are doing');
+    if (typeof v !== 'number' || v <= 0 || v > 1) {
```
gesinger: It is allowed to be 0 in the function itself.

test/videojs-contrib-hls.test.js
```diff
@@ -42,6 +43,9 @@ QUnit.module('HLS', {
     this.clock = this.env.clock;
     this.old = {};
 
+    this.old.originalSelectPlaylist = Hls.STANDARD_PLAYLIST_SELECTOR;
+    Hls.STANDARD_PLAYLIST_SELECTOR = movingAverageBandwidthSelector();
```
gesinger: Isn't this already the default?

test/videojs-contrib-hls.test.js
```diff
@@ -782,20 +785,29 @@ QUnit.test('raises the minimum bitrate for a stream proportionially', function(a
...
+  master.playlists[0].attributes.BANDWIDTH = 10;
   this.player.tech_.hls.bandwidth = 11;
 
+  // starting bandwidth is 4194304
+  // new bandwidth is 11
+  // EWMA is: DECAY * 11 + (1 - DECAY) * 4194304
+  const EWMA = Hls.EWMA_DECAY * 11 + (1 - Hls.EWMA_DECAY) * 4194304;
+  const safeBandwidth = Math.floor(EWMA / Hls.BANDWIDTH_VARIANCE) - 1;
+
   // 9.9 * 1.1 < 11
```
gesinger: Should this comment be removed?

test/videojs-contrib-hls.test.js
```diff
@@ -811,6 +823,10 @@ QUnit.test('uses the lowest bitrate if no other is suitable', function(assert) {
 
   // the lowest bitrate playlist is much greater than 1b/s
   this.player.tech_.hls.bandwidth = 1;
+  this.player.tech_.hls.playlists.master.playlists[0].attributes.BANDWIDTH = 9e10;
```
gesinger: Why did these need to change?

test/videojs-contrib-hls.test.js
```diff
@@ -2695,6 +2711,9 @@ QUnit.module('HLS Integration', {
     this.tech = new (videojs.getTech('Html5'))({});
     this.clock = this.env.clock;
 
+    this.originalSelectPlaylist = Hls.STANDARD_PLAYLIST_SELECTOR;
+    Hls.STANDARD_PLAYLIST_SELECTOR = movingAverageBandwidthSelector();
```
gesinger: Isn't this already the default?

test/master-playlist-controller.test.js
```diff
@@ -70,6 +71,9 @@ QUnit.module('MasterPlaylistController', {
     this.requests = this.env.requests;
     this.mse = useFakeMediaSource();
 
+    this.originalSelectPlaylist = Hls.STANDARD_PLAYLIST_SELECTOR;
+    Hls.STANDARD_PLAYLIST_SELECTOR = movingAverageBandwidthSelector();
```
gesinger: Isn't this already the default?

test/videojs-contrib-hls.test.js
mjneil: Yes, but because of the way the closure works, if we don't do this between tests, each test will be using the moving average of all the previous tests

test/videojs-contrib-hls.test.js
mjneil: yes good catch

test/videojs-contrib-hls.test.js
mjneil: Since the selector now uses a moving average instead of the instantaneous, the bandwidth value that is being compared is the moving average of 1 and 4Mb, instead of 1. Needed to update the bandwidth of the playlist so it would still select

test/videojs-contrib-hls.test.js
mjneil: Yes, but because of the way the closure works, if we don't do this between tests, each test will be using the moving average of all the previous tests

test/master-playlist-controller.test.js
mjneil: Yes, but because of the way the closure works, if we don't do this between tests, each test will be using the moving average of all the previous tests

src/videojs-contrib-hls.js
mjneil: Yes good point

test/videojs-contrib-hls.test.js
```diff
@@ -673,7 +678,7 @@ QUnit.test('buffer checks are noops when only the master is ready', function(ass
 
   assert.strictEqual(1, this.requests.length, 'one request was made');
   assert.strictEqual(this.requests[0].url,
-                     absoluteUrl('manifest/media1.m3u8'),
+                     absoluteUrl('manifest/media2.m3u8'),
```
gesinger: Comment above says `force media1 to be requested`

test/videojs-contrib-hls.test.js
```diff
@@ -692,18 +697,16 @@ QUnit.test('selects a playlist below the current bandwidth', function(assert) {
...
-  // playlist 1 has a very low bitrate
-  this.player.tech_.hls.playlists.master.playlists[1].attributes.BANDWIDTH = 1;
   // but the detected client bandwidth is really low
-  this.player.tech_.hls.bandwidth = 10;
+  this.player.tech_.hls.bandwidth = 1;
 
   playlist = this.player.tech_.hls.selectPlaylist();
   assert.strictEqual(playlist,
-                     this.player.tech_.hls.playlists.master.playlists[1],
+                     this.player.tech_.hls.playlists.master.playlists[2],
```
gesinger: I believe the title of this test is no longer accurate. Maybe `selects a playlist below the current moving average bandwidth`

test/videojs-contrib-hls.test.js
```diff
@@ -1628,7 +1643,7 @@ QUnit.test('resets the switching algorithm if a request times out', function(ass
     type: 'application/vnd.apple.mpegurl'
   });
   openMediaSource(this.player, this.clock);
-  this.player.tech_.hls.bandwidth = 1e20;
+  this.player.tech_.hls.bandwidth = 289000;
```
gesinger: Might want a comment about why this number is the way it is

PR #1152, opened 2017-06-05
[ABR] Abort on progress estimating buffer drain

mjneil:

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -79,13 +79,12 @@ const getProgressStats = (progressEvent) => {
     roundTripTime: roundTripTime || 0
   };
 
-  if (progressEvent.lengthComputable) {
-    stats.bytesReceived = progressEvent.loaded;
-    // This can result in Infinity if stats.roundTripTime is 0 but that is ok
-    // because we should only use bandwidth stats on progress to determine when
-    // abort a request early due to insufficient bandwidth
-    stats.bandwidth = Math.floor((stats.bytesReceived / stats.roundTripTime) * 8 * 1000);
-  }
+  stats.bytesReceived = progressEvent.loaded;
+  // This can result in Infinity if stats.roundTripTime is 0 but that is ok
+  // because we should only use bandwidth stats on progress to determine when
+  // abort a request early due to insufficient bandwidth
+  stats.bandwidth = Math.floor((stats.bytesReceived / stats.roundTripTime) * 8 * 1000);
+
   return stats;
 };
 
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -334,9 +334,9 @@ const waitForCompletion = (activeXhrs, decrypter, doneFn) => {
 const handleProgress = (segment, progressFn) => (event) => {
   segment.stats = videojs.mergeOptions(segment.stats, getProgressStats(event));
 
-  // record the time that we receive the byte of data
-  if (!segment.stats.firstByteReceived && segment.stats.bytesReceived) {
-    segment.stats.firstByteReceived = Date.now();
+  // record the time that we receive the first byte of data
+  if (!segment.stats.firstByteRoundTripTime && segment.stats.bytesReceived) {
+    segment.stats.firstByteRoundTripTime = Date.now();
   }
 
   return progressFn(event, segment);
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -332,7 +332,13 @@ const waitForCompletion = (activeXhrs, decrypter, doneFn) => {
  * @param {Event} event - the progress event object from XMLHttpRequest
  */
 const handleProgress = (segment, progressFn) => (event) => {
-  segment.stats = getProgressStats(event);
+  segment.stats = videojs.mergeOptions(segment.stats, getProgressStats(event));
+
+  // record the time that we receive the byte of data
+  if (!segment.stats.firstByteReceived && segment.stats.bytesReceived) {
+    segment.stats.firstByteReceived = Date.now();
+  }
+
   return progressFn(event, segment);
 };
 
--- a/src/playlist-selectors.js
+++ b/src/playlist-selectors.js
@@ -242,3 +242,39 @@ export const movingAverageBandwidthSelector = function(decay) {
                           parseInt(safeGetComputedStyle(this.tech_.el(), 'height'), 10));
   };
 };
+
+export const minRebufferingSelector = function(master,
+                                               bandwidth,
+                                               timeUntilRebuffer,
+                                               segmentDuration) {
+  // get a list of playlists with bandwidth info and sort
+  const playlists = master.playlists.filter(
+    (media) => media.attributes && media.attributes.BANDWIDTH
+  ).sort((a, b) => a.attributes.BANDWIDTH - b.attributes.BANDWIDTH);
+
+  let minEstimatedRoundTripTime = Number.MAX_VALUE;
+  let bestPlaylist = null;
+
+  // Find the playlist with the highest bandwidth that the player can load from
+  // safely without needing to rebuffer
+  for (let i = playlists.length - 1; i >= 0; i--) {
+    const playlist = playlists[i];
+    const estimatedSegmentSize = segmentDuration * playlist.attributes.BANDWIDTH;
+    const estimatedRoundTripTime = estimatedSegmentSize / bandwidth;
+
+    if (estimatedRoundTripTime < minEstimatedRoundTripTime) {
+      minEstimatedRoundTripTime = estimatedRoundTripTime;
+      bestPlaylist = playlist;
+    }
+
+    if (estimatedRoundTripTime < timeUntilRebuffer) {
+      // we can switch to the playlist and avoid any rebuffering!
+      break;
+    }
+  }
+
+  return {
+    playlist: bestPlaylist,
+    roundTripTime: minEstimatedRoundTripTime
+  };
+};
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -720,20 +720,18 @@ export default class SegmentLoader extends videojs.EventTarget {
   }
 
   abortRequestEarly_(stats) {
-    // Do not consider aborting early if the player is paused
     if (this.hls_.tech_.paused() ||
-        // Or if the current playlist is on the lowestEnabledRendition
-        // TODO: Replace timeout with a boolean indicating whether this playlist is the
-        //       lowestEnabledRendition.
+        // Don't abort if the current playlist is on the lowestEnabledRendition
+        // TODO: Replace using timeout with a boolean indicating whether this playlist is
+        //       the lowestEnabledRendition.
         !this.xhrOptions_.timeout ||
-        // Or if the the current playlist does not have bandwidth information that can
-        // be used in the estimation calculations below
+        // Don't abort if we have no bandwidth information to estimate segment sizes
         !(this.playlist_.attributes && this.playlist_.attributes.BANDWIDTH)) {
       return false;
     }
 
     // Wait at least 1 second since the first byte of data has been received before
-    // using the calculated bandwidth from the] progress event to allow the bitrate
+    // using the calculated bandwidth from the progress event to allow the bitrate
     // to stabilize
     if (Date.now() - stats.firstByteRoundTripTime < 1000) {
       return false;
@@ -751,7 +749,7 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     // Only consider aborting early if the estimated time to finish the download
     // is larger than the estimated time until the player runs out of forward buffer
-    if (requestTimeRemaining < timeUntilRebuffer) {
+    if (requestTimeRemaining <= timeUntilRebuffer) {
       return false;
     }
 
@@ -761,25 +759,22 @@ export default class SegmentLoader extends videojs.EventTarget {
                              timeUntilRebuffer,
                              this.pendingSegment_.duration);
 
-    let minimumTimeSaving = 0.5
+    const timeSavedBySwitching = requestTimeRemaining - roundTripTime;
+
+    let minimumTimeSaving = 0.5;
 
     // If we are already rebuffering, increase the amount of variance we add to the
-    // potential round trip time of the new request so that we are not too agressive
-    // with switching to a playlst that might save us a fraction of a second.
+    // potential round trip time of the new request so that we are not too aggressive
+    // with switching to a playlist that might save us a fraction of a second.
     if (timeUntilRebuffer <= TIME_FUDGE_FACTOR) {
       minimumTimeSaving = 1;
     }
 
-    // Finally, don't abort early and trigger the downswitch if the playlist
-    // that can minimize the time spent rebuffering is the same playlist as the
-    // current
     if (!playlist ||
         playlist.uri === this.playlist_.uri ||
-        // or if the amount of time we are saving does not reach the minimum
-        roundTripTime <= minimumTimeSaving ||
-        // or that switching will not be more than 50% time saving than just sticking
-        // with the current request
-        roundTripTime > requestTimeRemaining / 2) {
+        timeSavedBySwitching <= minimumTimeSaving ||
+        // Don't switch if we cant get more than 50% savings
+        timeSavedBySwitching <= requestTimeRemaining * 0.5) {
       return false;
     }
 
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -717,6 +717,26 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  updateHandler_(event, segment) {
+    if (!this.pendingSegment_ || segment.requestId !== this.pendingSegment_.requestId) {
+      return;
+    }
+
+    const bandwidthAdjustment = Math.min(0.8, segment.stats.roundTripTime / 5000);
+
+    // TODO: Replace timeout with a boolean indicating whether this playlist is the lowestEnabledRendition
+    if (this.xhrOptions_.timeout &&
+        segment.stats.roundTripTime > 1000 &&
+        segment.stats.bandwidth < this.playlist_.attributes.BANDWIDTH * bandwidthAdjustment) {
+      this.bandwidth = segment.stats.bandwidth;
+      this.trigger('bandwidthupdate');
+      this.abort();
+      return;
+    }
+
+    this.trigger('progress');
+  }
+
   /**
    * load a specific segment from a request into the buffer
    *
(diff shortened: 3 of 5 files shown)


src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // lowestEnabledRendition, consider the possibility of aborting the current request
+    // early for an emergency downswitch
+    // TODO: Replace timeout with a boolean indicating whether this playlist is the
+    //       lowestEnabledRendition
+    if (!this.hls_.tech_.paused() && this.xhrOptions_.timeout) {
+      const firstByteReceived = Date.now() - stats.firstByteReceived;
+
+      // Wait at least 1 second before using the calculated bandwidth from the
+      // progress event to allow the bitrate to stableize
+      if (firstByteReceived > 1000) {
```
gesinger: Same re: returning early and removing indentation.

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // If the player is not paused and the current playlist is not the
+    // lowestEnabledRendition, consider the possibility of aborting the current request
+    // early for an emergency downswitch
+    // TODO: Replace timeout with a boolean indicating whether this playlist is the
+    //       lowestEnabledRendition
+    if (!this.hls_.tech_.paused() && this.xhrOptions_.timeout) {
+      const firstByteReceived = Date.now() - stats.firstByteReceived;
+
+      // Wait at least 1 second before using the calculated bandwidth from the
+      // progress event to allow the bitrate to stableize
```
gesinger: stabilize

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      if (firstByteReceived > 1000) {
+        const realBandwidth = stats.bandwidth;
```
gesinger: `realBandwidth` doesn't add much context to what we're identifying. Maybe `measuredBandwidth` or simply `bandwidth` (and the other bandwidths will be prefixed with the types they are).

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        const estimatedSize =
```
gesinger: Maybe `estimatedSegmentBits`?

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          this.pendingSegment_.duration * this.playlist_.attributes.BANDWIDTH;
```
gesinger: We should check for attributes and attributes.BANDWIDTH higher up and return if not available.

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        const currentTime = this.currentTime_();
+        const estimatedDelay =
```
gesinger: Maybe `estimatedRemainingSeconds`?

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          (estimatedSize - (stats.bytesReceived * 8)) / realBandwidth;
+        const buffered = this.buffered_();
+        const bufferedEnd = buffered.length ? buffered.end(buffered.length - 1) : 0;
+        const rebufferingDelay =
```
gesinger: Maybe `secondsUntilRebuffering`?

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          (bufferedEnd - currentTime) / this.hls_.tech_.playbackRate();
+
+        // consider aborting early if the estimated time to finish the download
+        // is larger than the estimated time until the player runs out of forward buffer
+        if (estimatedDelay > rebufferingDelay) {
```
gesinger: Do we want to allow for some variance here? I can see us going either way, since right now we have a variance on the bandwidth when choosing the playlist. That said, it might be useful to take note of that variance here. If there is an allowed variance when selecting a rendition, then we shouldn't consider one here. If there isn't, maybe we should think about having one here.

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          // get a list of playlists with a lower bandwidth than the current
+          const playlists = this.hls_.playlists.master.playlists.slice().filter(
```
gesinger: Since filter creates a new array, do we have to slice?

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+                     media.attributes.BANDWIDTH <= this.playlist_.attributes.BANDWIDTH;
+            }
+          ).sort((a, b) => a.attributes.BANDWIDTH - b.attributes.BANDWIDTH);
+
+          let newEstimatedDelay;
+          let safeBandwidth;
+
+          // Find the playlist with the highest bandwidth that the player can load from
+          // safely without needing to rebuffer
+          for (let i = playlists.length - 1; i >= 0; i--) {
```
gesinger: Instead of doing the logic here, I think we can move this over to `playlist-selectors` as a function that takes in the playlists, bandwidth, and forward buffer, and returns the playlist that should be able to provide a segment within the remaining buffer time. It can also make use of some of the player size information to ensure that we over select for our window size.

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+            const newPlaylist = playlists[i];
+            const newSegmentDuration =
+              newPlaylist.targetDuration || this.pendingSegment_.duration;
+            const newEstimatedSize =
+              newSegmentDuration * newPlaylist.attributes.BANDWIDTH;
+
+            newEstimatedDelay =
+              (newEstimatedSize) / (realBandwidth);
```
gesinger: Don't need parenthesis.

src/segment-loader.js
```diff
@@ -717,6 +717,94 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+          // Find the playlist with the highest bandwidth that the player can load from
+          // safely without needing to rebuffer
+          for (let i = playlists.length - 1; i >= 0; i--) {
+            const newPlaylist = playlists[i];
+            const newSegmentDuration =
+              newPlaylist.targetDuration || this.pendingSegment_.duration;
+            const newEstimatedSize =
+              newSegmentDuration * newPlaylist.attributes.BANDWIDTH;
+
+            newEstimatedDelay =
```
gesinger: Same thing re: naming with these variables (as above)

src/segment-loader.js
mjneil: I don't think there is any need to consider the player size for this because it is only checking playlists with a lower bandwidth than the current, which will most likely be a smaller size than the current, and even in the chance there are bigger ones with lower bandwidth, if it means that we can avoid any rebuffering, that benefit out weighs picking an appropriate resolution for the window size

gesinger: Makes sense.

src/playlist-selectors.js
```diff
@@ -242,3 +242,39 @@ export const movingAverageBandwidthSelector = function(decay) {
                           parseInt(safeGetComputedStyle(this.tech_.el(), 'height'), 10));
   };
 };
+
+export const minRebufferingSelector = function(master,
```
gesinger: Maybe `maxBandwidthWithoutRebufferingSelector`?

src/playlist-selectors.js
```diff
@@ -242,3 +242,39 @@ export const movingAverageBandwidthSelector = function(decay) {
...
+
+    if (estimatedRoundTripTime < timeUntilRebuffer) {
+      // we can switch to the playlist and avoid any rebuffering!
+      break;
+    }
+  }
+
+  return {
+    playlist: bestPlaylist,
+    roundTripTime: minEstimatedRoundTripTime
```
gesinger: Instead of providing roundTripTime and returning an object, can we just return null if there's no playlist (since it's possible the playlist will be null anyway).

This would also allow us to remove minEstimatedRoundTripTime and simplify the loop a bit.

src/segment-loader.js
```diff
@@ -718,78 +719,65 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // Wait at least 1 second before using the calculated bandwidth from the
+    // progress event to allow the bitrate to stabilize
+    if (firstByteReceived < 1000) {
+      return false;
+    }
+
+    const measuredBandwidth = stats.bandwidth;
+    const estimatedSegmentSize =
+      this.pendingSegment_.duration * this.playlist_.attributes.BANDWIDTH;
+    const currentTime = this.currentTime_();
```
gesinger: Only used once, might be simpler killing this variable

src/segment-loader.js
```diff
@@ -718,78 +719,65 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    const estimatedRoundTripTime =
```
gesinger: Isn't this remaining time?

src/segment-loader.js
```diff
@@ -718,78 +719,65 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      minRebufferingSelector(this.hls_.playlists.master,
+                             measuredBandwidth,
+                             timeUntilRebuffer,
+                             this.pendingSegment_.duration);
+
+    // Finally, only abort early and trigger the downswitch if the playlist
+    // that can minimize the time spent rebuffering is a different playlist than the
+    // current, and that switching will be less time than just waiting for the
+    // current request to complete.
+    if (playlist &&
```
gesinger: Minor, but since the other conditionals do the `return false` early, otherwise continue, may look nicer to check for cases we want to return with false instead, then the final return is true.

src/playlist-selectors.js
mjneil: It's not grabbing just the max bandwidth that wont have rebuffering. Its returning the playlist that will provide the minimum amount of time rebuffering. If there is a playlist where the amount of time rebuffering is 0, then it will be the max-bandwidth that can do it, but if none can bring that to 0, it will pick whichever has the least impact on rebuffering times

src/playlist-selectors.js
mjneil: this roundTripTime value is used in the return function. We can calculate it again when we get it back, but that seems pointless when we already have the value here.

src/playlist-selectors.js
gesinger: Makes sense 👍

src/media-segment-request.js
```diff
@@ -333,7 +332,13 @@ const waitForCompletion = (activeXhrs, decrypter, doneFn) => {
  * @param {Event} event - the progress event object from XMLHttpRequest
  */
 const handleProgress = (segment, progressFn) => (event) => {
-  segment.stats = getProgressStats(event);
+  segment.stats = videojs.mergeOptions(segment.stats, getProgressStats(event));
+
+  // record the time that we receive the byte of data
```
gesinger: the first byte

src/segment-loader.js
```diff
@@ -717,6 +718,80 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        // TODO: Replace timeout with a boolean indicating whether this playlist is the
+        //       lowestEnabledRendition.
+        !this.xhrOptions_.timeout ||
+        // Or if the the current playlist does not have bandwidth information that can
+        // be used in the estimation calculations below
+        !(this.playlist_.attributes && this.playlist_.attributes.BANDWIDTH)) {
+      return false;
+    }
+
+    const firstByteReceived = Date.now() - stats.firstByteReceived;
```
gesinger: `firstByteReceived` still seems to imply a boolean to me, not a time. Maybe `stats.timeFirstByteRecieved`? The variable is also the same name as the stat, which is a bit confusing. Might be better as `timeSinceFirstByteReceived`, or, Ssnce it is only used in one spot, might be easier to just remove the variable.

src/segment-loader.js
```diff
@@ -717,6 +718,80 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    const requestTimeRemaining =
+      (estimatedSegmentSize - (stats.bytesReceived * 8)) / measuredBandwidth;
+    const buffered = this.buffered_();
+    const bufferedEnd = buffered.length ? buffered.end(buffered.length - 1) : 0;
+    const timeUntilRebuffer =
+      (bufferedEnd - this.currentTime_()) / this.hls_.tech_.playbackRate();
+
+    // Only consider aborting early if the estimated time to finish the download
+    // is larger than the estimated time until the player runs out of forward buffer
+    if (requestTimeRemaining < timeUntilRebuffer) {
```
gesinger: Minor, but maybe `<=`

src/segment-loader.js
```diff
@@ -717,6 +718,80 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // current, or that switching will not be less time than just waiting for the
+    // current request to complete.
+    if (!playlist ||
+        playlist.uri === this.playlist_.uri ||
+        roundTripTime >= requestTimeRemaining) {
+      return false;
+    }
+
+    // set the bandwidth to that of the desired playlist
+    // (Being sure to scale by BANDWIDTH_VARIANCE and adding one so the
```
gesinger: add

src/segment-loader.js
mjneil: If the two values are equal, we might as well just stick with what we got and not abort, so I think strict is reasonable here

src/segment-loader.js
```diff
@@ -731,11 +732,10 @@ export default class SegmentLoader extends videojs.EventTarget {
       return false;
     }
 
-    const firstByteReceived = Date.now() - stats.firstByteReceived;
-
-    // Wait at least 1 second before using the calculated bandwidth from the
-    // progress event to allow the bitrate to stabilize
-    if (firstByteReceived < 1000) {
+    // Wait at least 1 second since the first byte of data has been received before
+    // using the calculated bandwidth from the] progress event to allow the bitrate
```
gesinger: ]

src/segment-loader.js
```diff
@@ -761,18 +761,30 @@ export default class SegmentLoader extends videojs.EventTarget {
                              timeUntilRebuffer,
                              this.pendingSegment_.duration);
 
+    let minimumTimeSaving = 0.5
+
+    // If we are already rebuffering, increase the amount of variance we add to the
+    // potential round trip time of the new request so that we are not too agressive
+    // with switching to a playlst that might save us a fraction of a second.
```
gesinger: playlist

src/segment-loader.js
```diff
@@ -761,18 +761,30 @@ export default class SegmentLoader extends videojs.EventTarget {
                              timeUntilRebuffer,
                              this.pendingSegment_.duration);
 
+    let minimumTimeSaving = 0.5
```
gesinger: minimumTimeSavings

src/segment-loader.js
```diff
@@ -761,18 +761,30 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+
+    // If we are already rebuffering, increase the amount of variance we add to the
+    // potential round trip time of the new request so that we are not too agressive
+    // with switching to a playlst that might save us a fraction of a second.
+    if (timeUntilRebuffer <= TIME_FUDGE_FACTOR) {
+      minimumTimeSaving = 1;
+    }
+
     // Finally, don't abort early and trigger the downswitch if the playlist
```
gesinger: I don't think this comment (or the others in the conditional) add anything to the code itself.

src/segment-loader.js
```diff
@@ -717,6 +719,91 @@ export default class SegmentLoader extends videojs.EventTarget {
     };
   }
 
+  abortRequestEarly_(stats) {
+    // Do not consider aborting early if the player is paused
```
gesinger: Can remove this comment (and possibly the one above checking BANDWIDTH)

src/segment-loader.js
```diff
@@ -717,6 +719,91 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        //       lowestEnabledRendition.
+        !this.xhrOptions_.timeout ||
+        // Or if the the current playlist does not have bandwidth information that can
+        // be used in the estimation calculations below
+        !(this.playlist_.attributes && this.playlist_.attributes.BANDWIDTH)) {
+      return false;
+    }
+
+    // Wait at least 1 second since the first byte of data has been received before
+    // using the calculated bandwidth from the] progress event to allow the bitrate
```
gesinger: ]

src/segment-loader.js
gesinger: I agree. I think we need `<=` to achieve that.

src/segment-loader.js
```diff
@@ -717,6 +719,91 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    const { playlist, roundTripTime } =
+      minRebufferingSelector(this.hls_.playlists.master,
+                             measuredBandwidth,
+                             timeUntilRebuffer,
+                             this.pendingSegment_.duration);
+
+    let minimumTimeSaving = 0.5
+
+    // If we are already rebuffering, increase the amount of variance we add to the
+    // potential round trip time of the new request so that we are not too agressive
```
gesinger: aggressive

src/segment-loader.js
```diff
@@ -717,6 +719,91 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      return false;
+    }
+
+    const { playlist, roundTripTime } =
+      minRebufferingSelector(this.hls_.playlists.master,
+                             measuredBandwidth,
+                             timeUntilRebuffer,
+                             this.pendingSegment_.duration);
+
+    let minimumTimeSaving = 0.5
```
gesinger: `minimumTimeSavings`

src/segment-loader.js
```diff
@@ -717,6 +719,91 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+
+    // If we are already rebuffering, increase the amount of variance we add to the
+    // potential round trip time of the new request so that we are not too agressive
+    // with switching to a playlst that might save us a fraction of a second.
+    if (timeUntilRebuffer <= TIME_FUDGE_FACTOR) {
+      minimumTimeSaving = 1;
+    }
+
+    // Finally, don't abort early and trigger the downswitch if the playlist
```
gesinger: Can remove the comments around the conditionals here, since they don't add much over reading the code itself.

src/segment-loader.js
mjneil: You're right, thinking backwards again

PR #1155, opened 2017-06-07
Do not reset segmentloaders when switching media groups

imbcmdth:
## Description
When switching media-groups we do not need to reset the SegmentLoaders. Doing so causes us to evict all audio (and/or video) data from the source buffers which causes poor user experience when switching renditions that reference different media-groups.

## Specific Changes proposed
* Fixed a bug in SegmentLoader where sometimes `abort` would leave it in a bad state
* Fixed a bug in SegmentLoader where `dispose` would leave the monitor buffer timer running
* Refactored `setupAudio` into several functions so that they can be combined into two different work-flows - `setupAudio` which does a hard-reset of the SegmentLoaders and `mediaGroupsChanged` which does not


test/master-playlist-controller.test.js
```diff
@@ -295,10 +301,11 @@ QUnit.test('fast quality change resyncs audio segment loader', function(assert)
   assert.equal(this.requests.length, 1, 'one request');
   masterPlaylistController.fastQualityChange_();
   assert.equal(this.requests.length, 2, 'added a request for new media');
-  assert.equal(resets, 0, 'does not reset the audio segment loader yet');
+  assert.equal(resyncs, 0, 'does not resyncs the audio segment loader yet');
```
gesinger: does not resync

test/master-playlist-controller.test.js
```diff
@@ -281,11 +281,17 @@ QUnit.test('fast quality change resyncs audio segment loader', function(assert)
...
-  masterPlaylistController.audioSegmentLoader_.resetLoader = () => resets++;
+  masterPlaylistController.audioSegmentLoader_.resetLoader = function() {
+    resets++;
+    this.resyncLoader();
+  };
+
+  masterPlaylistController.audioSegmentLoader_.resyncLoader = () => resyncs++;
   masterPlaylistController.fastQualityChange_();
-  assert.equal(resets, 0, 'does not reset the audio segment loader when media same');
+  assert.equal(resyncs, 0, 'does not resyncs the audio segment loader when media same');
```
gesinger: does not resync

PR #720, opened 2016-05-31
use can handle source options to override mode 

BrandonOCasey:
## Description

Once video.js cuts a release that includes passing source handler options to canHandleSource we can merge this PR so that the mode can be overridden more easily.

This CANNOT BE MERGED YET!
## Specific Changes proposed

Implement video.js canHandleSource API change
## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] Reviewed by Two Core Contributors

PR #1156, opened 2017-06-07
Allow overrideNative to be set as a player-level option

imbcmdth:
## Description
Right now, `overrideNative` must be set globally on the `videojs.options` object. This is especially dangerous because of https://github.com/videojs/video.js/issues/4391. This PR lets set the `overrideNative` property as a player option so that you never touch the global videojs options again (which is probably for the best)!

## Specific Changes proposed
* Merge options earlier so that we can use them in the `canPlayType` and `canHandleSource` functions

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -321,7 +321,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.seekable_ = videojs.createTimeRanges();
     this.hasPlayed_ = () => false;
 
-    this.syncController_ = new SyncController();
+    this.syncController_ = new SyncController(options);
     this.segmentMetadataTrack_ = tech.addRemoteTextTrack({
       kind: 'metadata',
       label: 'segment-metadata'
@@ -354,18 +354,18 @@ export class MasterPlaylistController extends videojs.EventTarget {
       new SegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
         segmentMetadataTrack: this.segmentMetadataTrack_,
         loaderType: 'main'
-      }));
+      }), options);
 
     // alternate audio track
     this.audioSegmentLoader_ =
       new SegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
         loaderType: 'audio'
-      }));
+      }), options);
 
     this.subtitleSegmentLoader_ =
       new VTTSegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
         loaderType: 'vtt'
-      }));
+      }), options);
 
     this.setupSegmentLoaderListeners_();
 
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -329,7 +329,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.decrypter_ = worker(Decrypter);
 
-    let segmentLoaderOptions = {
+    const segmentLoaderSettings = {
       hls: this.hls_,
       mediaSource: this.mediaSource,
       currentTime: this.tech_.currentTime.bind(this.tech_),
@@ -351,19 +351,19 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // setup segment loaders
     // combined audio/video or just video when alternate audio track is selected
     this.mainSegmentLoader_ =
-      new SegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
+      new SegmentLoader(videojs.mergeOptions(segmentLoaderSettings, {
         segmentMetadataTrack: this.segmentMetadataTrack_,
         loaderType: 'main'
       }), options);
 
     // alternate audio track
     this.audioSegmentLoader_ =
-      new SegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
+      new SegmentLoader(videojs.mergeOptions(segmentLoaderSettings, {
         loaderType: 'audio'
       }), options);
 
     this.subtitleSegmentLoader_ =
-      new VTTSegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
+      new VTTSegmentLoader(videojs.mergeOptions(segmentLoaderSettings, {
         loaderType: 'vtt'
       }), options);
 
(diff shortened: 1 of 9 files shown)


src/videojs-contrib-hls.js
```diff
@@ -214,23 +214,23 @@ class HlsHandler extends Component {
...
-    }
-
     this.tech_ = tech;
     this.source_ = source;
     this.stats = {};
     this.ignoreNextSeekingEvent_ = false;
 
     // handle global & Source Handler level options
-    this.options_ = videojs.mergeOptions(videojs.options.hls || {}, options.hls);
+    this.options_ = options.hls;
```
gkatsev: if you pass `options` as the second argument to `super`, the Component constructor will take care of setting `this.options_` correctly.

src/videojs-contrib-hls.js
```diff
@@ -569,17 +566,21 @@ class HlsHandler extends Component {
...
-      if (videojs.options.hls &&
-          videojs.options.hls.mode &&
-          videojs.options.hls.mode !== mode) {
+      if (settings.hls &&
+          settings.hls.mode &&
+          settings.hls.mode !== mode) {
         return false;
       }
-      return HlsSourceHandler.canPlayType(srcObj.type);
+      return HlsSourceHandler.canPlayType(srcObj.type, settings);
```
gesinger: It seems that `settings` and `options` are used interchangeably throughout the code (for instance, here `settings` are passed as the `options` parameter). I think it may make more sense to contextualize more. Maybe something like `mergedOptions` after merged, or `playerOptions` for player specific ones.

gkatsev: The general convention we've had is that `options` is the name of the arguments passed in and `settings` is the merged options to be used internally

gkatsev: Not sure how consistent we've been in our projects but that was the idea.

gesinger: If that is the case, shouldn't we just call them `mergedOptions`?

imbcmdth: Changed to `localOptions` because `mergedOptions` and `videojs.mergeOptions(..)` is too confusing!

src/master-playlist-controller.js
```diff
@@ -354,18 +354,18 @@ export class MasterPlaylistController extends videojs.EventTarget {
       new SegmentLoader(videojs.mergeOptions(segmentLoaderOptions, {
         segmentMetadataTrack: this.segmentMetadataTrack_,
         loaderType: 'main'
-      }));
+      }), options);
```
mjneil: Why not just merge `options` into the first parameter instead of having two parameters that are both option objects?

imbcmdth: Because that are distinctly different "options" one comes from outside the project and the other is more of an object with functions necessary to the correct functioning of SegmentLoader.

Merging them feels very wrong.

mjneil: Maybe just rename `segmentLoaderOptions` to `segmentLoaderSettings` then

src/segment-loader.js
```diff
@@ -51,20 +51,18 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
  * @extends videojs.EventTarget
  */
 export default class SegmentLoader extends videojs.EventTarget {
-  constructor(options) {
+  constructor(settings, options) {
```
mjneil: not necessary but maybe make `options = {}` optional. That way you don't need to add the empty objects in the test files. Maybe in `SyncController` as well

gesinger:
Closed https://github.com/videojs/videojs-contrib-hls/pull/720 in favor of this PR

test/loader-common.js
```diff
@@ -43,17 +43,17 @@ export const LoaderCommonHooks = {
...
  * for use in constructing a segment loader. This function should be called with the QUnit
  * test environment the loader will be constructed in for proper this reference.
  *
- * @param {Object} options
- *        custom options for the loader
+ * @param {Object} settings
+ *        custom settings for the loader
  * @return {Object}
- *         Options object contiaing custom options merged with defaults
+ *         settings object contiaing custom settings merged with defaults
```
gesinger: Was there before, but `containing`

test/loader-common.js
```diff
@@ -65,31 +65,31 @@ export const LoaderCommonSettings = function(options) {
     decrypter: this.decrypter
   };
 
-  return videojs.mergeOptions(settings, options);
+  return videojs.mergeOptions(commonSettings, settings);
```
gesinger: Can get rid of `commonSettings` entirely if we want to do it in one statement

test/loader-common.js
```diff
@@ -65,31 +65,31 @@ export const LoaderCommonSettings = function(options) {
...
 
 /**
  * Sets up a QUnit module to run tests that should be run on all segment loader types.
  * Currently only two types, SegmentLoader and VTTSegmentLoader.
  *
  * @param {function(new:SegmentLoader|VTTLoader, Object)} LoaderConstructor
- *        Constructor for segment loader. Takes one parameter, an options object
- * @param {Object} loaderOptions
- *        Custom options to merge with defaults for the provided loader constructor
+ *        Constructor for segment loader. Takes one parameter, an settings object
```
gesinger: a

test/videojs-contrib-hls.test.js
```diff
@@ -1768,7 +1768,33 @@ QUnit.test('has no effect if native HLS is available', function(assert) {
...
+  player.src({
+    src: 'http://example.com/manifest/master.m3u8',
+    type: 'application/x-mpegURL'
+  });
+
+  assert.ok(player.tech_.hls, 'did load hls tech');
+  player.dispose();
+});
+
+QUnit.test('loads if native HLS is available and override is set locally', function(assert) {
```
gesinger: I think this one is global.

test/videojs-contrib-hls.test.js
```diff
@@ -1768,7 +1768,33 @@ QUnit.test('has no effect if native HLS is available', function(assert) {
   player.dispose();
 });
 
-QUnit.test('loads if native HLS is available and override is set', function(assert) {
+QUnit.test('loads if native HLS is available and override is set globally', function(assert) {
```
gesinger: I think this one is local.

PR #1166, opened 2017-06-15
Add hls usage tracking events

zhuangs:
## Description
Adding hls usage tracking events to track hls features

## Requirements Checklist
  - [ ] Docs/guides updated


src/rendition-mixin.js
```diff
@@ -31,7 +31,7 @@ const enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
     // Ensure the outside world knows about our changes
     changePlaylistFn();
   }
-
+  loader.trigger('hlsrenditiondisabled');
```
mjneil: This should be moved into the if block just above.
```
rep.enabled(false);
rep.enabled(false);
```
will cause two triggers, but I don't think it makes sense to trigger the event twice when its not switching from enabled/disabled

src/videojs-contrib-hls.js
```diff
@@ -208,11 +208,19 @@ class HlsHandler extends Component {
...
           get: () => {
             videojs.log.warn('player.hls is deprecated. Use player.tech_.hls instead.');
+            // fired the uasge event when player.hls is accessed
+            tech.trigger({type: 'usage', name: 'hls-player-access'});
             return this;
           }
         });
       }
     }
+    // Object.defineProperty(tech, 'hls', {
```
mjneil: can delete these comments

src/master-playlist-controller.js
```diff
@@ -563,6 +584,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.onSyncInfoUpdate_();
     });
 
+    this.mainSegmentLoader_.on('hlstimestampoffset', () => {
+      this.trigger({type: 'usage', name: 'hls-timestamp-offset'});
```
mjneil: should trigger this on `this.tech_`

src/master-playlist-controller.js
```diff
@@ -1417,6 +1448,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.mainSegmentLoader_.mimeType(mimeTypes[0]);
     if (mimeTypes[1]) {
       this.audioSegmentLoader_.mimeType(mimeTypes[1]);
+      // fired when video and audio is demuxed
+      this.tech_.trigger({type: 'usage', name: 'hls-demuxed'});
```
mjneil: We want this event to mean that video and audio is demuxed by _default_ instead of when alternate audio is present. However, it would still be nice to know when alternate audio is present. I suggest changing this event to `hls-alternate-audio`.

For the `hls-demuxed` event, I think a good place to put it is within `fillAudioTracks_`. Since the `DEFAULT` attribute for media tags is optional, the best way we can detect that video and audio are demuxed by default is if all audio renditions have a `URI`, since the lack of uri means the audio is muxed with the video for that rendition. `fillAudioTracks_` already has to loop through all the audio groups, so I suggest checking each one within the loop for a uri, and after the loop if they all had it, fire the event

src/rendition-mixin.js
zhuangs: You are right about that. But there is one more problem with this trigger, it seems I can't trigger event on the `loader` passed in. It says `loader.trigger is not a function`, isn't it supposed to be used like a regular playlist loader?

src/master-playlist-controller.js
```diff
@@ -419,6 +419,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
       this.trigger('audioupdate');
       this.trigger('selectedinitialmedia');
+
+      // fired when a user selects an alternate audio stream in HLS
+      this.tech_.trigger({type: 'usage', name: 'hls-audio-change'});
```
mjneil: I think it makes sense to move this event to `videojs-contrib-hls.js` in the `audioTrackChange_` method

src/videojs-contrib-hls.js
zhuangs: trying to do the detection of accessing `player.tech_.hls` here(fire `hls-tech-access` event when `player.tech_.hls` is accessed). was planning to define a get method for `tech_.hls` for the detection purpose but it seems that doesn't work. Do you have any good idea about how to detect the `hls-tech-access`?

src/rendition-mixin.js
mjneil: You should be able to trigger on the loader yes.

src/videojs-contrib-hls.js
mjneil: It didn't work because `tech.hls` is already a property (gets attached on `HlsHandler` creation). This event is a bit weird because we'd have to define the getter before creation, but that means an code within the project itself that accesses `tech_.hls` would fire that event. For now we can hold off on having an `hls-tech-access` event and stick with just the `hls-player-access`

src/master-playlist-controller.js
```diff
@@ -431,6 +433,22 @@ export class MasterPlaylistController extends videojs.EventTarget {
         return;
       }
 
+      for(var i = 0; i < updatedPlaylist.segments; i++) {
```
mjneil: We should move these checks out of the `loadedplaylist` listener and into `loadedmetadata` listener. `laodedplaylist` will be fired for every playlist refresh in a live stream. 

Also, since `loadedmetadata` is only fired once for the first media playlist we load, moving this there would remove the need to have extra state for `isAes_ and `isFmp4`.

I also think it would be a good idea to move these loops into helper functions either on the `PlaylistLoader` or in `playlist.js`

src/rendition-mixin.js
```diff
@@ -30,8 +30,9 @@ const enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
   if (enable !== currentlyEnabled && !blacklisted) {
     // Ensure the outside world knows about our changes
     changePlaylistFn();
+    // fired when a rendition is enabled/disabled in HLS
+    loader.trigger('hlsrenditiondisabled');
```
mjneil: since we control this event and its being triggered on the `PlaylistLoader` and not the `tech`, we don't need to have the `hls` prefix. But the `usage` event fired on the `tech` should still have the `hls-` prefix

src/segment-loader.js
```diff
@@ -938,6 +938,8 @@ export default class SegmentLoader extends videojs.EventTarget {
     if (segmentInfo.timestampOffset !== null &&
         segmentInfo.timestampOffset !== this.sourceUpdater_.timestampOffset()) {
       this.sourceUpdater_.timestampOffset(segmentInfo.timestampOffset);
+      // fired when a timestamp offset is set in HLS (can also identify discontinuities)
+      this.trigger('hlstimestampoffset');
```
mjneil: since we control this event and its being triggered on the `SegmentLoader` and not the `tech`, we don't need to have the `hls` prefix. But the `usage` event fired on the `tech` should still have the `hls-` prefix

src/playback-watcher.js
```diff
@@ -344,6 +344,9 @@ export default class PlaybackWatcher {
 
     // only seek if we still have not played
     this.tech_.setCurrentTime(nextRange.start(0) + Ranges.TIME_FUDGE_FACTOR);
+
+    // fired when a gap in the buffer is skipped in HLS
+    this.tech_.trigger({type: 'usage', name: 'hls-gap-skip'});
```
mjneil: We should update the 3 other `trigger` calls in `PlaybackWatcher` into `usage` events `hls-live-resync` `hls-video-underflow` and `hls-unknown-waiting`

test/playback-watcher.test.js
```diff
@@ -137,6 +166,8 @@ QUnit.test('skips over gap in Chrome due to video underflow', function(assert) {
 
   assert.equal(seeks.length, 1, 'one seek');
   assert.equal(seeks[0], 13, 'player seeked to current time');
+  // ?? not sure if it's correct that there is no skipped gap
+  assert.equal(hlsgapskip, 0, 'there is no skipped gap');
```
mjneil: for this test you'll want to be checking for the `hls-video-underflow` event. The gap skip event wont fire in the video underflow scenario

mjneil:
Let's add an event for `hls-playlist-cue-tags` here https://github.com/videojs/videojs-contrib-hls/blob/master/src/master-playlist-controller.js#L303

and an `hls-error-reload` around here https://github.com/videojs/videojs-contrib-hls/blob/master/src/reload-source-on-error.js#L51

src/playlist-loader.js
```diff
@@ -274,6 +274,34 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     return (loader.master.playlists.filter(isEnabled).length === 1);
   };
 
+  /**
+   * Returns whether the current playlist is an AES encrypted HLS stream
+   *
+   * @return {Boolean} true if it's an AES encrypted HLS stream
+   */
+  loader.isAes_ = function(media) {
```
mjneil: I think these would be better suited as utility functions in`playlist.js`. They don't rely on any state of the loader so lets not put them on it

src/playlist.js
```diff
@@ -411,12 +411,42 @@ export const isEnabled = function(playlist) {
   return (!playlist.disabled && !blacklisted);
 };
 
+/**
+ * Returns whether the current playlist is an AES encrypted HLS stream
+ *
+ * @return {Boolean} true if it's an AES encrypted HLS stream
+ */
+export const isAes = function(media) {
```
mjneil: these two methods could use a simple unit test each

src/reload-source-on-error.js
```diff
@@ -22,6 +22,8 @@ const initPlugin = function(player, options) {
   let seekTo = 0;
   let localOptions = videojs.mergeOptions(defaultOptions, options);
 
+  player.tech_.trigger({type: 'usage', name: 'hls-first-initialized'});
```
mjneil: maybe `hls-error-reload-initialized`

test/master-playlist-controller.test.js
```diff
@@ -1184,16 +1192,103 @@ QUnit.test('correctly sets alternate audio track kinds', function(assert) {
                'spanish track\'s kind is "alternative"');
 });
 
+QUnit.test('trigger events when an AES is detected', function(assert) {
+  let hlsaes = 0;
+  let playlistCopy = Hls.Playlist;
```
mjneil: You need to save a reference to the original `Hls.Playlist.isAes` function and restore it at the end instead of the `Hls.Playlist` object. since `Hls.Playlist` is an object, `playlistCopy` is just a reference to the object and not an actual copy. Same goes for `isFmp4`

test/playlist.test.js
```diff
@@ -633,6 +633,29 @@ QUnit.test('seekable and playlistEnd use available sync points for calculating',
...
+
+QUnit.test('determine if playlist is an AES encrypted HLS stream', function(assert) {
+  let media = '#EXTM3U\n' +
+              '#EXT-X-TARGETDURATION:15\n' +
+              '#EXT-X-KEY:METHOD=AES-128,URI="http://example.com/keys/key.php"\n' +
+              '#EXTINF:2.833,\n' +
+              'http://example.com/000001.ts\n' +
+              '#EXT-X-ENDLIST\n';
+
+  assert.ok(Playlist.isAes(media), 'media is an AES encrypted HLS stream');
```
mjneil: You either need to pass in an object representation of a playlist like in the test above or create a loader like the test 2 below and respond to the xhr request with this string. These tests should be failing, but passed in traviscli (and probably locally for you) because of the overriding of the `isAes` and `isFmp4` methods you did in `master-playlist-controller.test.js`

src/reload-source-on-error.js
```diff
@@ -22,6 +22,8 @@ const initPlugin = function(player, options) {
   let seekTo = 0;
   let localOptions = videojs.mergeOptions(defaultOptions, options);
 
+  player.tech_.trigger({type: 'usage', name: 'hls-error-reload-initialized'});
```
mjneil: we should add tests for the error reload events.

test/master-playlist-controller.test.js
```diff
@@ -1184,16 +1192,103 @@ QUnit.test('correctly sets alternate audio track kinds', function(assert) {
                'spanish track\'s kind is "alternative"');
 });
 
+QUnit.test('trigger events when an AES is detected', function(assert) {
+  let hlsaes = 0;
+  let playlistCopy = Hls.Playlist.isAes;
```
mjneil: `isAesCopy`

test/master-playlist-controller.test.js
```diff
@@ -1184,16 +1192,103 @@ QUnit.test('correctly sets alternate audio track kinds', function(assert) {
...
+  this.standardXHRResponse(this.requests.shift());
+  this.masterPlaylistController.mediaSource.trigger('sourceopen');
+
+  assert.equal(hlsaes, 1, 'an AES HLS stream is detected');
+  Hls.Playlist.isAes = playlistCopy;
+});
+
+QUnit.test('trigger events when an fMP4 stream is detected', function(assert) {
+  let hlsfmp4 = 0;
+  let playlistCopy = Hls.Playlist.isFmp4;
```
mjneil: `isFmp4Copy`

test/reload-source-on-error.test.js
```diff
@@ -25,6 +29,25 @@ QUnit.module('ReloadSourceOnError', {
       return this.tech;
     };
 
+    this.player.tech_ = () => {
+      return this.tech;
+    };
+
+    this.player.tech_.trigger = (event) => {
```
mjneil: while functionally this works, it doesn't make much sense to use `trigger()` to also get the values of whether other events have been triggered. 

You can mock the `tech` in a similar way the player is so that it gets event handling automatically.

```
this.tech = new videojs.EventTarget();
this.tech.currentSource_ = {
  src: 'thisisasource.m3u8',
  type: 'doesn\'t/matter'
};
this.player.tech = () => this.tech;
this.player.tech_ = this.tech;
```

With this you should be able to listen on the tech for the events you want within the individual test runs

test/rendition-mixin.test.js
```diff
@@ -62,6 +62,14 @@ const makeMockHlsHandler = function(playlistOptions) {
       }
     }
   };
+  let triggered = false;
+
+  hlsHandler.playlists.trigger = function(event) {
```
mjneil: same concept here regarding trigger

test/master-playlist-controller.test.js
```diff
@@ -1184,16 +1192,103 @@ QUnit.test('correctly sets alternate audio track kinds', function(assert) {
...
+      demuxed++;
+    }
+  });
+
+  this.masterPlaylistController.mediaSource.trigger('sourceopen');
+  // master
+  this.standardXHRResponse(this.requests.shift());
+  // media
+  this.standardXHRResponse(this.requests.shift());
+  // init segment
```
gesinger: I think this is detectable before many of these requests are fulfilled

test/playback-watcher.test.js
```diff
@@ -137,6 +166,7 @@ QUnit.test('skips over gap in Chrome due to video underflow', function(assert) {
 
   assert.equal(seeks.length, 1, 'one seek');
   assert.equal(seeks[0], 13, 'player seeked to current time');
+  assert.equal(hlsvideounderflow, 1, 'triggered an video underflow event');
```
gesinger: a

test/playback-watcher.test.js
```diff
@@ -356,20 +386,28 @@ QUnit.test('fires notifications when activated', function(assert) {
...
     };
   };
-  this.player.tech_.on('liveresync', () => liveresync++);
-  this.player.tech_.on('videounderflow', () => videounderflow++);
+  this.player.tech_.on('usage', (event) => {
+    if (event.name === 'hls-live-resync') {
+      hlsliveresync++;
+    }
+  });
+  this.player.tech_.on('usage', (event) => {
```
gesinger: Can combine both of these into one listener if you want

test/playlist.test.js
```diff
@@ -633,6 +633,58 @@ QUnit.test('seekable and playlistEnd use available sync points for calculating',
...
+    'http://example.com/000001.ts\n' +
+    '#EXT-X-ENDLIST\n'
+  );
+
+  media = loader.media();
+
+  assert.ok(Playlist.isAes(media), 'media is an AES encrypted HLS stream');
+});
+
+QUnit.test('determine if playlist contains fmp4 segment', function(assert) {
```
gesinger: an fmp4 segment

test/playlist.test.js
```diff
@@ -633,6 +633,58 @@ QUnit.test('seekable and playlistEnd use available sync points for calculating',
...
+  let media;
+  let loader = new PlaylistLoader('video/fmp4.m3u8', this.fakeHls);
+
+  loader.load();
+  this.requests.shift().respond(200, null,
+                                '#EXTM3U\n' +
+                                '#EXT-X-MAP:URI="main.mp4",BYTERANGE="720@0"\n' +
+                                '#EXTINF:10,\n' +
+                                '0.ts\n' +
```
gesinger: This is a bit deceptive. Our check for fmp4 is a little improper right now, in that we only check for `#EXT-X-MAP`, however, for good measure we should still have `0.mp4` here.

test/reload-source-on-error.test.js
```diff
@@ -103,10 +100,38 @@ QUnit.test('doesn\'t seek to currentTime in live', function(assert) {
...
+  });
+
+  this.player.tech_.on('usage', (event) => {
+    if (event.name === 'hls-error-not-reload') {
+      hlserrornotload++;
+    }
+  });
+
+  assert.equal(hlserrorreloadinitialized, 0, 'the plugin has not been initialized');
+  assert.equal(hlserrorload, 0, 'there is no source was set');
```
gesinger: no source was set

test/reload-source-on-error.test.js
```diff
@@ -103,10 +100,38 @@ QUnit.test('doesn\'t seek to currentTime in live', function(assert) {
...
+  assert.equal(hlserrornotload, 0, 'not reload event has not been triggered');
+
   this.player.reloadSourceOnError();
   this.player.trigger('error', -2);
   this.player.trigger('loadedmetadata');
 
+  assert.equal(hlserrorreloadinitialized, 1, 'the plugin has been initialized');
+  assert.equal(hlserrorload, 1, 'src was set after an error has caused the reload');
```
gesinger: after an error caused

test/segment-loader.test.js
```diff
@@ -212,6 +217,7 @@ QUnit.module('SegmentLoader', function(hooks) {
       this.updateend();
       this.clock.tick(1);
 
+      assert.equal(hlstimestampoffset, 0, 'there is no timestampOffset was set');
```
gesinger: no hls-timestamp-offset event was fired

test/segment-loader.test.js
```diff
@@ -222,6 +228,7 @@ QUnit.module('SegmentLoader', function(hooks) {
       // verify stats
       assert.equal(loader.mediaBytesTransferred, 20, '20 bytes');
       assert.equal(loader.mediaRequests, 2, '2 requests');
+      assert.equal(hlstimestampoffset, 1, 'there is one timestampOffset was set');
```
gesinger: an hls-timestamp-offset event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -129,17 +129,26 @@ QUnit.module('HLS', {
...
+  this.player.tech_.on('usage', (event) => {
+    if (event.name === 'hls-player-access') {
+      hlsplayeraccess++;
+    }
+  });
+
   videojs.log.warn = (text) => {
     warning = text;
   };
+  assert.equal(hlsplayeraccess, 0, 'there is no player.hls access');
```
gesinger: no hls-player-access event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -129,17 +129,26 @@ QUnit.module('HLS', {
...
   let hls = this.player.hls;
 
+  assert.equal(hlsplayeraccess, 1, 'player.hls is accessed once');
```
gesinger: an hls-player-access event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -1120,6 +1135,7 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
               'no media is initially set');
 
   assert.equal(blacklistplaylist, 0, 'there is no blacklisted playlist');
+  assert.equal(hlsrenditionblacklisted, 0, 'there is no blacklised rendition');
```
gesinger: no hls-rendition-blacklisted event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -1130,6 +1146,7 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
               'Problem encountered with the current HLS playlist. HLS playlist request error at URL: media.m3u8 Switching to another playlist.',
               'log generic error message');
   assert.equal(blacklistplaylist, 1, 'there is one blacklisted playlist');
+  assert.equal(hlsrenditionblacklisted, 1, 'there is one blacklisted rendition');
```
gesinger: an hls-rendition-blacklisted event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -1145,6 +1162,7 @@ QUnit.test('playlist 404 should blacklist media', function(assert) {
               'log specific error message for final playlist');
   assert.equal(retryplaylist, 1, 'retried final playlist for once');
   assert.equal(blacklistplaylist, 1, 'there is one blacklisted playlist');
+  assert.equal(hlsrenditionblacklisted, 1, 'there is one blacklised rendition');
```
gesinger: an hls-rendition-blacklisted event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -2492,6 +2518,7 @@ QUnit.test('when mediaGroup changes enabled track should not change', function(a
   let mpc = hls.masterPlaylistController_;
   let audioTracks = this.player.audioTracks();
 
+  assert.equal(hlsaudiochange, 0, 'there is no audio track change');
```
gesinger: no hls-audio-change event was fired

test/videojs-contrib-hls.test.js
```diff
@@ -2531,6 +2558,7 @@ QUnit.test('when mediaGroup changes enabled track should not change', function(a
   assert.notEqual(oldMediaGroup, hls.playlists.media().attributes.AUDIO, 'selected a new playlist');
   audioTracks = this.player.audioTracks();
 
+  assert.equal(hlsaudiochange, 1, 'there is one audio track change');
```
gesinger: an hls-audio-change event was fired

test/master-playlist-controller.test.js
```diff
@@ -1127,18 +1127,26 @@ QUnit.test('calls to update cues on media when no master', function(assert) {
 
 QUnit.test('respects useCueTags option', function(assert) {
   let origHlsOptions = videojs.options.hls;
+  let hlsplaylistcuetags = 0;
```
gesinger: For this and other event counters, I think it is more readable with camelCase, e.g., `hlsPlaylistCueTags` or even `hlsPlaylistCueTagsEvents` to be more precise about what is being tracked

test/master-playlist-controller.test.js
zhuangs: the demuxed event is triggered after the video segment request is fulfilled, should i remove the audio media request or just leave it there?

mjneil: the `hls-demuxed` event is triggered in `fillAudioTracks_` which is called after `loadedmetadata` handler which happens after the first non-master playlist request, so there should be no need to wait for a video or audio segment request to be fulfilled before seeing this event

src/rendition-mixin.js
```diff
@@ -30,8 +30,12 @@ const enableFunction = (loader, playlistUri, changePlaylistFn, enable) => {
   if (enable !== currentlyEnabled && !blacklisted) {
     // Ensure the outside world knows about our changes
     changePlaylistFn();
+    if (currentlyEnabled) {
```
mjneil: minor but can we check `if(enable)` and switch the order of the triggers. At first glance I thought these were wrong because `currentlyEnabled` is true but we trigger `disabled`

test/reload-source-on-error.test.js
```diff
@@ -103,10 +100,38 @@ QUnit.test('doesn\'t seek to currentTime in live', function(assert) {
...
+
+  this.player.tech_.on('usage', (event) => {
+    if (event.name === 'hls-error-reload-canceled') {
+      hlsErrorReloadCanceledEvents++;
+    }
+  });
+
+  assert.equal(hlsErrorReloadInitializedEvents, 0, 'the plugin has not been initialized');
+  assert.equal(hlsErrorReloadEvents, 0, 'no source was set');
+  assert.equal(hlsErrorReloadCanceledEvents, 0, 'not reload event has not been triggered');
```
mjneil: `reload canceled event has ...`

README.md
```diff
@@ -476,6 +479,51 @@ are triggered on the player object.
...
+| Name          | Description   |
+| ------------- | ------------- |
+| hls-gap-skip  | player skipped a gap in the buffer |
+| hls-player-access | player.hls was accessed |
+| hls-audio-change | a user selected an alternate audio stream |
+| hls-rendition-disabled | a rendition was disabled |
+| hls-rendition-enabled | a rendition was enabled |
+| hls-rendition-blacklisted | a rendition was blacklisted |
+| hls-timestamp-offset | a timestamp offset was set in HLS (can identify discontinuities) |
+| hls-playlist-cue-tags | a playlist used cue tags (see useCueTags for details) |
```
mjneil: this should be a presence stat

src/master-playlist-controller.js
```diff
@@ -409,6 +409,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.fillSubtitleTracks_();
       this.setupSubtitles();
 
+      this.presenceUsage(this.master(), media);
```
mjneil: we should make this private `presenceUsage_`

src/master-playlist-controller.js
```diff
@@ -536,8 +538,62 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    this.masterPlaylistLoader_.on('renditionenabled', () => {
+      this.tech_.trigger({type: 'usage', name: 'hls-rendition-enabled'});
+    });
   }
 
+  presenceUsage(master, media) {
+    let mediaGroups = master.mediaGroups || {};
+    let defaultDemuxed = true;
+
+    if (!mediaGroups ||
```
mjneil: this whole `if` block is not needed

src/master-playlist-controller.js
```diff
@@ -536,9 +538,60 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+
+    if (Hls.Playlist.isAes(media)) {
+      this.tech_.trigger({type: 'usage', name: 'hls-aes'});
+    }
+
+    if (Hls.Playlist.isFmp4(media)) {
+      this.tech_.trigger({type: 'usage', name: 'hls-fmp4'});
+    }
+
+    if (audioGroupKeys.length && mediaGroups.AUDIO[audioGroupKeys[0]].length > 1) {
```
mjneil: The properties within `mediaGroups.AUDIO` are objets instead of arrays so this check will fail. You could check `Object.keys(mediaGroups.AUDIO[audioGroupKeys[0]]).length`. Could also make use of the audio group loop you do above for demuxed and use a counter to see if there is more than 1 alternate audio

src/master-playlist-controller.js
```diff
@@ -536,9 +538,61 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+    if (Hls.Playlist.isAes(media)) {
+      this.tech_.trigger({type: 'usage', name: 'hls-aes'});
+    }
+
+    if (Hls.Playlist.isFmp4(media)) {
+      this.tech_.trigger({type: 'usage', name: 'hls-fmp4'});
+    }
+
+    if (audioGroupKeys.length
+        && Object.keys(mediaGroups.AUDIO[audioGroupKeys[0]]).length > 1) {
```
mjneil: To keep style consistency, lets put the `&&` at the end of the previous line instead of the start of this line

README.md
```diff
@@ -476,6 +479,51 @@ are triggered on the player object.
...
+### HLS Usage Events
+
+Usage tracking events are fired when we detect a certain HLS feature, encoding setting,
+or API is used. These can be helpful for analytics, and to pinpoint the cause of HLS errors.
+For instance, if errors are being fired in tandem with a usage event indicating that the
+player was playing an AES encrypted stream, then we have a possible avenue to explore when
+debugging the error.
+
+Note that although these usage events are listed below, they may change at any time without
+a major version change.
```
mjneil: Maybe add that 
>HLS usage events are triggered on the `tech` with the exception of the 3 `hls-reload-error` events, which are triggered on the `player`.

test/reload-source-on-error.test.js
```diff
@@ -104,19 +111,22 @@ QUnit.test('by default, only allows a retry once every 30 seconds', function(ass
   let hlsErrorReloadEvents = 0;
   let hlsErrorReloadCanceledEvents = 0;
 
-  this.player.tech_.on('usage', (event) => {
+  this.player.ready = (callback) => {
```
mjneil: This shouldn't be necessary since you set `this.player.ready` in the `beforeEach`

zhuangs: my bad forgot to delete this one and deleted

Issue #1165, opened 2017-06-15
Failed to construct 'VTTCue' for segmentMetadataTrack


mjneil:
## Description
Sometimes bad content can trip up the segment probe and we get invalid timing data back. This timing data is used for the segment metadata cues, but if we provide a non-finite double value, an error will be thrown. These cues are nice to have, but not essential to playback, so failure to add them should not stop playback

PR #1167, opened 2017-06-16
don't crash when segment metadata cues can't be created

mjneil:
A bad segment might trip up our segment probe. When trying to create the metadata cue for that segment, the player can crash. We shouldn't disrupt playback for something non-essential to playback

#1165

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -43,6 +43,8 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
     appendedLastSegment;
 };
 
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -1086,6 +1086,11 @@ export default class SegmentLoader extends videojs.EventTarget {
     const start = segment.start;
     const end = segment.end;
 
+    // Do not try adding the cue if the start and end times are invalid.
+    if (!Number.isFinite(start) || !Number.isFinite(end)) {
```
gesinger: `isFinite` is not supported in IE. Do we have a polyfill for it?

PR #1173, opened 2017-06-19
use sync controller to add extra request estimation

mjneil:
some smarter abort early logic


src/playlist-selectors.js
```diff
@@ -255,44 +255,27 @@ export const movingAverageBandwidthSelector = function(decay) {
...
- * @return {Object}
- *         Object with two properties
- *         playlist - The playlist object that will result in the minimum rebuffering
- *         roundTripTime - The estimated amount of time in seconds it will take to
- *         switch to this playlist and download a segment
+ * @return {Array}
+ *         List of playlists with estimated segment round trip times sorted by round trip
+ *         time
  */
 export const minRebufferingSelector = function(master,
```
gesinger: Since this is no longer selecting anything (instead it returns everything), we should change the function name.

gesinger: Actually, since it is much smaller, it might be worth inlining it. It would also shorten later code.

src/segment-loader.js
```diff
@@ -741,31 +740,58 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      (playlist) => playlist.attributes && playlist.attributes.BANDWIDTH
+    ).map((playlist) => {
+      const syncPoint = this.syncController_.getSyncPoint(playlist,
+                                                          this.duration_(),
+                                                          this.currentTimeline_,
+                                                          currentTime);
+      // There is no sync point for this playlist, so switching to it will require a
+      // sync request first. This will double the round trip time
+      const numRequests = syncPoint ? 1 : 2;
+      const segmentSize = segmentDuration * playlist.attributes.BANDWIDTH;
```
gesinger: Minor, but the segments on different playlists might have different durations, thus could have a small impact. I doubt it will be noticeable, but it's possible. (Same for two request sets, but probably isn't worth worrying too much about.)

src/segment-loader.js
```diff
@@ -741,31 +740,58 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      const roundTripTime = segmentSize * numRequests / measuredBandwidth;
+
+      return {
+        playlist,
+        roundTripTime
+      };
+    }).sort((a, b) => {
+      if (a.roundTripTime <= timeUntilRebuffer && b.roundTripTime <= timeUntilRebuffer) {
+        // if both playlists will prevent rebuffernig, prioritize highest bandwidth
```
gesinger: rebuffering

src/segment-loader.js
```diff
@@ -741,31 +740,58 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        return b.playlist.attributes.BANDWIDTH - a.playlist.attributes.BANDWIDTH;
+      }
+
+      // if neither can result in no rebuffering, sort by round trip time
```
gesinger: Comment is out-of-date, can remove.

src/segment-loader.js
mjneil: This is true. There are two problems regarding the segment duration. One is that not all segments have to have the same duration. Second is that if we have not yet requested one of the playlists, we don't even have target duration yet to estimate segment sizes on that playlist. I thought that the best alternative would be to use the duration of the segment we are requesting, because although not required, its likely that across renditions, the segments are segmented in a similar way, so if we want to cancel the current request to request a smaller replacement, we will probably be requesting a segment with similar duration to the one we are cancelling

PR #1175, opened 2017-06-21
Fix endOfStream for demuxed audio and video

mjneil:
## Description
When audio is demuxed from the video, the audio and video segment loaders enter a race for who can finish first and call `MediaSource.endOfStream()`. This means whichever loader finishes last won't actually be able to append all its segments. This change moves the responsibility of calling `endOfStream` to the `MasterPlaylistController`

## Specific Changes proposed
* `SegmentLoader`s now trigger an `ended` event and set a private `ended_` state instead of calling `MediaSource.endOfStream()`
* `MasterPlaylistController` listens for the `ended` event on the `SegmentLoader`s and calls `MediaSource.endOfStream()` if all active `SegmentLoader`s have `ended_ === true`
* `ended_` state is reset on `sourceopen` events from the `MediaSource` as well as when `SegmentLoader.resetEverything()` is called.

## Requirements Checklist


src/master-playlist-controller.js
```diff
@@ -1142,6 +1150,28 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
   /**
+   * Calls endOfStream on the media source when all active stream types have called
+   * endOfStream
+   *
+   * @param {string} streamType
+   *        Stream type of the segment loader that called endOfStream
+   * @private
+   */
+  onEndOfStream() {
+    let isEndOfStream = this.mainSegmentLoader_.ended_;
```
gesinger: Can probably just do one line:

```
let isEndOfStream = this.audioPlaylistLoader_ ? this.audioSegmentLoader_.ended_ && this.mainSegmentLoader_.ended : this.mainSegmentLoader_.ended;
```

or

```
let isEndOfStream = this.mainSegmentLoader_.ended && (this.audioPlaylistLoader_ ? this.audioSegmentLoader_.ended_  : true);
```

mjneil: long ternaries can be hard to read

PR #1176, opened 2017-06-21
Abr Improvements

mjneil:
## Description
This PR contains the following improvements to the ABR algorithm: 
 * Use a starting `bandwidth` value of `0.0625 MB/s` on Android devices
 * Do not allow an up-switch in quality until a certain amount of forward buffer has been filled, `BUFFER_LOW_WATER_LINE`
 * Dynamically increase the `BUFFER_LOW_WATER_LINE` and `GOAL_BUFFER_LENGTH` from `0 - > 30` and `30 -> 60` respectively during the first 30 seconds of playback
 * Abort segment requests before completion if bandwidth reported by the XHR `progress` event shows that network conditions are not fast enough to complete the request without causing rebuffering

## Requirements Checklist
  - [ ] Docs/guides updated


imbcmdth:
Good thing git can now squash and merge.. This might be the ugliest commit list in the history of videojs! 😆

src/master-playlist-controller.js
```diff
@@ -548,7 +549,34 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       // with the updated bandwidth information
-      this.masterPlaylistLoader_.media(this.selectPlaylist());
+      const nextPlaylist = this.selectPlaylist();
+      const currentPlaylist = this.masterPlaylistLoader_.media();
+      const buffered = this.tech_.buffered();
+      const forwardBuffer = buffered.length ?
+        buffered.end(buffered.length - 1) - this.tech_.currentTime() : 0;
+
+      const currentTime = this.tech_.currentTime();
+      const initial = Config.BUFFER_LOW_WATER_LINE;
```
mjneil: We should make getter functions on MPC for low water line and goal buffer length, and pass the GBL one down to the segment loader to use

src/media-segment-request.js
```diff
@@ -333,7 +332,13 @@ const waitForCompletion = (activeXhrs, decrypter, doneFn) => {
  * @param {Event} event - the progress event object from XMLHttpRequest
  */
 const handleProgress = (segment, progressFn) => (event) => {
-  segment.stats = getProgressStats(event);
+  segment.stats = videojs.mergeOptions(segment.stats, getProgressStats(event));
+
+  // record the time that we receive the first byte of data
+  if (!segment.stats.firstByteRoundTripTime && segment.stats.bytesReceived) {
+    segment.stats.firstByteRoundTripTime = Date.now();
```
mjneil: `firstByteRoundTripTime` is probably not the most accurate property name

src/videojs-contrib-hls.js
```diff
@@ -58,6 +58,98 @@ Object.defineProperty(Hls, 'GOAL_BUFFER_LENGTH', {
...
+      videojs.log.warn('value passed to Hls.BUFFER_LOW_WATER_LINE ' +
+                       'must be a number and greater than or equal to 0 and less than' +
+                       'Hls.GOAL_BUFFER_LENGTH');
+      return;
+    }
+    Config.MAX_BUFFER_LOW_WATER_LINE = v;
+  }
+});
+
+Object.defineProperty(Hls, 'BUFFER_LOW_WATER_RATE', {
```
mjneil: We should probably add tests for the two `MAX` and two `RATE` variables as well

test/master-playlist-controller.test.js
```diff
@@ -814,6 +817,134 @@ QUnit.test('re-triggers bandwidthupdate events on the tech', function(assert) {
   assert.equal(bandwidthupdateEvents, 2, 'triggered bandwidthupdate');
 });
 
+// TODO: Add test for ignoring low waterline in LIVE playback
```
mjneil: let's do this todo

src/segment-loader.js
```diff
@@ -725,6 +727,133 @@ export default class SegmentLoader extends videojs.EventTarget {
   }
 
   /**
+   * Determines if the network has enough bandwidth to complete the current segment
+   * request in a timely manner. If not, the request will be aborted early and bandwidth
+   * updated to trigger a playlist switch.
+   *
+   * @param {Object} stats
+   *        Object continaing stats about the request timing and size
```
gesinger: containing

src/segment-loader.js
```diff
@@ -725,6 +727,133 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    }
+
+    const switchCandidate = this.hls_.playlists.master.playlists.filter(
+      (playlist) => playlist.attributes && playlist.attributes.BANDWIDTH
+    ).map((playlist) => {
+      const syncPoint = this.syncController_.getSyncPoint(playlist,
+                                                          this.duration_(),
+                                                          this.currentTimeline_,
+                                                          currentTime);
+      // There is no sync point for this playlist, so switching to it will require a
```
gesinger: If there is no

src/segment-loader.js
```diff
@@ -725,6 +727,133 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    }
+
+    if (!switchCandidate.playlist ||
+        switchCandidate.playlist.uri === this.playlist_.uri ||
+        timeSavedBySwitching < minimumTimeSaving) {
+      return false;
+    }
+
+    // set the bandwidth to that of the desired playlist
+    // (Being sure to scale by BANDWIDTH_VARIANCE and add one so the
```
gesinger: Can remove parenthesis

src/segment-loader.js
```diff
@@ -776,7 +898,7 @@ export default class SegmentLoader extends videojs.EventTarget {
         seekable.start(0) < currentTime) {
       removeToTime = seekable.start(0);
     } else {
-      removeToTime = currentTime - 60;
+      removeToTime = currentTime - 30;
```
gesinger: Comment for 1 minute is out of date (above). And we should probably mention that we have to keep it lower than max LWL. Though we should probably calculate based on LWL, particularly if people set their own values.

src/videojs-contrib-hls.js
```diff
@@ -58,6 +58,98 @@ Object.defineProperty(Hls, 'GOAL_BUFFER_LENGTH', {
   }
 });
 
+Object.defineProperty(Hls, 'MAX_GOAL_BUFFER_LENGTH', {
```
gesinger: A lot of these are simply copies. We should have a method that generates them, and make all of them one line function calls.

src/config.js
```diff
@@ -1,6 +1,12 @@
...
+  MAX_GOAL_BUFFER_LENGTH: 60,
+  GOAL_BUFFER_RATE: 1,
   // A fudge factor to apply to advertised playlist bitrates to account for
   // temporary flucations in client bandwidth
-  BANDWIDTH_VARIANCE: 1.2
+  BANDWIDTH_VARIANCE: 1.2,
+  // How much of the buffer must be filled before we consider upswitching
+  BUFFER_LOW_WATER_LINE: 0,
+  MAX_BUFFER_LOW_WATER_LINE: 30,
+  BUFFER_LOW_WATER_RATE: 1
```
gesinger: For consistency, maybe `BUFFER_LOW_WATER_LINE_RATE`?

src/master-playlist-controller.js
```diff
@@ -1557,4 +1582,32 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     AdCueTags.updateAdCues(media, this.cueTagsTrack_, offset);
   }
+
+  /**
+   * Calculates the desired forward buffer length based on current time
+   *
+   * @return {Number} Desired forward buffer length in seconds
+   */
+  goalBufferLength() {
```
OshinKaramian: There is some duplication of this function in the test suite.  You could:

1.Take currentTime in as a variable into the function and use that for testing OR
2. Take currentTime in and set it to a default value, then change it/set it in the test.

This would also let you write a couple of tiny unit tests to check the internal logic of this function if you wanted to.

You could probably do the same for bufferLowWaterLine if you wanted to test it similarly.

src/playlist-selectors.js
```diff
@@ -113,81 +113,102 @@ export const comparePlaylistResolution = function(left, right) {
...
-  let bandwidthBestVariant;
-  let resolutionPlusOne;
-  let resolutionBestVariant;
-  let haveResolution;
-  let resolutionPlusOneList = [];
-  let resolutionPlusOneSmallest = [];
-  let resolutionBestVariantList = [];
-
-  stableSort(sortedPlaylists, comparePlaylistBandwidth);
+const simpleSelector = function(master, playerBandwidth, playerWidth, playerHeight) {
```
OshinKaramian: Given that I am a big n00b it would be nice to have this function commented, at the moment I'm guessing it selects the proper rendition based on the criteria that's passed in.

src/playlist-selectors.js
```diff
@@ -113,81 +113,102 @@ export const comparePlaylistResolution = function(left, right) {
...
-  haveResolution = bandwidthPlaylists.filter(function(elem) {
-    return elem.attributes &&
-           elem.attributes.RESOLUTION &&
-           elem.attributes.RESOLUTION.width &&
-           elem.attributes.RESOLUTION.height;
-  });
+  let haveResolution = bandwidthPlaylistReps.filter((rep) => rep.width && rep.height);
+
+  // sort variants by resolution
+  stableSort(haveResolution, (left, right) => left.width - right.width);
```
OshinKaramian: This seems to be sorting and then filtering it, is there any reason not to sort the final output from whatever filters you've applied?

src/segment-loader.js
```diff
@@ -725,6 +727,133 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+   * request in a timely manner. If not, the request will be aborted early and bandwidth
+   * updated to trigger a playlist switch.
+   *
+   * @param {Object} stats
+   *        Object continaing stats about the request timing and size
+   * @return {Boolean} True if the request was aborted, false otherwise
+   * @private
+   */
+  abortRequestEarly_(stats) {
+    if (this.hls_.tech_.paused() ||
```
OshinKaramian: Anyway to break this almost 100 line function down into a couple of smaller functions?  

Maybe some of the "ifs" could be moved elsewhere? Or some of the logic in the sort functions?

src/segment-loader.js
```diff
@@ -725,6 +727,133 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+   *        A simplified segment object copy
+   * @private
+   */
+  handleProgress_(event, simpleSegment) {
+    if (!this.pendingSegment_ ||
+        simpleSegment.requestId !== this.pendingSegment_.requestId) {
+      return;
+    }
+
+    if (this.abortRequestEarly_(simpleSegment.stats)) {
```
OshinKaramian: Could just move this condition up into the if statement above.

src/sync-controller.js
```diff
@@ -46,7 +46,7 @@ export const syncPointStrategies = [
   {
     name: 'Segment',
     run: (syncController, playlist, duration, currentTimeline, currentTime) => {
-      let segments = playlist.segments;
+      let segments = playlist.segments || [];
```
OshinKaramian: Could do something like let { segments = [] } = playlist;

src/videojs-contrib-hls.js
OshinKaramian: I would also add that you could shove the property string 'MAX_GOAL_BUFFER_LENGTH' into a variable somewhere.

If these functions are all similar enough you could probably dump them all in an array and run through it to define these objects.

src/videojs-contrib-hls.js
```diff
@@ -278,10 +370,11 @@ class HlsHandler extends Component {
...
 
     // start playlist selection at a reasonable bandwidth for
-    // broadband internet
-    // 0.5 MB/s
+    // broadband internet (0.5 MB/s) or mobile (0.0625 MB/s)
     if (typeof this.options_.bandwidth !== 'number') {
-      this.options_.bandwidth = 4194304;
+      // only use Android for mobile because iOS does not support MSE (and uses
+      // native HLS)
+      this.options_.bandwidth = videojs.browser.IS_ANDROID ? 500000 : 4194304;
```
OshinKaramian: Would be neat to store these constants in a couple of variables.

test/loader-common.js
```diff
@@ -27,8 +27,22 @@ export const LoaderCommonHooks = {
...
+    this.playbackRate = 1;
     this.fakeHls = {
-      xhr: xhrFactory()
+      xhr: xhrFactory(),
+      tech_: {
+        paused: () => this.paused,
+        playbackRate: () => this.playbackRate
+      }
+    };
+    this.goalBufferLength = () => {
```
OshinKaramian: This is the function I mentioned before that you may be able to avoid duplicating here.

test/loader-common.js
```diff
@@ -275,6 +290,71 @@ export const LoaderCommonFactory = (LoaderConstructor,
...
+            playlist4
+          ]
+        }
+      };
+
+      loader.on('bandwidthupdate', () => bandwidthupdates++);
+      loader.playlist(playlist1, xhrOptions);
+      loader.load();
+
+      this.clock.tick(1);
```
OshinKaramian: If the order of each test here is unimportant (one test doesn't rely on another) you could break it up into a few smaller/more specific tests.  It might make it easier to debug if one specific case fails but the rest work.

src/segment-loader.js
mjneil: I think maybe we should calculate based on GBL. The reason we lowered this was we increased GBL from 30 -> 60, so we decreased the trim. Maybe we should have a rule of only having 90 (or something similar, while also having some minimum trim incase someone sets GBL very high, we still want to trim content) seconds buffered max and depending on GBL, trim the remaining?

src/videojs-contrib-hls.js
mjneil: Should we make them all have the same constraints (e.g. positive integer) and remove the low waterline constraints for being below GBL? We can just make sure to check for valid numbers whenever we use them and make no assumptions about what the numbers are, which we already do I believe

src/playlist-selectors.js
mjneil: ONe reason is that we reuse `haveResolution` in different filters to get different values, so it's probably better to sort once then filter twice than filter twice and sort twice. We only ever sort twice, once by bandwidth and once by resolution and I think its easier to reason about the filters when they are working on sorted lists

src/config.js
mjneil: should `GOAL_BUFFER_RATE` be updated to `GOAL_BUFFER_LENGTH_RATE` then?

src/playlist-selectors.js
mjneil: Yea it selects the best rendition available given the bandwidth and player size constraints

src/config.js
gesinger: Makes sense to me

src/playlist-selectors.js
OshinKaramian: Ah, makes sense.

src/segment-loader.js
gesinger: Makes sense.

src/videojs-contrib-hls.js
gesinger: I think since the variable name is only used in one spot, it might make more sense to keep it as a string inline. If we end up using it in more than one spot, then we should consider making a constant for it.

I think we can remove the various constraints for now. Most people should realize that setting a LWL above GBL just won't work. If people run into trouble with them in the future, then we can add in the config specific constraints again.

test/loader-common.js
mjneil: This test is actually just one test, order is very important here. There is only one segment request in flight for this test, each `dispatchEvent` call is mocking a progress event for the xhr request. using the value on `loaded` within the event and the number of clock `tick`s that have passed since the request has been made is what is being used for the estimated bandwidth

src/segment-loader.js
mjneil: deciding to go back to 60 second trim. Comment in code above stating `150mb` should actually be `150MB` which is ~5 minutes of playback, so we should be safe keeping 60 sec of back buffer

test/loader-common.js
OshinKaramian: Gotcha, that makes sense.

src/master-playlist-controller.js
mjneil: In the test suite i decided to set `this.goalBufferLength = MasterPlaylistController.prototype.goalBufferLength.bind(this)` to eliminate the duplication. How are you with that approach?

src/segment-loader.js
```diff
@@ -725,6 +728,117 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+                                                 currentTime,
+                                                 this.hls_.tech_.playbackRate()) - 1;
+
+    // Only consider aborting early if the estimated time to finish the download
+    // is larger than the estimated time until the player runs out of forward buffer
+    if (requestTimeRemaining <= timeUntilRebuffer) {
+      return false;
+    }
+
+    const switchCandidate = minRebufferMaxBandwidthSelector({
```
OshinKaramian: 😎

src/playlist.js
```diff
@@ -411,12 +411,58 @@ export const isEnabled = function(playlist) {
...
+                                                   bandwidth,
+                                                   playlist,
+                                                   bytesReceived = 0) {
+  if (!hasAttribute('BANDWIDTH', playlist)) {
+    return NaN;
+  }
+
+  const size = segmentDuration * playlist.attributes.BANDWIDTH;
+
+  return (size - (bytesReceived * 8)) / bandwidth;
```
gesinger: Should this be `bitsReceived`

src/segment-loader.js
gesinger: 😎

src/segment-loader.js
```diff
@@ -898,7 +883,7 @@ export default class SegmentLoader extends videojs.EventTarget {
         seekable.start(0) < currentTime) {
       removeToTime = seekable.start(0);
     } else {
-      removeToTime = currentTime - 30;
+      removeToTime = currentTime - 60;
```
gesinger: This may be too high.

src/videojs-contrib-hls.js
```diff
@@ -40,132 +45,32 @@ const Hls = {
...
-    if (typeof v !== 'number' || v <= 0) {
-      videojs.log.warn('value passed to Hls.BUFFER_LOW_WATER_RATE ' +
-                       'must be a number and greater than 0');
-      return;
-    }
-    Config.BUFFER_LOW_WATER_RATE = v;
-  }
-});
+      if (typeof value !== 'number' || value < 0) {
+        videojs.log.warn(`value passed to Hls.${prop} must be a positive number or 0`);
```
gesinger: Might be easier to read as `must be greater than or equal to 0`

src/videojs-contrib-hls.js
```diff
@@ -374,7 +279,8 @@ class HlsHandler extends Component {
     if (typeof this.options_.bandwidth !== 'number') {
       // only use Android for mobile because iOS does not support MSE (and uses
       // native HLS)
-      this.options_.bandwidth = videojs.browser.IS_ANDROID ? 500000 : 4194304;
+      this.options_.bandwidth =
+        videojs.browser.IS_ANDROID ? BANDWIDTH_MOBILE : BANDWIDTH_BROADBAND;
```
gesinger: Might be easier read/more precise as `INITIAL_BANDWIDTH_MOBILE` and `INITIAL_BANDWIDTH_DESKTOP`

src/playlist.js
mjneil: No, thats why we multiply by 8 to convert to bits

src/videojs-contrib-hls.js
mjneil: You just like making staying under 90 chars per line difficult

src/playlist-selectors.js
```diff
@@ -242,3 +278,80 @@ export const movingAverageBandwidthSelector = function(decay) {
...
+    const rebufferingImpact = (requestTimeEstimate * numRequests) - timeUntilRebuffer;
+
+    return {
+      playlist,
+      rebufferingImpact
+    };
+  });
+
+  const noRebufferingPlaylists = rebufferingEstimates.filter(
+    (estimate) => estimate.rebufferingImpact < 0);
```
gesinger: Should this be <=?

test/configuration.test.js
```diff
@@ -72,12 +82,154 @@ QUnit.test('GOAL_BUFFER_LENGTH set warning and invalid', function(assert) {
...
   assert.equal(Config.GOAL_BUFFER_LENGTH, 30, 'default');
 
-  Hls.GOAL_BUFFER_LENGTH = 0;
+  Hls.GOAL_BUFFER_LENGTH = -1;
   assert.equal(this.env.log.warn.calls, 2, 'logged two warnings');
 
   assert.equal(Config.GOAL_BUFFER_LENGTH, 30, 'default');
 });
 
+QUnit.test('MAX_GOAL_BUFFER_LENGTH get warning', function(assert) {
```
gesinger: Although I'm usually for non shared tests, these are similar enough that it might be worth abstracting the test logic, and also checking the warning messages.

test/master-playlist-controller.test.js
```diff
@@ -872,6 +877,157 @@ QUnit.test('re-triggers bandwidthupdate events on the tech', function(assert) {
...
+  };
+
+  currentTime = 0;
+  currentPlaylistBandwidth = 1000;
+  nextPlaylistBandwidth = 1000;
+  buffered = [];
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.equal(mediaChanges.length,
+               1,
+               'changes media when no buffer and equal bandwidth playlist');
```
gesinger: I don't think that a change should happen when the bandwidths are equal (either by how the code works or by what we should do).

test/master-playlist-controller.test.js
```diff
@@ -872,6 +877,157 @@ QUnit.test('re-triggers bandwidthupdate events on the tech', function(assert) {
...
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.equal(mediaChanges.length,
+               0,
+               'did not change media when insufficient forward buffer and higher ' +
+               'bandwidth playlist');
+  buffered = [[0, 21]];
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.equal(mediaChanges.length,
+               1,
+               'changes media when sufficient forward buffer and higher ' +
```
gesinger: Might be worth having another for forward buffer equal to dynamic LWL

test/master-playlist-controller.test.js
```diff
@@ -872,6 +877,157 @@ QUnit.test('re-triggers bandwidthupdate events on the tech', function(assert) {
...
+               2,
+               'changes live media when insufficient forward buffer and higher ' +
+               'bandwidth playlist');
+  buffered = [[0, 100], [100, 130]];
+  this.masterPlaylistController.mainSegmentLoader_.trigger('bandwidthupdate');
+  assert.equal(mediaChanges.length,
+               3,
+               'changes live media when sufficient forward buffer and higher ' +
+               'bandwidth playlist');
+});
```
gesinger: Would be good to have some tests for always switching when duration < LWL.

test/master-playlist-controller.test.js
mjneil: It looks like it succeeds here because `currentTime === 0` and initial `BUFFER_LOW_WATER_LINE` is 0, so the check `forwardBuffer >= bufferLowWaterLine` is `true`.

PR #1199, opened 2017-07-17
[ABR] Smooth manual switching, Quality switch history, Map buffered quality by playhead position

tchakabam:
## Description

We have several requirements:

* Switch quality "manuall"y without flushing. In our case application business logic will drive the quality for the user, but we want it to be non-flushing even if driven through the Representations interface
* Display the history of quality switches (manual and automatical ones)
* Display the currently played quality at any point
* Display the effective bitrate (average per segment) of the stream at any point 

-> Therefore 

## Specific Changes proposed

* We add an option to use a `smoothQualityChangeFn` instead of the default `fastQualityChangeFn`.
* We add a history of quality switches
* We add a map of the buffer vs representation attributes that allows to lookup a playhead time in buffer and returns the quality attributes
* We add a manual quality switching menu to the demo page
* We add displaying current quality in the demo page
* We fix a bug in `playlist-selector` where when all playlists happen to be disabled there is no fallback. We use the first element in the sorted unfiltered list of reps as an ultimate fallback.

* For debugging, we add a plot to the demo page which displays the buffered quality (playlist bandwidth & effective bitrate):

videojs-hls-buffer-state-graph

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,24 +46,97 @@
     <li><a href="/examples">Browse Examples</a></li>
   </ul>
 
+  <select id="quality-picker"></select>
+
+  <p>
+    Current quality attributes:
+    <br>
+    <pre id="current-quality-attr"></pre>
+  </p>
+
+  <script src="/node_modules/jquery/dist/jquery.js"></script>
   <script src="/node_modules/video.js/dist/video.js"></script>
+  <script src="/node_modules/videojs-contrib-quality-levels/dist/videojs-contrib-quality-levels.js"></script>
   <script src="/dist/videojs-contrib-hls.js"></script>
   <script>
     (function(window, videojs) {
-      var player = window.player = videojs('videojs-contrib-hls-player');
+
+      var videojsHlsDemo = {
+        player: null,
+        url: null
+      };
+
+      var playerOptions = {
+        //
+      };
+
+      var player = videojs('videojs-contrib-hls-player', playerOptions);
 
       // hook up the video switcher
       var loadUrl = document.getElementById('load-url');
       var url = document.getElementById('url');
       loadUrl.addEventListener('submit', function(event) {
         event.preventDefault();
+        videojsHlsDemo.url = url.value;
         player.src({
           src: url.value,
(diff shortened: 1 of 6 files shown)


gesinger:
Hey @tchakabam , thanks for the PR, we really appreciate it! The graph looks really cool. We're still looking at some of the changes, particularly related to switching without clearing the buffer, but for the quality graph and tracking the buffered qualities, did you happen to look into the `utils/stats/index.html` page ( https://github.com/videojs/videojs-contrib-hls/blob/master/utils/stats/index.html ) and the `segment-metadata` track ( https://github.com/videojs/videojs-contrib-hls#segment-metadata ) ?

tchakabam:
@gesinger Hey, well in that case I have built a feature without knowing part of the feature already exists hahaha :D 

So now I have look at the segment-metadata stuff. For our needs it does what we want actually for now :) 

But it is a bit unhandy to resolve the actual media attributes or quality-levels properties based on the m3u8 URL. but i do understand you can not but living objects into the text-track.

but why not copy in the playlist attributes?

maybe we could improve that in this PR as well?

About the smooth quality switching: We are coordinating several Videojs HLS instances on the same page, and telling them to switch around qualities from an external controller.

Therefore we are doing "manual" switching based on representations API (or via the qualityLevels plugin, which eventually used Representations).

So when doing that and it doing always "fast" switching, it turns out to actually waste bandwidth a lot because it will flush buffers already loaded, i.e overwrite and load from an earlier point to be "fast".

We dont want to be "fast" we just want to maintain safe buffer levels :) 

I think even for manual switching it should be left to user choice if the switch is "flushing" or not. Even totally unrelated to our "special" use-case which requires us to have that in order to implement a multi-player coordinating switching behavior basically.

Of course we could build a PlaylistSelector that takes the "orders" from outside, but it seems like the Representations API should just also allow to choose about this smooth vs fast choice as well, right?

forbesjo:

PR #1207, opened 2017-07-24
[ABR] Handle "no playlist enabled" state in simple selector gracefully

tchakabam:
## Description

When we have no playlist enabled, currently this causes undefined behavior and an exception.

When enforcing a single playlist (setting all to disabled and enabling one) currently
we might have a moment where no playlist is enabled.

But also the user may cause this state for various reasons on the API.
We should be able to handle this gracefully (while eventually logging a warning)

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-selectors.js
+++ b/src/playlist-selectors.js
@@ -129,6 +129,7 @@ export const comparePlaylistResolution = function(left, right) {
  * bandwidth variance
  */
 const simpleSelector = function(master, playerBandwidth, playerWidth, playerHeight) {
+  let fallbackPlaylistRep = null;
   // convert the playlists to an intermediary representation to make comparisons easier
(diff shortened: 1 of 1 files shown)


tchakabam:
Note: this is already contained in https://github.com/videojs/videojs-contrib-hls/pull/1199 but I'll remove it from there, as it will be more consumable to ship things bit by bit per scope :)

mjneil:
Thanks for the PR, this is definitely something we should handle gracefully. 

Implementation looks solid, would you be able to add a unit test for this behavior?

(I agree, bit by bit is easier to consume 🙂)

tchakabam:
sure, will do :+1:

forbesjo:
We'll update this PR with tests.

forbesjo:

Issue #1205, opened 2017-07-21
Cannot read property 'BANDWIDTH' of undefined


SzHeJason:
![image](https://user-images.githubusercontent.com/20609396/28398586-6f7db2de-6d3a-11e7-9bd0-1b09418ccb4c.png)
![image](https://user-images.githubusercontent.com/20609396/28398599-8690e810-6d3a-11e7-884c-c0031d49ce61.png)

zcoding:
+1

zcoding:
video.js v5.19.2
videojs-contrib-hls v5.8.2

fku233:
+1

 ```
 <script src="https://unpkg.com/video.js/dist/video.js"></script>
 <script src="https://unpkg.com/videojs-contrib-hls/dist/videojs-contrib-hls.js"></script>
```

gesinger:
Thanks for the issue, however, we haven't seen it ourselves (or in our [starter template]( http://jsbin.com/vokipos/8/edit?html,console,output)). Would you be able to provide an example using our [starter template](http://jsbin.com/vokipos/8/edit?html,console,output) that shows the issue? Thank you.

SzHeJason:
platform : MAC
browser : Chrome and firefox (ps : Safari is normal )
[Demo](http://jsbin.com/gawokunebo/edit?html,console,output)
![image](https://user-images.githubusercontent.com/20609396/28573631-cc1a7f1c-717d-11e7-89e1-34ef95356855.png)

DullSky:
+1

mjneil:
This bug happens when using a non master manifest as the source as the parser won't have any attribute data to add to it. we can address this either at the [parser level ](https://github.com/videojs/m3u8-parser/issues/25) or within `PlaylistLoader`

gesinger:
@mjneil is right. We will get to https://github.com/videojs/m3u8-parser/issues/25 shortly.

gesinger:
In progress with PR: https://github.com/videojs/videojs-contrib-hls/pull/1214

thecotne:
@gesinger PR is merged so it will be fixed in next release right?

gesinger:
Yup @thecotne . We'll try to cut a release soon to get this issue resolved.

mjneil:
A fix for this has been released in v5.8.3. Please open a new issue if you continue to experience any issues.

PR #1210, opened 2017-07-25
More useful info in segment-metadata cue (bandwidth/resolution/codecs/byte-length)

tchakabam:
## Description

adds effective segment bitrate (byte-length), recommended bandwidth, resolution and codecs string to cue metadata

this way we don't need to resolve this information from the master playlist using the URIs, which is very unhandy. The URI itself is usually not the data we really need.

Is there any cost involved in adding this?

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1227,7 +1227,6 @@ export default class SegmentLoader extends videojs.EventTarget {
       bandwidth: segmentInfo.playlist.attributes.BANDWIDTH,
       resolution: segmentInfo.playlist.attributes.RESOLUTION,
       codecs: segmentInfo.playlist.attributes.CODECS,
-      bitrate: (8 * segmentInfo.byteLength / (end - start)),
       byteLength: segmentInfo.byteLength,
       uri: segmentInfo.uri,
       timeline: segmentInfo.timeline,
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1224,6 +1224,11 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     const Cue = window.WebKitDataCue || window.VTTCue;
     const value = {
+      bandwidth: segmentInfo.playlist.attributes['BANDWIDTH'],
+      resolution: segmentInfo.playlist.attributes['RESOLUTION'],
+      codecs: segmentInfo.playlist.attributes['CODECS'],
+      bitrate: (8 * segmentInfo.byteLength / (end - start)),
+      byteLength: segmentInfo.byteLength,
       uri: segmentInfo.uri,
       timeline: segmentInfo.timeline,
       playlist: segmentInfo.playlist.uri,
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -1224,9 +1224,9 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     const Cue = window.WebKitDataCue || window.VTTCue;
     const value = {
-      bandwidth: segmentInfo.playlist.attributes['BANDWIDTH'],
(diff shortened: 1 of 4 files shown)


src/segment-loader.js
```diff
@@ -1224,6 +1224,11 @@ export default class SegmentLoader extends videojs.EventTarget {
 
     const Cue = window.WebKitDataCue || window.VTTCue;
     const value = {
+      bandwidth: segmentInfo.playlist.attributes['BANDWIDTH'],
```
mjneil: `attributes` may be `undefined` on the playlist object so you'll want to check for it before accessing the inner properties.

Also travis is failing because the linter prefers dot notation `playlist.attributes.BANDWIDTH`

tchakabam: should we not rather fix the fact that `attributes` may be `undefined`? it could be initialized `{}` by default?

sure, we can go for dot notation :)

and how about we just add `attributes` as such to the cue data? that way the user has all the info, no need to make this here a replication of the semantic that is already in attributes with lots of checks and so on. Because the attributes themselves may not be present in some playlists. It's just about making available all the info we have I guess :)

mjneil: fixing the `attributes` is something we want to address, see #1205 and https://github.com/videojs/m3u8-parser/issues/25 so we could just wait on that before merging these changes in. Also just adding `attributes` as a property to `value` would work as well I'd think, though it may be better to individually set values so the cue isn't holding on the memory reference of the playlist object. Because of the way our [PlaylistLoader works](https://github.com/videojs/videojs-contrib-hls/blob/master/src/playlist-loader.js#L75) live playlist refreshes create a deep copy of the playlist and updating the copy, instead of just updating the original reference. The `attributes` aren't going to change on refreshes, and its possible there are other places in our code that is holding onto old playlist references, but its probably a good idea to not do so if we know about it

mjneil: `attributes: videojs.mergeOptions(segmentInfo.playlist.attributes, {})`

squarebracket: `attributes` being `undefined` was fixed in #1214. You do still need to change to dot notation though.

src/segment-loader.js
```diff
@@ -1224,6 +1224,11 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+      resolution: segmentInfo.playlist.attributes['RESOLUTION'],
+      codecs: segmentInfo.playlist.attributes['CODECS'],
+      bitrate: (8 * segmentInfo.byteLength / (end - start)),
```
squarebracket: I feel like `bitrate` is a bit ambiguous here. Without seeing the code, I'd probably assume it would be the same thing as `bandwidth`. I think I favour dropping it for clarity's sake since it can be calculated from the other values. Is there a reason you're including it, other than convenience?

squarebracket:
It would be nice if you could modify [this test](https://github.com/videojs/videojs-contrib-hls/blob/master/test/segment-loader.test.js#L412) to test that the new attributes are added to the cue. You could modify [this line](https://github.com/videojs/videojs-contrib-hls/blob/master/test/test-helpers.js#L380) for the test helper playlist creator to be
```
    attributes: conf && typeof conf.attributes !== 'undefined' ? conf.attributes : {}
```
and then you could simply attach some test data to the [playlist creation](https://github.com/videojs/videojs-contrib-hls/blob/master/test/segment-loader.test.js#L416) with
```
        let playlist = playlistWithDuration(50, {attributes: {BANDWIDTH: 4500000, ...}});
```

tchakabam:
oops, haven't looked into this since a while. yeah sure we can do that! will keep you posted about it.

PR #1212, opened 2017-07-26
Add option to select lowest rendition available on startup

OshinKaramian:
## Description
I'm still working through unit tests because many of them broke, but would like eyes on this sooner rather than later in case I went down a bizarre path.

This PR will select the lowest bitrate playlist with video on startup.  If no such playlist exists, it picks the lowest bitrate playlist (assuming everything else is audo only).

This feature is gated behind an options flag.  In order to enable it set options to: 

```
{
  html5: {
    hls: {
      enableLowInitialPlaylist: true
   }
}
```
## Requirements Checklist
  - [ ] Docs/guides updated


src/master-playlist-controller.js
```diff
@@ -66,36 +67,6 @@ const objectChanged = function(a, b) {
 };
 
 /**
- * Parses a codec string to retrieve the number of codecs specified,
```
OshinKaramian: This function doesn't seem to be reliant on being here (there are a few other functions here that could be moves as well).  I shifted this into a util module as I wanted to use this to check whether a playlist contained video or not in the playlist selection algorithm below.

src/playlist-selectors.js
```diff
@@ -356,3 +357,27 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
...
+  const playlistsWithVideo =
+    playlists.filter(playlist => parseCodecs(playlist.attributes.CODECS).videoCodec);
+
+  // If there are any items in this playlist with a video codec, return
+  // the one with the lowest bitrate, otherwise return the lowest audio rendition
+  if (playlistsWithVideo.length !== 0) {
+    return playlistsWithVideo[0];
+  }
+
+  return playlists[0];
```
OshinKaramian: Made an assumption here that may or may not be true, where the playlist are sorted by bitrate.  If that's not true I can add a sort in here.

mjneil: You'll have to sort, no guarantee of order

src/videojs-contrib-hls.js
```diff
@@ -328,6 +330,10 @@ class HlsHandler extends Component {
       this.selectPlaylist ?
         this.selectPlaylist.bind(this) : Hls.STANDARD_PLAYLIST_SELECTOR.bind(this);
 
+    this.masterPlaylistController_.initialSelectPlaylist =
+      this.selectPlaylist ?
```
mjneil: After a source change, this will set the `initialSelectPlaylist` to be the standard selector and not the initial one. You should add another properties to the source handler for `selectInitialPlaylist` and then this becomes `this.selectInitialPlaylist ? this.selectInitialPlaylist.bind(this) : Hls.INITIAL_PLAYLIST_SELECTOR.bind(this)`.

side note, by my snippet you'll see i personally prefer `selectInitialPlaylist` over `initialSelectPlaylist`, but either is descriptive and fine, so feel free to change or keep the naming. @gesinger probably has his own preference

OshinKaramian: Good catch, I reworked things a bit (I set the selection in the master-playlist-controller) and must have mucked that up.

src/playlist-selectors.js
```diff
@@ -356,3 +357,27 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
...
+ * Chooses the appropriate media playlist, which in this case is the lowest bitrate
+ * one with video.  If no renditions with video exist, return the lowest audio rendition.
+ *
+ * @return {Object|null}
+ *         {Object} return.playlist
+ *         The lowest bitrate playlist that contains a video codec.  If no such rendition
+ *         exists pick the lowest audio rendition.
+ */
+export const lowestBitrateCompatibleVariantSelector = function() {
+  const { playlists } = this.playlists.master;
```
mjneil: since sort messes with the array, you should use a copy of playslists with `slice`

src/playlist-selectors.js
```diff
@@ -356,3 +357,27 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
...
+
+  // Sort ascending by bitrate
+  stableSort(playlists,
+    (a, b) => comparePlaylistBandwidth(a, b));
+
+  const playlistsWithVideo =
+    playlists.filter(playlist => parseCodecs(playlist.attributes.CODECS).videoCodec);
+
+  const selectedPlaylists = playlistsWithVideo.length !== 0 ? playlistsWithVideo : playlists;
```
mjneil: you could omit this and just `return playlistsWithVideo[0] || playlists[0];`

test/master-playlist-controller.test.js
```diff
@@ -212,6 +212,50 @@ QUnit.test('resets SegmentLoader when seeking in flash for both in and out of bu
 
   });
 
+QUnit.only('selects lowest bitrate rendition when enableLowInitialPlaylist is set',
+  function(assert) {
+    this.player = createPlayer({ html5: { hls: { enableLowInitialPlaylist: true } } });
```
mjneil: since the initial selector is called within the `loadedplaylist` event handler which fires everytime a live playlist is refreshed, I think it would be worth simulating a live refresh and confirming that the initial selector is not called a second time

test/master-playlist-controller.test.js
```diff
@@ -212,6 +212,50 @@ QUnit.test('resets SegmentLoader when seeking in flash for both in and out of bu
...
+    this.clock.tick(1);
+
+    const masterPlaylistController = this.player.tech_.hls.masterPlaylistController_;
+    const masterPlaylistLoader = masterPlaylistController.masterPlaylistLoader_;
+
+    let numCallsToSelectInitialPlaylistCalls = 0;
+    let numCallsToSelectPlaylist = 0;
+
+    // master
+    this.standardXHRResponse(this.requests.shift());
```
mjneil: These two responses should go after the `sourceopen` `trigger`, also I would expect that the `numCallsToSelectInitialPlaylistCalls` should increment after just the master response

src/master-playlist-controller.js
```diff
@@ -430,7 +403,11 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
       if (!updatedPlaylist) {
         // select the initial variant
-        this.initialMedia_ = this.selectPlaylist();
+        const initialPlaylistSelector =
+          this.enableLowInitialPlaylist ?
+            this.selectInitialPlaylist : this.selectPlaylist;
+
+        this.initialMedia_ = initialPlaylistSelector();
```
mjneil: its not necessary to store the initial media on the `MasterPlaylistController`

mjneil: You can do this without needing to create the `initialPlaylistSelector` variable

test/master-playlist-controller.test.js
```diff
@@ -212,14 +212,64 @@ QUnit.test('resets SegmentLoader when seeking in flash for both in and out of bu
...
+
+    this.masterPlaylistController.mediaSource.trigger('sourceopen');
+    // master
+    this.standardXHRResponse(this.requests.shift());
+    // media
+    this.standardXHRResponse(this.requests.shift());
+
+    this.clock.tick(1);
+
+    // Trigger playlist event which should utilize selectInitialPlaylist and
```
mjneil: this comment is a bit outdated now

test/master-playlist-controller.test.js
```diff
@@ -212,14 +212,64 @@ QUnit.test('resets SegmentLoader when seeking in flash for both in and out of bu
...
+    // not selectPlaylist
+    assert.equal(numCallsToSelectInitialPlaylistCalls, 1, 'selectInitialPlaylist');
+    assert.equal(numCallsToSelectPlaylist, 0, 'selectPlaylist');
+
+    // Simulate a live reload
+    this.masterPlaylistController.masterPlaylistLoader_.trigger('loadedplaylist');
+    this.clock.tick(5);
```
mjneil: shouldn't need this tick

src/playlist-selectors.js
```diff
@@ -356,3 +357,25 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
 
   return rebufferingEstimates[0] || null;
 };
+
+/**
+ * Chooses the appropriate media playlist, which in this case is the lowest bitrate
+ * one with video.  If no renditions with video exist, return the lowest audio rendition.
+ *
```
mjneil: Can you add `Expects to be called within the context of an instance of HlsHandler` to this. see `lastBandwidthSelector` jsdoc

mjneil:
Do you think it would be appropriate to move the other codec functions/constants within `MasterPlaylistController.js` into `utils/codecs.js`? 

specifically 
* `defaultCodecs`
* `mapLegacyAvcCodecs_`
* `makeMimeTypeString`
* `getContainerType`
* `getCodecs`
* `mimeTypesForPlaylist_`

OshinKaramian:
I do think so as I kind of started doing that then pulled back. I was trying to go with a softer touch on this PR (also it's unclear if this will end up getting folded into the greater project).  I think doing that in a separate PR might make the most sense.

OshinKaramian:
@mjneil I opened up https://github.com/videojs/videojs-contrib-hls/pull/1217 in response to your comment about moving some more generic things out of master-playlist-controller.

src/master-playlist-controller.js
```diff
@@ -283,7 +254,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
       bandwidth,
       externHls,
       useCueTags,
-      blacklistDuration
+      blacklistDuration,
+      enableLowInitialPlaylist
```
forbesjo: We'll want this added to the README eventually

src/master-playlist-controller.js
```diff
@@ -429,8 +402,17 @@ export class MasterPlaylistController extends videojs.EventTarget {
       let updatedPlaylist = this.masterPlaylistLoader_.media();
 
       if (!updatedPlaylist) {
-        // select the initial variant
-        this.initialMedia_ = this.selectPlaylist();
+        let selectedMedia;
```
OshinKaramian: Just a note here.

One of the tests was breaking on this before (https://github.com/videojs/videojs-contrib-hls/blob/fast-start-time/test/videojs-contrib-hls.test.js#L1221) (I didn't have the temp variable, and instead set this.initialMedia_ appropriately).

Once I set things to the temporary variable the problem resolved itself.  I'm assuming there are side effects off of setting this.initialMedia_... I didn't really dig deep enough to understand what was going on.  It seems a little dangerous to have side effects off of setting a variable (when the setter isn't explicitly a function).

test/videojs-contrib-hls.test.js
```diff
@@ -1786,7 +1787,7 @@ QUnit.test('uses mobile default bandwidth if browser is Android', function(asser
   openMediaSource(this.player, this.clock);
 
   assert.equal(this.player.tech_.hls.bandwidth,
-               500000,
+               4194304,
```
OshinKaramian: We discussed that we were going to use the desktop bandwidth number everywhere, so this test had to be updated to account for that (I figured I would update over removing).

src/playlist-selectors.js
```diff
@@ -375,5 +375,5 @@ export const lowestBitrateCompatibleVariantSelector = function() {
   const playlistsWithVideo =
     playlists.filter(playlist => parseCodecs(playlist.attributes.CODECS).videoCodec);
 
-  return playlistsWithVideo[0] || playlists[0];
+  return playlistsWithVideo[0] || null;
```
mjneil: instead of returning null, you could call the last bandwidth selector with the current context
` return playlistsWithVideo[0] || lasBandwidthSelector.call(this);` I think this would also remove the need to for the temporary `selectedMedia`

src/master-playlist-controller.js
mjneil: We should probably dig into why this was happening since I don't think it should cause the test to break

mjneil: Looked into this a bit more, its because if you set `this.initialMedia_` and then try and load that media, but the request fails e.g. 404, when on the final rendition, we get back to here, and `updatedPlaylist` is still `undefined` because we haven't successfully loaded one yet, but `initialMedia_` has been set from the previous time here. I think it would make sense to set `initialMedia_` in the `loadedmetadata` handler if it hasnt been set yet as the first time we get into `loadedmetadata` is after the first media playlist has been loaded.

src/playlist-selectors.js
```diff
@@ -352,3 +353,27 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
...
+ * @return {Object|null}
+ *         {Object} return.playlist
+ *         The lowest bitrate playlist that contains a video codec.  If no such rendition
+ *         exists pick the lowest audio rendition.
+ */
+export const lowestBitrateCompatibleVariantSelector = function() {
+  const playlists = this.playlists.master.playlists.slice();
+
+  // Sort ascending by bitrate
+  stableSort(playlists,
```
mjneil: Based on the findings from the 404 unit test failing (discussion in other comment thread), this selector should probably filter by `isEnabled` similar to the `simpleSelector` since this may try and re-select a playlist that just 404'd

src/playlist-selectors.js
OshinKaramian: If possible I'd like to avoid coupling this function to a separate algorithm, and allow the consumer to decide the alternative.  I think long term adding the complexity here over elsewhere will be a bit confusing, as well as making unit testing and changes harder/more complex.

src/master-playlist-controller.js
OshinKaramian: That loop seems like a code smell to me, though I can't think of a better way to deal with it in the current context.

I believe I handled the scenario in your comment below with the latest commit.

src/playlist-selectors.js
OshinKaramian: Good catch, I believe the latest commit should deal with that scenario.

src/playlist-selectors.js
mjneil: makes sense 👍

src/playlist-selectors.js
```diff
@@ -352,3 +353,39 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
...
+    (a, b) => comparePlaylistBandwidth(a, b));
+
+  // filter out any playlists that have been excluded due to
+  // incompatible configurations or playback errors
+  const enabledPlaylists = playlists.filter(
```
mjneil: this could be one line `const enabledPlaylists = playlists.filter(Playlist.isEnabled);`

src/videojs-contrib-hls.js
```diff
@@ -279,12 +279,15 @@ class HlsHandler extends Component {
...
-      this.options_.bandwidth =
-        videojs.browser.IS_ANDROID ? INITIAL_BANDWIDTH_MOBILE : INITIAL_BANDWIDTH_DESKTOP;
+      this.options_.bandwidth = INITIAL_BANDWIDTH;
     }
 
+    // If the bandwidth number is unchanged from the initial setting
+    // then this takes precedence over the enableLowInitialPlaylist option
+    this.options_.enableLowInitialPlaylist =
+      (this.options_.enableLowInitialPlaylist &&
+       this.options_.bandwidth === INITIAL_BANDWIDTH) || false;
```
mjneil: the `|| false` is not necessary `a && b` will just be the value of `b` when `a == true`

test/playlist-selectors.test.js
```diff
@@ -121,3 +122,34 @@ function(assert) {
...
+    ];
+
+    const expectedPlaylist = this.hls.playlists.master.playlists[2];
+    const testPlaylist = lowestBitrateCompatibleVariantSelector.call(this.hls);
+
+    assert.equal(testPlaylist, expectedPlaylist,
+      'Selected lowest compatible playlist with video assets');
+  });
+
+test('lowestBitrateCompatibleVariantSelector picks lowest audio rendition if no video exists',
```
mjneil: The description is no longer accurate (now returns null) Same goes for the assertion description below

src/playlist-selectors.js
```diff
@@ -352,3 +355,37 @@ export const minRebufferMaxBandwidthSelector = function(settings) {
...
+ *
+ * Expects to be called within the context of an instance of HlsHandler
+ *
+ * @return {Object|null}
+ *         {Object} return.playlist
+ *         The lowest bitrate playlist that contains a video codec.  If no such rendition
+ *         exists pick the lowest audio rendition.
+ */
+export const lowestBitrateCompatibleVariantSelector = function() {
+  const playlists = this.playlists.master.playlists.slice();
```
mjneil: could replace this `slice` call with the enabled filter, then you wouldn't need to do it later or have the local const

PR #1214, opened 2017-07-28
attach attributes property to playlist objects in cases the m3u8-pars…

mjneil:
…er does not

## Description
The m3u8-parser does not attach `attributes` property to the playlist object when it is parsing a media playlist or a master manifest that has an `EXT-X-STREAM-INF` tag without an attribute list. This has caused some `undefined` reference errors as `attributes` is not always checked first. 

## Specific Changes proposed
This change attaches `attributes` property to any playlist object returned from the parser that does not have it.

addresses #1205

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -6,7 +6,7 @@
  *
  */
 import resolveUrl from './resolve-url';
-import { mergeOptions, EventTarget } from 'video.js';
+import { mergeOptions, EventTarget, log } from 'video.js';
 import { isEnabled } from './playlist.js';
 import m3u8 from 'm3u8-parser';
 import window from 'global/window';
@@ -508,12 +508,17 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
           playlist = loader.master.playlists[i];
           loader.master.playlists[playlist.uri] = playlist;
           playlist.resolvedUri = resolveUrl(loader.master.uri, playlist.uri);
-          // Although the spec states an #EXT-X-STREAM-INF tag MUST have a
-          // BANDWIDTH attribute, we can play the stream without it. This means a poorly
-          // formated master playlist may not have an attribute list. An attributes
-          // property is added here to prevent undefined references when we encounter
-          // this scenario.
-          playlist.attributes = playlist.attributes || {};
+
+          if (!playlist.attributes) {
+            // Although the spec states an #EXT-X-STREAM-INF tag MUST have a
+            // BANDWIDTH attribute, we can play the stream without it. This means a poorly
+            // formatted master playlist may not have an attribute list. An attributes
+            // property is added here to prevent undefined references when we encounter
+            // this scenario.
+            playlist.attributes = {};
+
+            log.warn('Invalid playlist STREAM-INF detected. Missing attribute list.');
+          }
         }
 
         // resolve any media group URIs
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -517,7 +517,8 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
             // this scenario.
             playlist.attributes = {};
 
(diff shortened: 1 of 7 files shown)


src/playlist-loader.js
```diff
@@ -510,6 +508,12 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
           playlist = loader.master.playlists[i];
           loader.master.playlists[playlist.uri] = playlist;
           playlist.resolvedUri = resolveUrl(loader.master.uri, playlist.uri);
+          // Although the spec states an #EXT-X-STREAM-INF tag MUST have a
+          // BANDWIDTH attribute, we can play the stream without it. This means a poorly
+          // formated master playlist may not have an attribute list. An attributes
```
gesinger: formatted

src/playlist-loader.js
```diff
@@ -510,6 +508,12 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
+          // property is added here to prevent undefined references when we encounter
+          // this scenario.
+          playlist.attributes = playlist.attributes || {};
```
gesinger: Would be good to at least have a warning log here.

src/playlist-loader.js
```diff
@@ -510,6 +508,17 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
...
+
+          if (!playlist.attributes) {
+            // Although the spec states an #EXT-X-STREAM-INF tag MUST have a
+            // BANDWIDTH attribute, we can play the stream without it. This means a poorly
+            // formatted master playlist may not have an attribute list. An attributes
+            // property is added here to prevent undefined references when we encounter
+            // this scenario.
+            playlist.attributes = {};
+
+            log.warn('Invalid playlist STREAM-INF detected. Missing attribute list.');
```
gesinger: Since `BANDWIDTH` is the only required attribute, maybe it makes sense to just write something like "Invalid playlist STREAM-INF detected. Missing BANDWIDTH attribute."

src/rendition-mixin.js
```diff
@@ -54,21 +54,16 @@ class Representation {
                               .fastQualityChange_
                               .bind(hlsHandler.masterPlaylistController_);
 
-    // Carefully descend into the playlist's attributes since most
-    // properties are optional
-    if (playlist.attributes) {
-      let attributes = playlist.attributes;
+    // some playlist attributes are optional
+    if (playlist.attributes.RESOLUTION) {
+      let resolution = playlist.attributes.RESOLUTION;
```
OshinKaramian: Kind of a nitpick, by why let and not const?  If statement is the scope and the value doesn't seem to change.

test/playlist-loader.test.js
```diff
@@ -73,6 +73,24 @@ QUnit.test('moves to HAVE_MASTER after loading a master playlist', function(asse
...
+
+  this.requests.pop().respond(200, null,
+                              '#EXTM3U\n' +
+                              '#EXT-X-STREAM-INF:BANDWIDTH=1\n' +
+                              'video1/media.m3u8\n' +
+                              '#EXT-X-STREAM-INF:\n' +
+                              'video2/media.m3u8\n');
+
+  assert.ok(loader.master, 'infers a master playlist');
+  assert.ok(loader.master.playlists[0], 'parsed invalid stream');
```
gesinger: Technically, I think the second one is the one that is invalid, but it may be worth checking both.

test/playlist-loader.test.js
```diff
@@ -73,6 +73,24 @@ QUnit.test('moves to HAVE_MASTER after loading a master playlist', function(asse
...
+  assert.ok(loader.master.playlists[0].attributes, 'attached attributes property');
+  assert.equal(this.env.log.warn.calls, 1, 'logged a warning');
```
gesinger: Can also check the warning text.

Issue #1185, opened 2017-07-03
Duplicate In Manifest WebVTT Captions


triwav:
Hello everyone. First want to say thanks for all the work on getting WebVTT captions added to this library. I am seeing an issue with captions showing up multiple times in the player. I'm also seeing this same issue in JW Player however Safari seems to handle things correctly. Below is a link to a sample: http://brianleighty.com/videojs-hls-test (ignore the hard coded captions at the top) and here's a link to the m3u8 itself that you can view in Safari https://www.dvidshub.net/webcast/11812.m3u8?dvrTime=1499094660
I'm running 5.7.0 of videojs-contrib-hls and am currently running 5.20.1 of videojs. I can try 6.X but I don't think that should make a difference. In testing different streams, it seems the issue lies with the fact that we have captions spanning multiple segments. My work around that I haven't tested yet is to make captions not span multiple segments but just have it start at the same time in the next segment but thought I would report this as an issue as this should work but isn't. Let me know if you need any more details. Thanks.

gesinger:
Hey @triwav , thanks for reporting and for providing the example. I think it's doubling up because there are multiple instances of each caption. For instance, the first caption segment has 

```
WEBVTT
X-TIMESTAMP-MAP=MPEGTS:900000,LOCAL:00:00:10.000

01:24.360 --> 01:28.381
.
>> I THINK IT IS IN DIRECT
```

while the second has

```
WEBVTT
X-TIMESTAMP-MAP=MPEGTS:900000,LOCAL:00:00:10.000

01:24.360 --> 01:28.381
.
>> I THINK IT IS IN DIRECT

01:28.381 --> 01:34.531
CONTRADICTION TO WHAT WAS SAID
LAST WEEK, THAT HE DOESN'T
```

Notice that "I THINK IT IS IN DIRECT" is specified multiple times (as you said, because it crosses segment boundaries). We don't dedupe captions, as we just read them as if they are one big document that has been cut up into segments. That is probably the reason they are doubling up in the stream.

triwav:
Gotcha. Are there any plans to change this? Like I said I can work around it but that sounds like it could be handled better.

gesinger:
I think you're right @triwav . The spec reads:

"Each WebVTT Segment MUST contain all subtitle cues that are intended to be displayed during the period indicated by the segment EXTINF duration."

[...]

PR #1219, opened 2017-08-01
Fix: Double caption issue

shahlabs:
## Description
It fixes double-captions when a cue is across vtt file

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -281,7 +281,8 @@ export default class VTTSegmentLoader extends SegmentLoader {
 
     if (segmentInfo.cues.length) {
      // remove any overlapping cues to prevent doubling
-      this.remove(segmentInfo.cues[0].endTime, segmentInfo.cues[segmentInfo.cues.length - 1].endTime);
+      this.remove(segmentInfo.cues[0].endTime,
+                  segmentInfo.cues[segmentInfo.cues.length - 1].endTime);
     }
 
     segmentInfo.cues.forEach((cue) => {
--- a/src/vtt-segment-loader.js
+++ b/src/vtt-segment-loader.js
@@ -282,7 +282,7 @@ export default class VTTSegmentLoader extends SegmentLoader {
   /**
    * The below code helps in removing double captions
    */
-    if(segmentInfo.cues.length) {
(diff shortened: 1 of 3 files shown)


forbesjo:
For testing the advanced bipbop test stream can be used. The English caption track has "English subtitle 15" across two vtt files

forbesjo:
This fixes #1185

src/vtt-segment-loader.js
```diff
@@ -279,6 +279,16 @@ export default class VTTSegmentLoader extends SegmentLoader {
 
     this.mediaSecondsLoaded += segment.duration;
 
+  /**
+   * The below code helps in removing double captions
```
gesinger: For comments like this, it's probably better to put it right where the action is taking place, and you don't have to specify that the code is below. Maybe something like:

```
// remove any overlapping cues to prevent doubling
this.remove(firstCueEndTime, lastCueEndTime);
```

src/vtt-segment-loader.js
```diff
@@ -279,6 +279,16 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+   */
+    if (segmentInfo.cues.length) {
+      let lastCueEndTime = segmentInfo.cues[segmentInfo.cues.length - 1].endTime;
```
gesinger: Minor, but these can be `const` since they never change.

src/vtt-segment-loader.js
```diff
@@ -279,6 +279,16 @@ export default class VTTSegmentLoader extends SegmentLoader {
...
+      let firstCueEndTime = segmentInfo.cues[0].endTime;
+
+      this.remove(firstCueEndTime, lastCueEndTime);
```
gesinger: This is more of a preference thing, but you can remove the variables and just do it all in one line:

```
this.remove(segmentInfo.cues[0].endTime, segmentInfo.cues[segmentInfo.cues.length - 1].endTime)
```

shahlabs: The values do change for segmentInfo cues,  So I used let..But I did change the 3 lines code into 1 line and updated the comment above the code changes as you suggested

PR #1211, opened 2017-07-25
Prevent rendition switch loop due to inconsistent network/caching

gesinger:
Still need to do some manual testing, but wanted to put this up for comments.

## Requirements Checklist
- [X] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [X] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/liwecukasi/edit?html,output))
- [ ] Reviewed by Two Core Contributors

PR #1220, opened 2017-08-03
Blacklist playlist for 2 minutes on early abort to prevent cache loop

gesinger:
## Description
This is a simpler, more reliable, and overall better fix for loops due to inconsistent network/caching than https://github.com/videojs/videojs-contrib-hls/pull/1211

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -13,6 +13,8 @@ import worker from 'webworkify';
 import Decrypter from './decrypter-worker';
 import Config from './config';
 
+const ABORT_EARLY_BLACKLIST_SECONDS = 60 * 2;
+
 let Hls;
 
 // Default codec parameters if none were provided for video and/or audio
@@ -649,6 +651,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     this.mainSegmentLoader_.on('ended', () => {
       this.onEndOfStream();
+    });
+
+    this.mainSegmentLoader_.on('earlyabort', () => {
+      // skip playlist selection since a bandwidthupdate will trigger
+      this.blacklistCurrentPlaylist({
+        message: 'Aborted early because we don\'t have the bandwidth to complete the ' +
+          'request without rebuffering.'
+      }, ABORT_EARLY_BLACKLIST_SECONDS, true);
     });
 
     this.audioSegmentLoader_.on('ended', () => {
@@ -1296,8 +1306,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
    *
    * @param {Object=} error an optional error that may include the playlist
    * to blacklist
+   * @param {Number=} blacklistDuration an optional number of seconds to blacklist the
+   * playlist
+   * @param {Boolean=} skipPlaylistSelection an optional boolean to skip playlist
+   * selection
    */
-  blacklistCurrentPlaylist(error = {}) {
+  blacklistCurrentPlaylist(error = {}, blacklistDuration, skipPlaylistSelection = false) {
     let currentPlaylist;
     let nextPlaylist;
 
@@ -1330,9 +1344,16 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return this.masterPlaylistLoader_.load(isFinalRendition);
     }
     // Blacklist this playlist
-    currentPlaylist.excludeUntil = Date.now() + this.blacklistDuration * 1000;
+    currentPlaylist.excludeUntil = Date.now() +
+      (blacklistDuration ? blacklistDuration : this.blacklistDuration) * 1000;
     this.tech_.trigger('blacklistplaylist');
     this.tech_.trigger({type: 'usage', name: 'hls-rendition-blacklisted'});
+
+    if (skipPlaylistSelection) {
+      videojs.log.warn('Problem encountered with the current HLS playlist.' +
+                       (error.message ? ' ' + error.message : ''));
+      return;
+    }
 
     // Select a new playlist
     nextPlaylist = this.selectPlaylist();
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -654,11 +654,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('earlyabort', () => {
-      // skip playlist selection since a bandwidthupdate will trigger
       this.blacklistCurrentPlaylist({
-        message: 'Aborted early because we don\'t have the bandwidth to complete the ' +
+        message: 'Aborted early because there isn\'t enough bandwidth to complete the ' +
           'request without rebuffering.'
-      }, ABORT_EARLY_BLACKLIST_SECONDS, true);
+      }, ABORT_EARLY_BLACKLIST_SECONDS);
     });
 
     this.audioSegmentLoader_.on('ended', () => {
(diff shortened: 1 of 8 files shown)


src/segment-loader.js
```diff
@@ -15,6 +15,7 @@ import { minRebufferMaxBandwidthSelector } from './playlist-selectors';
 
 // in ms
 const CHECK_BUFFER_DELAY = 500;
+const ABORT_EARLY_BLACKLIST_MILLIS = 1000 * 60 * 2;
```
mjneil: 2 minutes seems very long to me

src/segment-loader.js
```diff
@@ -825,6 +826,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     // BANDWIDTH_VARIANCE and add one so the playlist selector does not exclude it
     this.bandwidth =
       switchCandidate.playlist.attributes.BANDWIDTH * Config.BANDWIDTH_VARIANCE + 1;
+    this.playlist_.excludeUntil = Date.now() + ABORT_EARLY_BLACKLIST_MILLIS;
```
mjneil: Should we log a warning like we do when blacklisting other playlists? Related, should we change this to be an event that can notify `MasterPlaylistController` that it needs to blacklist the playlist so that all blacklisting goes through the same path?

gesinger: Definitely, good call. I'll do both.

src/segment-loader.js
gesinger: It's a tough one, and I can see the argument from both sides. But if we don't have the bandwidth now to support the playlist, either because of inconsistent network or because of a low bandwidth network, but can support a different one, I can see this playlist being a poor choice for future selection, as it will likely cause more problems. Two minutes is an arbitrary selection, but seemed long enough to help prevent the segment loops. Maybe it is too long though. Do you think one minute might be better, or did you have another duration in mind?

mjneil: I agree that if we had to abort early because of poor network, it is likely that it will still be an issue in the future. My concern is that this would happen during startup or very early in playback, but then as we have substantial forward buffer, we could up-switch without causing rebuffering, but it may still be within that 2 minutes. This could just be a hypothetical situation that wouldn't actually play out in

src/master-playlist-controller.js
```diff
@@ -1296,8 +1306,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
    *
    * @param {Object=} error an optional error that may include the playlist
    * to blacklist
+   * @param {Number=} blacklistDuration an optional number of seconds to blacklist the
+   * playlist
+   * @param {Boolean=} skipPlaylistSelection an optional boolean to skip playlist
+   * selection
    */
-  blacklistCurrentPlaylist(error = {}) {
+  blacklistCurrentPlaylist(error = {}, blacklistDuration, skipPlaylistSelection = false) {
```
mjneil: is `skipPlaylistSelection` really necessary? I think it would be better to leave playlist selection here as is and just trigger `abortearly` and not `bandwidthupdate` from `segmentLoader`. If we are blacklisting the current playlist, I don't really see a scenario in which we would not want to switch from it. I think it's also reasonable to not trigger `bandwidthupdate` for the early abort case, as the `bandwidth` set in the segment loader after an early abort is not necessarily the bandwidth of the user, but a specific bandwidth to force playlist selection to select a certain variant. It's a bit of a lie for segment loader to claim it has an update to calculated bandwidth.

test/master-playlist-controller.test.js
```diff
@@ -1067,6 +1067,217 @@ function(assert) {
...
+  assert.equal(warnings[0],
+               'Problem encountered with the current HLS playlist. ' +
+                 'Aborted early because we don\'t have the bandwidth to complete the ' +
+                 'request without rebuffering.',
+               'warning message is correct');
+
+  videojs.log.warn = origWarn;
+});
+
+QUnit.test('does not get stuck in a loop due to inconsistent network/caching',
```
OshinKaramian: This test looks like a bit of a bear.  It might be helpful to do one of a few things to make things clearer to future explorers:

1. Find a way to break this test up into a few smaller tests... maybe evaluate each stage as its own test.  I don't know if qunit lets your group things as you can in [mocha](https://mochajs.org/#asynchronous-code) with describe blocks, but you could have something like:

```
describe('caching', () => {
  describe('does not get stuck in a loop due to inconsistent network/caching' () => {
    before(() => { // bootstrap data these tests use });
    function setUpStep1() => { ... };
    function setUpStep2() => { ... };

    it('tests the first step', () => {
     setupStep1();
     assert(x,y, 'output of step 1 looks good!');
    });

   it ('tests second step', () => {
     setupStep1();
     setupStep2();
     assert(x,y, 'output of step 2 looks good!');
   });
  })
})
```

This might yield more code, but the tests themselves might end up looking a bit clearer, since each is testing a distinct scenario with regards to your use case.  It might make debugging a bit easier (instead of the caching test as a whole breaking on an assert that you need to dig through, the test that broke should give you the information that you need to debug a little bit faster).

There is a library that could work (there could be others), I don't know if we want to introduce it here: https://github.com/square/qunit-bdd

2. If you don't want to do that, maybe adding a comment block on this test that describes what's going on would be useful.  Looking at this test I'm aware that there is some kind of loop that is caused by caching, but the use case is hard to tell without reading through the entire test.  If someone introduces something that breaks this test it might be difficult to debug whether something is supposed to break based on the change, or if their change broke functionality that we need to preserve.

3. Maybe find a way to break some things up into separate files.  The path of having a test per file seems to have kind of outgrown its usefulness here.  We could break the per-file tests into folders, then have test files for various scenarios (or something).  Basically breaking this part of the file into something like test/master-playlist-controller/caching.test.js.

src/master-playlist-controller.js
gesinger: Good call on the bandwidth being artificial, therefore not triggering. Will do that instead.

test/master-playlist-controller.test.js
gesinger: I agree that the test is too complicated. We discussed a bit via chat, about how this seems to combine unit and integration testing to create one inordinately long test, and that it would be better to split it up into functional pieces. However, for the time being, I think I'm going to go with option 2 (a comment block explaining the test steps) as this is a bit of an odd case, and we should eventually find a way to extract this kind of integration test into a separate test setup.

test/master-playlist-controller.test.js
```diff
@@ -1096,19 +1096,35 @@ QUnit.test('blacklists playlist on earlyabort', function(assert) {
...
+                 'request without rebuffering. Switching to another playlist.',
                'warning message is correct');
 
   videojs.log.warn = origWarn;
 });
 
 QUnit.test('does not get stuck in a loop due to inconsistent network/caching',
 function(assert) {
+  /*
+   * This test is a long one, but it is meant to follow a true path to a possible loop
```
mjneil: Maybe mention this behavior is amplified by browser and cdn caching

test/master-playlist-controller.test.js
```diff
@@ -1110,7 +1110,9 @@ QUnit.test('blacklists playlist on earlyabort', function(assert) {
 QUnit.test('does not get stuck in a loop due to inconsistent network/caching',
 function(assert) {
   /*
-   * This test is a long one, but it is meant to follow a true path to a possible loop
+   * This test is a long one, but it is meant to follow a true path to a possible loop.
+   * The reason for the loop is due to inconsistent network bandwidth, often caused or
+   * amplified by caching at the browser or edge server level.
```
mjneil: 🎸 🔈

PR #1223, opened 2017-08-08
various improvements related to abr switching algos

tchakabam:
## Description   

various improvements related to abr switching algos.

- We need to add to an EWMA-based algo

- We are centralizing applying of player width/height restrictions, but also are making these optional (as it depends on use-case specifics if one would like to have these or not) 

- We would like to allow a more clear and flexible 3rd party sub-implementation of the `PlaylistSelector` scheme, i.e ABR switching algos. We are exposing the proper playlistSelectors in the API and are harmonizing their interface. Also we are exposing playlist-filter-funcs, which can be used as tooling for 3rd parties to create own implementations (or creating further playlist-selectors within this codebase).

- we would like to monitor the estimator from outside with `stats` 

## Specific Changes proposed
 - put selectPlaylist setup in MasterPlaylistController constructor
    - set segments stats even on request error (to have latency "rtt")
    - segment-loader: add a latency average to the throughput stats
    - abr: in playlist-selectors rename in order to differentiate
    between filters and actual selectors (which of return a function not a playlist)
    - add an "ewma" bandwidth estimator based ona weighted rolling average
    - centralize where we apply restrictions related to player/screen width
    - make "automatic" restriction to player size optional
    - add optional max-widht/height/bandwidth restrictions for all playlist-selectors
    - expose all playlist-selector-factories in Hls static API in order to easily make use of them
    by setting the static STANDARD_PLAYLIST_SELECTOR
    - expose filterPlaylists seperately (to build selector easily outside of the lib)
    - expose computeDecayRateByHalfLife to compute meaningful decay values
    for all moving average implementations (ma and ewma).
    - add an "estimatedBandwidth" stat and private func setEstimatedBandwidth_ for the abr logics to notify estimations for metrics
    which playlist-selectors should call to make this value available

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -292,6 +292,12 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     Hls = externHls;
 
+    // `this` in selectPlaylist should be the HlsHandler for backwards
+    // compatibility with < v2
+
+    const playlistSelectorFactory = Hls.STANDARD_PLAYLIST_SELECTOR;
+
+    this.selectPlaylist = playlistSelectorFactory.call(tech.hls).bind(tech.hls);
     this.withCredentials = withCredentials;
     this.tech_ = tech;
     this.hls_ = tech.hls;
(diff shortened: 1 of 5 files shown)


tchakabam:
this now also includes the fix from: https://github.com/videojs/videojs-contrib-hls/pull/1207

forbesjo:

PR #1230, opened 2017-08-10
Fixing InvalidStateError in Win10 IE11

shahlabs:
## Description
Fixing InvalidStateError logged on console in Windows 10, IE11

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -60,6 +60,7 @@ export default class PlaybackWatcher {
       this.tech_.off('seekablechanged', fixesBadSeeksHandler);
       this.tech_.off('waiting', waitingHandler);
       this.tech_.off(timerCancelEvents, cancelTimerHandler);
+      this.tech_.off('canplay', canPlayHandler);
       if (this.checkCurrentTimeTimeout_) {
         window.clearTimeout(this.checkCurrentTimeTimeout_);
       }
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -51,7 +51,9 @@ export default class PlaybackWatcher {
     this.tech_.on('seekablechanged', fixesBadSeeksHandler);
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
-    this.monitorCurrentTime_();
+    if (this.tech_.hasStarted_) {
+      this.monitorCurrentTime_();
+    }
 
     // Define the dispose function to clean up our events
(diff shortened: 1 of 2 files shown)


src/playback-watcher.js
```diff
@@ -51,7 +51,9 @@ export default class PlaybackWatcher {
     this.tech_.on('seekablechanged', fixesBadSeeksHandler);
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
-    this.monitorCurrentTime_();
+    if (this.tech_.hasStarted_) {
```
gesinger: If the player has not started yet, but starts later, will PlaybackWatcher end up running?

mjneil: I don't think so. I think we would have to wait for either the `ready` or `canplay` event from the tech

src/playback-watcher.js
```diff
@@ -51,7 +51,9 @@ export default class PlaybackWatcher {
     this.tech_.on('seekablechanged', fixesBadSeeksHandler);
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
-    this.monitorCurrentTime_();
+    if (this.tech_.hasStarted_ || this.tech_.isReady_) {
```
mjneil: If the tech is not ready by the time this constructor runs, this will continue to be broken. You should setup an event listener for the `ready` event instead of checking the private variables on the tech. The tech has a `this.tech_.ready()` function that will either setup a listener for when the `ready` event will be fired, or will run your callback immediately if the tech is already `ready`.

src/playback-watcher.js
```diff
@@ -44,14 +44,15 @@ export default class PlaybackWatcher {
...
+    let canPlayHandler = () => this.monitorCurrentTime_();
     let waitingHandler = () => this.techWaiting_();
     let cancelTimerHandler = () => this.cancelTimer_();
     let fixesBadSeeksHandler = () => this.fixesBadSeeks_();
 
     this.tech_.on('seekablechanged', fixesBadSeeksHandler);
     this.tech_.on('waiting', waitingHandler);
     this.tech_.on(timerCancelEvents, cancelTimerHandler);
-    this.monitorCurrentTime_();
+    this.tech_.on('canplay', canPlayHandler);
```
mjneil: you should `off` the handler in the dispose method below like the other handlers

PR #1234, opened 2017-08-14
always activate loading in segment loaders after a seek

mjneil:
## Description
After the video has `ended` the player goes into a paused state. If you try and seek anywhere in the video after this, the player is still paused, so the `SegmentLoader`s do not being loading again. This results in the user seeing a loading spinner and black screen until they hit play. What should happen is the player begins loading segments at the new seek position and append to the buffer so the first frame of the position is displayed.

### DIFF
--- a/test/videojs-contrib-hls.test.js
+++ b/test/videojs-contrib-hls.test.js
@@ -2780,7 +2780,7 @@ QUnit.test('cleans up the buffer when loading VOD segments', function(assert) {
   this.clock.tick(1);
   this.player.currentTime(120);
   this.player.tech_.hls.mediaSource.sourceBuffers[0].trigger('updateend');
-  this.clock.tick(1);
+  this.clock.tick(2);
   this.standardXHRResponse(this.requests[3]);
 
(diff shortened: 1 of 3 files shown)


test/videojs-contrib-hls.test.js
```diff
@@ -2780,7 +2780,7 @@ QUnit.test('cleans up the buffer when loading VOD segments', function(assert) {
   this.clock.tick(1);
   this.player.currentTime(120);
   this.player.tech_.hls.mediaSource.sourceBuffers[0].trigger('updateend');
-  this.clock.tick(1);
+  this.clock.tick(2);
```
mjneil: This change was needed because previously, after an `updateend` event, `SegmentLoader.handleUpdateEnd_` would call `SegmentLoader.monitorBuffer_` which would setup a timeout to call `SegmentLoader.fillBuffer_` on the next clock tick, but with the addition of calling `load` on a seek, this would clear that timeout and set a new one for the next clock tick.

gesinger: This would be good to add as a comment above the clock tick (since clock ticks are generally pretty mysterious through our tests).

PR #1242, opened 2017-08-25
Keep track of approximate total bytes in SourceBuffer / Prevent quota exceeded with 4K streams

tchakabam:
## Description

    // With high-bitrate streams (like 4K etc, up to 16-20 Mbit/s)
    // we can reach a QuotaExceeded error easily within just 60 seconds
    // of buffer. Therefore we need to actively check the exact number
    // of bytes occupied in the buffer.
    // If we reach the limit, we are triming pre-emptively, enqueuing this action
    // to the source-updater before any further loading can happen.
    // Even if the back-buffer is alrady trimmed, we should wait before
    // we append more data if the buffer is still too occupied.
    //
    // We monitor the bytes in buffer (see checkBuffer_())
    // and drive the loading / triming actively
    // to make sure we do not reach this limit.

## Specific Changes proposed

* add method `totalBytesInBuffer` to `SourceBuffer`
* modifiy slightly `SourceBuffer` interface to pass `duration`
* use above method to limit loading in `SegmentLoader` when limit is exceeded
* move significant constants in `Config` module
* `trimBackBuffer_` improvements

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -602,7 +602,7 @@ export default class SegmentLoader extends videojs.EventTarget {
       return null;
     }
 
-    console.log('Total Mbytes in SourceBuffer:', (bufferedBytes / 1e6).toFixed(2));
+    this.logger_('Total Mbytes in SourceBuffer:', (bufferedBytes / 1e6).toFixed(2));
 
     // With high-bitrate streams (like 4K etc, up to 16-20 Mbit/s)
     // we can reach a QuotaExceeded error easily within just 60 seconds
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -595,9 +595,34 @@ export default class SegmentLoader extends videojs.EventTarget {
       lastBufferedEnd = buffered.end(buffered.length - 1);
     }
 
-    let bufferedTime = Math.max(0, lastBufferedEnd - currentTime);
+    const bufferedBytes = this.sourceUpdater_.totalBytesInBuffer();
+    const bufferedTime = Math.max(0, lastBufferedEnd - currentTime);
 
     if (!playlist.segments.length) {
+      return null;
+    }
+
+    console.log('Total Mbytes in SourceBuffer:', (bufferedBytes / 1e6).toFixed(2));
+
+    // With high-bitrate streams (like 4K etc, up to 16-20 Mbit/s)
+    // we can reach a QuotaExceeded error easily within just 60 seconds
+    // of buffer. Therefore we need to actively check the exact number
+    // of bytes occupied in the buffer.
+    // If we reach the limit, we are triming pre-emptively, enqueuing this action
+    // to the source-updater before any further loading can happen.
+    // Even if the back-buffer is alrady trimmed
+    //
+    // We also monitor the bytes in buffer (see checkBuffer_())
+    // and drive the loading / triming actively
+    // to make sure we do not reach this limit.
+    if (bufferedBytes >= Config.MAX_SOURCE_BUFFER_OCCUPATION_BYTES) {
+      // If we reached the limit let's trim the buffer already
+      // to make sure we free some space up, in case we reach some kind
+      // of jamming situation.
+      // We keep some back-buffer but reduce the allowed size significantly
+      // compared to the initial configuration.
+      this.logger_('SourceBuffer occupation exceeds limit, triming back-buffer and deferring further loading');
+      this.trimBackBuffer_(Config.BACK_BUFFER_LENGTH / 3);
       return null;
     }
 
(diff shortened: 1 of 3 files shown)


src/segment-loader.js
```diff
@@ -859,7 +885,18 @@ export default class SegmentLoader extends videojs.EventTarget {
...
-    this.trimBackBuffer_(segmentInfo);
+
+    // Chrome has a hard limit of 150MB of
+    // buffer and a very conservative "garbage collector"
+    // We manually clear out the old buffer to ensure
+    // we don't trigger the QuotaExceeded error
+    // on the source buffer during subsequent appends
+    //
+    // We also monitor the bytes in buffer (see checkBuffer_())
+    // and drive the loading / triming actively
```
gesinger: trimming

src/segment-loader.js
```diff
@@ -859,7 +885,18 @@ export default class SegmentLoader extends videojs.EventTarget {
   loadSegment_(segmentInfo) {
     this.state = 'WAITING';
     this.pendingSegment_ = segmentInfo;
-    this.trimBackBuffer_(segmentInfo);
+
+    // Chrome has a hard limit of 150MB of
```
gesinger: Not too big of a deal, but I think this comment still makes sense to be in the `trimBackBuffer` function, since that is where the logic of how much to clear happens. `loadSegment` just needs to know to call `trimBackBuffer`

src/segment-loader.js
```diff
@@ -1078,20 +1109,23 @@ export default class SegmentLoader extends videojs.EventTarget {
...
       }
     }
 
     segmentInfo.byteLength = segmentInfo.bytes.byteLength;
+
+    let segmentDuration;
+
     if (typeof segment.start === 'number' && typeof segment.end === 'number') {
-      this.mediaSecondsLoaded += segment.end - segment.start;
+      this.mediaSecondsLoaded += segmentDuration = segment.end - segment.start;
```
gesinger: Makes sense, but is a bit hard to read. Maybe the conditionals just set the value of `segmentDuration` (or a ternary is used), then `this.mediaSecondsLoaded += segmentDuration` exists after?

src/source-updater.js
```diff
@@ -47,13 +47,54 @@ export default class SourceUpdater {
...
+    this.bufferItemsList_ = [];
+
     if (mediaSource.readyState === 'closed') {
       mediaSource.addEventListener('sourceopen', createSourceBuffer);
     } else {
       createSourceBuffer();
     }
   }
 
+  appendToBufferInfoQueue_(timestampOffset, duration, byteLength) {
```
gesinger: Instead of putting all of the buffer info in `SourceUpdater` and keeping track of it, maybe it makes sense for us to add byte info to the segment metadata track https://github.com/videojs/videojs-contrib-hls#segment-metadata and we can make use of that.

squarebracket: For reference, the PR to include the required information in the segment metadata is #1210.

tchakabam:
Hey, thanks for the review!

And yes @squarebracket is right, if we combine both contributions, no need anymore for the rough estimation with the `/3`.

I hope I can progress on both of these this week! Thanks for the patience ;)

squarebracket:
Hey @tchakabam!

We've been talking in the slack channel about how to handle `QuoteExceededError`s in the most robust fashion. There's a consensus that we should catch the exception and dynamically change `BACK_BUFFER_LENGTH`, `MAX_GOAL_BUFFER_LENGTH`, and `MAX_SOURCE_BUFFER_OCCUPATION_BYTES` based on the amount of time and data in the buffer when the exception occurs. This way we could handle various buffer sizes without knowing them in advance, and adjust them on-the-fly if system memory pressures force them to decrease in size.

That solution would require some additional functionality to what you've got here, and a new PR to [contrib-media-sources](https://github.com/videojs/videojs-contrib-media-sources), since there's a `SourceBuffer` shim that's used when transmuxing from MPEG-TS. (It calls `appendBuffer` async, so it can't be caught by contrib-hls.)

Would that be something you're up for? I can walk you through the details if you're interested. No pressure if not; I can code up the PRs and credit you with the basis for the solution :)

Thanks again for taking the time to contribute and moving us forwards on a fix.

tchakabam:
@squarebracket Hey, thanks for the update on this.

These days I've got quite some new things on my plate, and unfortunately wont get to work on this one here anytime soon.

So please go on with that proposed solution :) Sounds very good to me :+1: 

Adapting dynamically is definitely favorable, 

and even more favoreable to leverage the contrib-media-sources layer to handle this stuff, as it will be a common concern for any user of the API :)  

Feel free to close or continue this here as you see fit.

squarebracket:
Sounds good :) thanks again for the initial PR.

I'll close this one once I've got a new one open.

fbeaudet:
Hi :)
I'm having this issue only in firefox, I think my samples are limit. Looking forward for that robust solution, gonna update as soon as it's out.
Thank you!

genelamarellis:
I know it's been a while, but any progress on this? Running into the exact same issue! :)

squarebracket:
Yes, actually! My apologies, I forgot to comment in this ticket.

These features are now being tracked in #1416. It would certainly be a big help to me if everyone interested in these features could go build the project with those changes and test it out with their sources / devices.

Closing this PR in favour of #1416.

PR #1257, opened 2017-09-14
Blacklist audio only playlists on probe if already playing video

gesinger:
## Description
When a master playlist doesn't contain codec info, we have to wait until we probe to know if a rendition is audio only. Because we can't switch from audio and video to audio only, blacklist that playlist if we're already playing video.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -51,7 +51,7 @@ export const illegalMediaSwitch = (loaderType, startingMedia, newSegmentMedia) =
   // Although these checks should most likely cover non 'main' types, for now it narrows
   // the scope of our checks.
   if (loaderType !== 'main' || !startingMedia || !newSegmentMedia) {
-    return;
+    return null;
   }
 
   if (!newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
@@ -69,6 +69,8 @@ export const illegalMediaSwitch = (loaderType, startingMedia, newSegmentMedia) =
       ' We can\'t switch to a stream with video from an audio only stream.' +
       ' To get rid of this message, please add codec information to the manifest.';
   }
+
+  return null;
 };
 
 /**
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -47,6 +47,50 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
 
 const finite = (num) => typeof num === 'number' && isFinite(num);
 
+export const illegalMediaSwitch = (loaderType, startingMedia, newSegmentMedia) => {
+  // Although these checks should most likely cover non 'main' types, for now it narrows
+  // the scope of our checks.
+  if (loaderType !== 'main' || !startingMedia || !newSegmentMedia) {
+    return;
+  }
+
+  if (!newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
+    return 'Neither audio nor video found in segment.';
+  }
+
+  if (startingMedia.containsAudio && startingMedia.containsVideo &&
+      newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
+    return 'Only audio found in segment when we expected audio and video.' +
+      ' We can\'t switch to audio only from audio and video.' +
+      ' To get rid of this message, please add codec information to the manifest.';
+  }
+
+  if (!startingMedia.containsAudio && startingMedia.containsVideo &&
+      newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
+    return 'Only audio found in segment when we expected video only.' +
+      ' We can\'t switch to audio only from video only.' +
+      ' To get rid of this message, please add codec information to the manifest.';
+  }
+
+  if (startingMedia.containsAudio && !startingMedia.containsVideo) {
+    let errorMessage;
+
+    if (newSegmentMedia.containsAudio && newSegmentMedia.containsVideo) {
+      errorMessage = 'Audio and video found in segment when we expected audio only.';
+    }
+
+    if (!newSegmentMedia.containsAudio && newSegmentMedia.containsVideo) {
+      errorMessage =
+        'Only video found in segment when we expected audio only.';
+    }
+
+    if (errorMessage) {
+      return errorMessage + ' We can\'t switch since we started on audio only.' +
+        ' To get rid of this message, please add codec information to the manifest.';
+    }
+  }
+};
+
 /**
  * An object that manages segment loading and appending.
  *
@@ -84,7 +128,7 @@ export default class SegmentLoader extends videojs.EventTarget {
     this.mediaSource_ = settings.mediaSource;
     this.hls_ = settings.hls;
     this.loaderType_ = settings.loaderType;
-    this.isUsingVideo_ = void 0;
+    this.startingMedia_ = void 0;
     this.segmentMetadataTrack_ = settings.segmentMetadataTrack;
     this.goalBufferLength_ = settings.goalBufferLength;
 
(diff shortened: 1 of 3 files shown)


test/segment-loader.test.js
```diff
@@ -603,5 +603,125 @@ QUnit.module('SegmentLoader', function(hooks) {
...
+          source: 'audio'
+        };
+      };
+      this.requests[0].response = new Uint8Array(10).buffer;
+      this.requests.shift().respond(200, null, '');
+
+      assert.equal(errors.length, 0, 'no errors');
+    });
+
+    QUnit.test('does not error when going from audio only to avoid and video',
```
shahlabs: It should be audio and video

forbesjo:
Looks good to me

test/segment-loader.test.js
```diff
@@ -603,5 +603,125 @@ QUnit.module('SegmentLoader', function(hooks) {
...
+          source: 'audio'
+        };
+      };
+      this.requests[0].response = new Uint8Array(10).buffer;
+      this.requests.shift().respond(200, null, '');
+
+      assert.equal(errors.length, 0, 'no errors');
+    });
+
+    QUnit.test('does not error when going from audio only to audio and video',
```
mjneil: Do we support this?

shahlabs: If I switch from audio only to audio/video rendition it gives me error as VIDEOJS: ERROR: DOMException: Failed to set the 'duration' property on 'MediaSource': The 'updating' attribute is true on one or more of this MediaSource's SourceBuffers.

gesinger: I'm not sure if we support it, but I put the test in since we don't want to throw the error there for now (until we re-consider). I can remove the test case if we want, until we are sure about behavior.

Seems from @shahlabs ' testing we may want to consider an error anyway.

mjneil: Well according to the MSE spec, going from audio only to audio + video will be dependent on browser

>For example, a user agent may throw a QuotaExceededError exception if the media element has reached the HAVE_METADATA readyState. This can occur if the user agent's media engine does not support adding more tracks during playback.
https://w3c.github.io/media-source/#dom-mediasource-addsourcebuffer

So if we want to support this, we should make sure which browsers support it, and handle the error in browsers that dont. I think its preferable to just not allow switching between audio only and video+audio and vice versa.

Might be worth seeing what happens when going from video only to audio+video as well, though I'm not sure if that is something anyone is trying to do in their manifests.

src/segment-loader.js
```diff
@@ -1048,12 +1049,35 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // source buffers.
+    if (typeof this.isUsingVideo_ === 'undefined' && timingInfo) {
+      this.isUsingVideo_ = timingInfo.source === 'video';
+    }
+
+    // Although we support starting on audio only, and continuing playback from there, we
+    // don't support switching from audio and video to audio only. If the codecs were in
+    // the master manifest, then the renditions incompatible with the first rendition
+    // would already be blacklisted.
+    if (this.isUsingVideo_ &&
```
mjneil: WRT the other discussion on going from audio only to video+audio

If we decide that we want to support that, this will NOT trigger an error in the event that you start audio only, switch to video+audio, and then try to switch back to audio only since `this.isUsingVideo_` will be set to `false` from the initial audio only timing source

src/segment-loader.js
```diff
@@ -47,6 +47,50 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
...
+  if (startingMedia.containsAudio && startingMedia.containsVideo &&
+      newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
+    return 'Only audio found in segment when we expected audio and video.' +
+      ' We can\'t switch to audio only from audio and video.' +
+      ' To get rid of this message, please add codec information to the manifest.';
+  }
+
+  if (!startingMedia.containsAudio && startingMedia.containsVideo &&
+      newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
+    return 'Only audio found in segment when we expected video only.' +
```
mjneil: a bit misleading saying we expected "video only" as we can go from starting with video only to video + audio (sort of, if starting was video only because there was a separate audio segment loader and not because there was just only video stream)

src/segment-loader.js
```diff
@@ -47,6 +47,50 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
...
+  // the scope of our checks.
+  if (loaderType !== 'main' || !startingMedia || !newSegmentMedia) {
+    return;
+  }
+
+  if (!newSegmentMedia.containsAudio && !newSegmentMedia.containsVideo) {
+    return 'Neither audio nor video found in segment.';
+  }
+
+  if (startingMedia.containsAudio && startingMedia.containsVideo &&
```
mjneil: I think we could simplify this and still give informative messages
```
if (startingMedia.containsVideo && !newSegmentMedia.containsVideo) {
  return 'Only audio found in segment when we expected video.' +
             ' We can\'t switch to audio only from a stream that had video' +
             ' To get rid of this message, please add codec information to the manifest.';
}
if (!startingMedia.containsVideo && newSegmentMedia.containsVideo) { 
  return 'Video found in segment when we expected only audio.' +
             ' We can\'t switch to a Video stream from a stream that had only audio' +
             ' To get rid of this message, please add codec information to the manifest.';
}
```

test/segment-loader.test.js
```diff
@@ -20,6 +23,75 @@ const ogAddSegmentMetadataCue_ = SegmentLoader.prototype.addSegmentMetadataCue_;
...
+  startingMedia = { containsAudio: true, containsVideo: true };
+  newSegmentMedia = { containsAudio: true, containsVideo: false };
+  assert.equal(illegalMediaSwitch('main', startingMedia, newSegmentMedia),
+               'Only audio found in segment when we expected video.' +
+               ' We can\'t switch to audio only from a stream that had video.' +
+               ' To get rid of this message, please add codec information to the' +
+               ' manifest.',
+               'error when muxed to audio only');
+
+  startingMedia = { containsAudio: true, containsVideo: true };
```
mjneil: Maybe add a test that has `startingMedia = { containsAudio: true, containsVideo: false };` and `newSegmentMedia = { containsAudio: false, containsVideo: false };` just to confirm it also returns `'Neither audio nor video found in segment.'`

test/segment-loader.test.js
```diff
@@ -603,5 +675,92 @@ QUnit.module('SegmentLoader', function(hooks) {
       assert.ok(!playlistUpdated.segments[0].end,
                 'no end info for first segment of new playlist');
     });
+
+    QUnit.test('errors when trying to switch from audio and video to audio only',
```
mjneil: Maybe another test like this but switching from audio only to video and audio

test/segment-loader.test.js
```diff
@@ -20,6 +23,75 @@ const ogAddSegmentMetadataCue_ = SegmentLoader.prototype.addSegmentMetadataCue_;
 
 SegmentLoader.prototype.addSegmentMetadataCue_ = function() {};
 
+QUnit.module('SegmentLoader Isolated Functions');
+
+QUnit.test('illegalMediaSwitch detects illegal media switches', function(assert) {
```
mjneil: I think it would be good to adda test that passes `type === 'audio'` since the audio segment loader will also be calling this function

src/segment-loader.js
```diff
@@ -1048,12 +1073,36 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+        // certain that all streams will provide it.
+        (timingInfo.containsAudio || timingInfo.containsVideo)) {
+      this.startingMedia_ = {
+        containsAudio: timingInfo.containsAudio,
+        containsVideo: timingInfo.containsVideo
+      };
+    }
 
-    this.syncController_.probeSegmentInfo(segmentInfo);
+    let illegalMediaSwitchError =
```
mjneil: `const`

src/segment-loader.js
```diff
@@ -47,6 +47,30 @@ const detectEndOfStream = function(playlist, mediaSource, segmentIndex) {
 
 const finite = (num) => typeof num === 'number' && isFinite(num);
 
+export const illegalMediaSwitch = (loaderType, startingMedia, newSegmentMedia) => {
+  // Although these checks should most likely cover non 'main' types, for now it narrows
+  // the scope of our checks.
+  if (loaderType !== 'main' || !startingMedia || !newSegmentMedia) {
+    return;
```
mjneil: Since you make use of the return value, it might be a good idea to explicitly return `null` or `false`

test/segment-loader.test.js
mjneil: Looks like you forgot to updated `containsVideo` here, this test is exactly the same as the one on line 55. Also would be nice to move the two tests for `'Neither audio nor video found in segment.'` next to each other just for reading through, but not necessary

PR #1258, opened 2017-09-17
enable node_modules caching on travis

thijstriemstra:


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "prebuild": "npm run clean",
-    "build": "npm-run-all -p build:*",
```
forbesjo: This pattern runs `build:js` and `build:test` in parallel, it will ignore the third level
See https://github.com/mysticatea/npm-run-all/blob/master/docs/npm-run-all.md#glob-like-pattern-matching-for-script-names

thijstriemstra: Ok in my setup it's a little different I notice now, and you're right so ignore this change.

thijstriemstra:
@forbesjo should i make a separate pr for the caching or can you just pick the caching-related commit?

forbesjo:
We'll add the caching to #1322, @gesinger could you add this to your PR? Here's the related blog post https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/

gesinger:
Thanks for the contribution @thijstriemstra . It's now merged as part of https://github.com/videojs/videojs-contrib-hls/pull/1322 . Thanks again!

PR #1259, opened 2017-09-19
Fix video corruption on rendition switches in IE11 Win 8.1+ and Edge 

mjneil:
## Description
Some browsers (cough IE11 and Edge cough) have a more simple, though still spec compliant, implementation of MediaSourceExtensions compared to other browsers. The main issue with this simpler implementation is that MSE will drop data until the next key frame that comes after the time already sent to the decoder. What this means is you cannot append content that overlaps with the playhead (something videojs-contrib-hls relies on for faster mid-segment quality switching) without the risk of visual corruption since the frames needed to decode the new content at the playhead are dropped before being sent to the decoder. 

REQUIRES THE FOLLOWING FOR FULL SUPPORT
https://github.com/videojs/videojs-contrib-media-sources/pull/164
https://github.com/videojs/mux.js/pull/162

## Specific Changes proposed
 * Before appending each segment, `hls-segment-time-mapping` is triggered on the `tech` for `HtmlMediaSource` from videojs-contrib-media-sources to listen to so that the `VirtualSourceBuffer` can properly convert display time to TS time and trim segment appends accordingly.
    * This approach is used so that if native media sources are being used, the event won't have any effect and things will continue working normally.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -612,6 +612,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.tech_.trigger('hls-reset');
     });
 
+    this.mainSegmentLoader_.on('segmenttimemapping', (event) => {
+      this.tech_.trigger({
+        type: 'hls-segment-time-mapping',
+        mapping: event.mapping
+      });
+    });
+
     this.audioSegmentLoader_.on('ended', () => {
       this.onEndOfStream();
     });
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -609,10 +609,14 @@ export class MasterPlaylistController extends videojs.EventTarget {
     });
 
     this.mainSegmentLoader_.on('reseteverything', () => {
+      // If playing an MTS stream, a videojs.MediaSource is listening for
+      // hls-reset to reset caption parsing state in the transmuxer
       this.tech_.trigger('hls-reset');
     });
 
     this.mainSegmentLoader_.on('segmenttimemapping', (event) => {
+      // If playing an MTS stream in html, a videojs.MediaSource is listening for
+      // hls-segment-time-mapping update its internal mapping of stream to display time
       this.tech_.trigger({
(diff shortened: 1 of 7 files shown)


src/master-playlist-controller.js
```diff
@@ -612,6 +612,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.tech_.trigger('hls-reset');
     });
 
+    this.mainSegmentLoader_.on('segmenttimemapping', (event) => {
```
gesinger: I know we discussed it, but I'm still uneasy about using the event to communicate with videojs-contrib-media-sources. Although it's better not to add non-standard APIs to videojs-contrib-media-sources, making an event is essentially creating an API for it anyway, and makes the code harder to follow. Although it does help in cases where we don't want to fail on native media sources, I'm thinking that it might be better for us to have a specific method in source updater and contrib-media-sources to accept the data, just to make things easier to follow. Eventually contrib-media-sources will go away, removing the issue.

I do understand the arguments for using the event though (in that we aren't adding a "supported" non standard API to videojs-contrib-media-sources). If we do go with the event route, we should at least add some comments here and in videojs-contrib-media-sources to more easily understand how it is being passed between the two projects.

src/segment-loader.js
```diff
@@ -1120,6 +1118,15 @@ export default class SegmentLoader extends videojs.EventTarget {
       this.trigger('timestampoffset');
     }
 
+    if (timingInfo && timingInfo.hasMapping) {
```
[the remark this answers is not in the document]
mjneil: made a function `mappingForTimeline`

src/master-playlist-controller.js
mjneil: I do understand that the event is more difficult to follow than a direct api, but since we plan on moving away from contrib-media-sources (sooner the better eh) and we already use this event approach else where, I added comments so we can save time on testing a refactor. I've also added comments in the contrib-media-sources pr

PR #1260, opened 2017-09-19
Seek to seekable start when before the seekable window

forbesjo:
## Description
Seek to seekable start when before the seekable window

## Specific Changes proposed
Instead of seeking to live when doing a bad seek (for example if the seekable shifts while scrubbing) we'll snap to the beginning of the live window.

## Requirements Checklist

### DIFF
--- a/src/playback-watcher.js
+++ b/src/playback-watcher.js
@@ -154,33 +154,31 @@ export default class PlaybackWatcher {
    * @private
    */
   fixesBadSeeks_() {
-    let seekable = this.seekable();
-    let currentTime = this.tech_.currentTime();
+    const seeking = this.tech_.seeking();
+    const seekable = this.seekable();
+    const seekableStart = seekable.start(0);
+    const seekableEnd = seekable.end(seekable.length - 1);
+    const currentTime = this.tech_.currentTime();
+    let seekTo;
 
-    if (this.tech_.seeking()) {
-      if (this.afterSeekableWindow_(seekable, currentTime)) {
-        let seekableEnd = seekable.end(seekable.length - 1);
+    if (seeking && this.afterSeekableWindow_(seekable, currentTime)) {
+      // sync to live point (if VOD, our seekable was updated and we're simply adjusting)
+      seekTo = seekableEnd;
+    }
 
-        // sync to live point (if VOD, our seekable was updated and we're simply adjusting)
-        this.logger_(`Trying to seek outside of seekable at time ${currentTime} with ` +
-                    `seekable range ${Ranges.printableRange(seekable)}. Seeking to ` +
-                    `${seekableEnd}.`);
-        this.tech_.setCurrentTime(seekableEnd);
-        return true;
-      }
+    if (seeking && this.beforeSeekableWindow_(seekable, currentTime)) {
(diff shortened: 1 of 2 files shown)


src/playback-watcher.js
```diff
@@ -157,16 +157,30 @@ export default class PlaybackWatcher {
...
-                   `seekable range ${Ranges.printableRange(seekable)}. Seeking to ` +
-                   `${seekableEnd}.`);
-      this.tech_.setCurrentTime(seekableEnd);
-      return true;
+    if (this.tech_.seeking()) {
+      if (this.afterSeekableWindow_(seekable, currentTime)) {
+        let seekableEnd = seekable.end(seekable.length - 1);
+
+        // sync to live point (if VOD, our seekable was updated and we're simply adjusting)
+        this.logger_(`Trying to seek outside of seekable at time ${currentTime} with ` +
```
mjneil: Since the much of this is the same, maybe just set a seekTime variable and do the log/seek at the end if it was set
```
let seekTime = NaN;
if (after) seekTime = seekable.end;
if (before) seekTime = seekable.start + 0.1;
if (!isNaN(seekTime)) log; setCurrentTime(seekTime);
```

forbesjo: Yep will do

mjneil:
Github won't let me comment on the line because no changes were made, but line 468 of playback-watcher.test.js test description is `'seeks to live point if we try to seek outside of seekable'` which is not quite true anymore since it may seek to the start of the window depending on where you are trying to seek

src/playback-watcher.js
```diff
@@ -154,18 +154,30 @@ export default class PlaybackWatcher {
    * @private
    */
   fixesBadSeeks_() {
-    let seekable = this.seekable();
-    let currentTime = this.tech_.currentTime();
+    const seeking = this.tech_.seeking();
+    const seekable = this.seekable();
+    const seekableStart = seekable.start(0);
```
mjneil: `seekable` can have a length of 0, in which case this will error. Both `beforeSeekableWindow_` and `afterSeekableWindow_` check the length and return false if it is 0, so you could not create these const here and just access `seekable.start(0)` if you get into the appropriate if block. Otherwise you will have to check the length here as well

forbesjo: I'll move these down to their appropriate if blocks

src/playback-watcher.js
```diff
@@ -154,18 +154,32 @@ export default class PlaybackWatcher {
...
+    const seeking = this.tech_.seeking();
+    const seekable = this.seekable();
+    const currentTime = this.tech_.currentTime();
+    let seekTo;
 
-    if (this.tech_.seeking() &&
-        this.outsideOfSeekableWindow_(seekable, currentTime)) {
-      let seekableEnd = seekable.end(seekable.length - 1);
+    if (seeking && this.afterSeekableWindow_(seekable, currentTime)) {
+      const seekableStart = seekable.start(0);
```
mjneil: you switched up `seekableStart` and `seekableEnd`

forbesjo: doh

forbesjo: At least the tests caught it

PR #1265, opened 2017-09-25
Playlist loader cleanup

gesinger:
## Description
ES6ified and cleaned up a couple functions and added tests for them.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -168,8 +168,6 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
   }
 
   playlistRequestError = function(xhr, url, startingState) {
-    loader.setBandwidth(request || xhr);
-
     // any in-flight request is now finished
     request = null;
 
@@ -194,8 +192,6 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
     let parser;
     let refreshDelay;
     let update;
-
-    loader.setBandwidth(request || xhr);
 
     // any in-flight request is now finished
     request = null;
@@ -403,13 +399,6 @@ const PlaylistLoader = function(srcUrl, hls, withCredentials) {
         loader.trigger('mediachange');
       }
     });
-  };
-
-  /**
-   * set the bandwidth on an xhr to the bandwidth on the playlist
-   */
-  loader.setBandwidth = function(xhr) {
-    loader.bandwidth = xhr.bandwidth;
   };
 
   // live playlist staleness timeout
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -26,7 +26,7 @@ import window from 'global/window';
   * playlists.
   * @return a list of merged segment objects
   */
-const updateSegments = function(original, update, offset) {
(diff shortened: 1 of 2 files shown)


src/playlist-loader.js
```diff
@@ -52,59 +50,58 @@ const updateSegments = function(original, update, offset) {
...
+    if (segment.key && !segment.key.resolvedUri) {
+      segment.key.resolvedUri = resolveUrl(mergedPlaylist.resolvedUri, segment.key.uri);
+    }
+    if (segment.map && !segment.map.resolvedUri) {
+      segment.map.resolvedUri = resolveUrl(mergedPlaylist.resolvedUri, segment.map.uri);
+    }
+  });
+
+  // TODO Right now in the playlists array there are two references to each playlist, one
+  // that is referenced by index, and one by URI. The index reference may no longer be
```
mjneil: I'm pretty sure everywhere within contrib-hls uses playlists as an array and would require that the index based referenced be updated in order to get the new segments

gesinger: It might be worth considering, but can take out the TODO for now so it doesn't produce excess noise in the linter.

src/playlist-loader.js
```diff
@@ -52,59 +50,82 @@ const updateSegments = function(original, update, offset) {
   * master playlist with the updated media playlist merged in, or
   * null if the merge produced no change.
   */
-const updateMaster = function(master, media) {
-  let changed = false;
-  let result = mergeOptions(master, {});
+export const updateMaster = function(master, media) {
```
mjneil: surprised you didn't use arrow notation here

test/playlist-loader.test.js
```diff
@@ -655,57 +655,50 @@ QUnit.test('resolveMediaGroupUris resolves media group URIs', function(assert) {
...
           }
         },
-        AUDIO2: {
-          LOW: {
-            uri: 'audio-2-low-uri'
-          },
-          HIGH: {
-            uri: 'audio-2-high-uri'
+        high: {
+          main: {},
```
mjneil: not necessary, but if you want to cover another edge case, you can have a muxed audio label without a uri in one audio group, and the same labeled audio in another group can have unmuxed so it would have a uri. Playlist loader doesn't care about this detail though, its taken into account in media-groups, but incase you wanted another test case

test/playlist-loader.test.js
```diff
@@ -28,6 +34,745 @@ QUnit.module('Playlist Loader', {
...
 
+QUnit.test('updateSegments copies over properties', function(assert) {
+  assert.deepEqual(
+    [
+      { uri: 'test-uri-0', startTime: 0, endTime: 10 },
+      {
+        uri: 'test-uri-1',
+        startTime: 10,
+        endTime: 20,
+        attributes: { BANDWIDTH: 99, height: 4 }
```
mjneil: You don't need to change these because whats really being tested is object copy and merging, which doesn't care about the content, but it was a little confusing to see a combination of segment and playlist objects being used to test `updateSegments`.

PR #1266, opened 2017-09-29
fix InvalidStateError for live playback in IE11

mjneil:
Live streams in IE11 throw an `InvalidStateError` for trying to set the `currentTime` while the `readyState` is `0`.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -698,6 +698,8 @@ export class MasterPlaylistController extends videojs.EventTarget {
         // 3) the player has not started playing
         !this.hasPlayed_()) {
 
+      let setHasPlayed = false;
+
       // when the video is a live stream
       if (!media.endList) {
         this.trigger('firstplay');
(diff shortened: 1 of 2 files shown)


src/master-playlist-controller.js
```diff
@@ -893,14 +917,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     }
 
     // In flash playback, the segment loaders should be reset on every seek, even
-    // in buffer seeks
-    const isFlash =
-      (this.mode_ === 'flash') ||
-      (this.mode_ === 'auto' && !videojs.MediaSource.supportsNativeMediaSources());
```
mjneil: I snuck this change into this PR because the mode can only be `flash` or `html5`, so checking for the case of `auto` is unnecessary.

src/master-playlist-controller.js
```diff
@@ -698,17 +698,41 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+            // readyState is 0, so it must be delayed.
+            this.tech_.one('loadedmetadata', () => {
+              // we hold off on setting hasPlayed on true until after we seek to prevent
+              // segment loaders from requesting more than 1 segment before the seek to
+              // live point
+              this.tech_.setCurrentTime(seekable.end(0));
+              this.hasPlayed_ = () => true;
+            });
+          } else {
+            setHasPlayed = true;
```
OshinKaramian: Why is this set here, and not where hasPlayed_ is set?  You could move this into the loadedmetadata function above, and set it on line 733 as well.  

I'm mostly suggesting this because I think it will be easier from a maintenance standpoint.

mjneil: I'm not sure I follow. This function `setupFirstPlay` is called when `play` is triggered until `this.hasPlayed_()` returns `true`. Currently, with a VOD, we update `hasPlayed_` to return true the first time we get a `play` call and we've loaded the media. Currently in a live stream, we update `hasPlayed_` to true the first time we get a `play` call, we've loaded the media, and have calculated seekable, so that we can seek to the live point and that segmentloader knows it can start filling the buffer. Since now we have to wait for `loadedmetadata` from the tech in IE11 before we can seek, we wait to update `hasPlayed_` until we've gotten `loadedmetadata` which happens after we append the first bytes to the source buffer. Since this is asynchronous, this local var `setHasPlayed` is used to know if we can set `hasPlayed_` to true yet. Does this answer your question?

Issue #1267, opened 2017-10-03
VIDEOJS: ERROR: TypeError: Cannot read property 'playlist' of undefined


ahmadawada11:
I got a exception when I am changing the resolution several time. For example, if my list has 2 resolutions which are A and B. If I selected A and then B, and then selected A the exception will be thrown as the following:

VIDEOJS: ERROR: TypeError: Cannot read property 'playlist' of undefined

I am using videojs version 6.2.7, contrib-quality-level, hls-source-selector and hls version 5.11.0 plugins.

 this error doesn't happen with hls version 5.7.0

ahmadawada11:
the changes solved the issue did this fix included in v5.11.1?

mjneil:
No sorry, the fix is still being reviewed/tested

mjneil:
The fix has been released in v5.12.0. Closing this issue for now, please re-open or submit a new one if you continue to experience issues.

Issue #1206, opened 2017-07-24
undefined has no properties


roweldeguzman:
VideoJs version 6.2.0
Videojs-contrib-hls.js version 5.8.2

![video-js-hls-error](https://user-images.githubusercontent.com/8720937/28518523-bf64a538-709a-11e7-8fdf-070278c4d99c.png)

gesinger:
Hey @roweldeguzman , thanks for the issue, however, it is hard to debug without a sample page and information about the issue. Would you be able to provide a sample page showing the issue using our [starter template](http://jsbin.com/vokipos/8/edit?html,console,output), as well as fill out the information in our [issue template](https://github.com/videojs/videojs-contrib-hls/issues/new)? Thank you.

roweldeguzman:
hi @gesinger ,

seems this error only occur when in live broadcasting. VOD has no error showing.

gesinger:
Do you have a link to your live stream @roweldeguzman , and steps to reproduce? Thanks.

roweldeguzman:
Hi @gesinger ,

here is jsbin [http://jsbin.com/viwoga/edit?html,output](url) to show the error in console. for Live just wait a few seconds after you play the video.

mjneil:
It looks like your stream has audio that is unsupported by chrome
`window.MediaSource.isTypeSupported('video/mp4; codecs="avc1.42001f,mp4a.40.34"')`
which prompts contrib-hls to blacklist the playlist, but since you only have a single variant, there are no enabled playlists for the selector to pick from. #1207 would fix the error from occuring, but audio will continue to not work unless you update your encoder settings

mjneil:
You should no longer see the undefined error in v5.12.0. Closing this issue for now, please re-open or submit a new one if you continue to experience issues.

PR #1269, opened 2017-10-03
Fine tune blacklist handling

mjneil:
## Description
Currently it doesn't really matter whether a representation is disabled through user interaction via the representations api, or blacklisted internally because of network or playback errors. This can cause undefined references in our playlist selector (https://github.com/videojs/videojs-contrib-hls/issues/1206 https://github.com/videojs/videojs-contrib-hls/issues/1267)  when whatever playlists have not been manually disabled are all blacklisted internally due to errors.

## Specific Changes proposed
* never allow playlist selector to select a playlist that has been permanently blacklisted due to incompatible configuration
* When filtering playlists within the playlist selectors, if there are no enabled playlists (i.e. not blacklisted internally AND not disabled by the user) available, then fall back to using the list of playlists not disabled by the user regardless of blacklist state.
* make sure playlists blacklisted from an illegal media switch is permanently blacklisted, as there is no reason to try it again at a later time.
* The representation api will return a list that filters out just incompatible playlists instead of both incompatible playlists and temporary blacklisted playlists.

### DIFF
--- a/src/playlist.js
+++ b/src/playlist.js
@@ -399,6 +399,18 @@ export const isBlacklisted = function(playlist) {
 };
 
 /**
+ * Check whether the playlist is compatible with current playback configuration or has
+ * been blacklisted permanently for being incompatible.
+ *
+ * @param {Object} playlist the media playlist object
+ * @return {boolean} whether the playlist is incompatible or not
+ * @function isIncompatible
+ */
+export const isIncompatible = function(playlist) {
+  return playlist.excludeUntil && playlist.excludeUntil === Infinity;
+};
+
+/**
  * Check whether the playlist is enabled or not.
  *
  * @param {Object} playlist the media playlist object
@@ -409,6 +421,17 @@ export const isEnabled = function(playlist) {
   const blacklisted = isBlacklisted(playlist);
 
   return (!playlist.disabled && !blacklisted);
+};
+
+/**
+ * Check whether the playlist has been manually disabled through the representations api.
+ *
+ * @param {Object} playlist the media playlist object
+ * @return {boolean} whether the playlist is disabled manually or not
+ * @function isDisabled
+ */
+export const isDisabled = function(playlist) {
+  return playlist.disabled;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


test/playlist.test.js
```diff
@@ -686,6 +686,95 @@ QUnit.test('estimates segment request time based on bandwidth', function(assert)
...
+  playlist.excludeUntil = Date.now() - 1;
+  assert.ok(Playlist.isEnabled(playlist), 'playlist is enabled');
+
+  playlist.excludeUntil = Date.now() + 9999;
+  assert.notOk(Playlist.isEnabled(playlist), 'playlist is not enabled');
+
+  playlist.excludeUntil = Infinity;
+  assert.notOk(Playlist.isEnabled(playlist), 'playlist is not enabled');
+
+  delete playlist.excludeUntil;
```
OshinKaramian: Could we maybe break this test up in some way?  Maybe one for if playerlist.disabled is set to true? Or maybe based on the expectation (should be enabled vs should not be enabled)?

It also might be safer to not maintain the `const playlist` here and instead pass the raw values for `playlist` in.  As an example:

```
playlist.excludeUntil = Date.now() - 1;
assert.ok(Playlist.isEnabled(playlist), 'playlist is enabled');
```

Just becomes:
```
assert.ok(Playlist.isEnabled({ excludeUntil: Date.now() - 1 }), 'playlist is enabled');
```

This way you don't need to worry about deleting things and the order of the tests won't matter.

test/playlist.test.js
```diff
@@ -686,6 +686,95 @@ QUnit.test('estimates segment request time based on bandwidth', function(assert)
...
+  afterEach() {
+    this.env.restore();
+  }
+});
+
+QUnit.test('determines if a playlist is incompatible', function(assert) {
+  // incompatible means that the playlist was blacklisted due to incompatible
+  // configuration e.g. audio only stream when trying to playback audio and video.
+  // incompaatibility is denoted by a blacklist of Infinity.
+  const playlist = {};
```
OshinKaramian: Similar to my comment below, you could just get rid of `const playlist` and pass the raw values into the method under test.

src/playlist.js
```diff
@@ -487,7 +510,9 @@ Playlist.duration = duration;
 Playlist.seekable = seekable;
 Playlist.getMediaInfoForTime = getMediaInfoForTime;
 Playlist.isEnabled = isEnabled;
+Playlist.isDisabled = isDisabled;
```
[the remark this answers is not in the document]
mjneil: Ya we'll just merge that one first and i can rebase this

PR #1286, opened 2017-10-26
Remove extraneous methods from PlaylistLoader

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/playlist-loader.js
+++ b/src/playlist-loader.js
@@ -38,6 +38,18 @@ export const updateSegments = (original, update, offset) => {
   return result;
 };
 
+export const resolveSegmentUris = (segment, baseUri) => {
+  if (!segment.resolvedUri) {
+    segment.resolvedUri = resolveUrl(baseUri, segment.uri);
+  }
+  if (segment.key && !segment.key.resolvedUri) {
+    segment.key.resolvedUri = resolveUrl(baseUri, segment.key.uri);
+  }
+  if (segment.map && !segment.map.resolvedUri) {
+    segment.map.resolvedUri = resolveUrl(baseUri, segment.map.uri);
+  }
+};
+
 /**
   * Returns a new master playlist that is the result of merging an
   * updated media playlist into the original version. If the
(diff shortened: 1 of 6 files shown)


src/playlist-loader.js
```diff
@@ -165,6 +169,28 @@ export const refreshDelay = (media, update) => {
   return delay;
 };
 
+/*
+ * Returns whether the current playlist is the lowest rendition
+ *
+ * @return {Boolean} true if on lowest rendition
+ */
+export const isLowestEnabledRendition = (master, media) => {
```
mjneil: I feel like this would make more sense being in the playlist.js file

gesinger: Good call. Looks much nicer there 👍

test/playlist-loader.test.js
```diff
@@ -791,6 +794,43 @@ QUnit.test('uses last segment duration for refresh delay', function(assert) {
     'used half targetDuration when update is false');
 });
 
+QUnit.test('isLowestEnabledRendition detects if we are on lowest rendition',
```
mjneil: I think putting these tests in `playlist.test.js` would be more appropriate and then you can test on just plain objects with the info thats needed instead of all this extra creating a loader, responding to request, etc.

mjneil: master-playlist-controller.test.js also already has integration tests for isLowestEnabledRendition, so even more reason to convert these to unit tests

test/playlist.test.js
```diff
@@ -4,6 +4,64 @@ import QUnit from 'qunit';
 import xhrFactory from '../src/xhr';
 import { useFakeEnvironment } from './test-helpers';
 
+QUnit.module('Playlist Status');
```
mjneil: There is already an entire module of tests for blacklisting and enabled states near the bottom of this file...
https://github.com/videojs/videojs-contrib-hls/blob/master/test/playlist.test.js#L791

PR #1292, opened 2017-11-07
Fix: SyncController.selectSyncPoint_ and segment-loader properly handling gaps in live mode

tchakabam:
## Description
Please describe the change as necessary.
If it's a feature or enhancement please be as detailed as possible.
If it's a bug fix, please link the issue that it fixes or describe the bug in as much detail.

## Specific Changes proposed

We didn't handle unadvertised gaps in playlists in live cases yet where we were sync'ing in playlist-mode.

This adds handling the case where the initial guess was wrong and we have started loading to far ahead and not on the target time.

Also, we generally should append with the actual timestamps offset, as otherwise we are not able for handling timestamp gaps at all properly.

Also, there seemed to be a bug in the SyncController, where we were iterating from 1 instead of 0 on the sync methods.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -648,9 +648,11 @@ export default class SegmentLoader extends videojs.EventTarget {
    */
   checkBuffer_(buffered, playlist, mediaIndex, hasPlayed, currentTime, syncPoint) {
     let lastBufferedEnd = 0;
+    let lastBufferedStart = 0;
     let startOfSegment;
 
     if (buffered.length) {
+      lastBufferedStart = buffered.start(buffered.length - 1);
       lastBufferedEnd = buffered.end(buffered.length - 1);
     }
 
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -647,8 +647,8 @@ export default class SegmentLoader extends videojs.EventTarget {
    * @returns {Object} a segment request object that describes the segment to load
    */
   checkBuffer_(buffered, playlist, mediaIndex, hasPlayed, currentTime, syncPoint) {
-    let lastBufferedEnd = 0;
     let lastBufferedStart = 0;
+    let lastBufferedEnd = 0;
     let startOfSegment;
 
     if (buffered.length) {
--- a/src/segment-loader.js
+++ b/src/segment-loader.js
@@ -687,9 +687,11 @@ export default class SegmentLoader extends videojs.EventTarget {
 
       syncPointError = lastBufferedStart - currentTime;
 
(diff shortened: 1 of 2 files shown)


src/segment-loader.js
```diff
@@ -680,6 +682,22 @@ export default class SegmentLoader extends videojs.EventTarget {
       'fetchAtBuffer:', this.fetchAtBuffer_,
       'bufferedTime:', bufferedTime);
 
+    let syncPointError = 0;
+    if (currentTime < lastBufferedStart) {
```
gesinger: I think that handling the gap resolution logic may be better in playback watcher, probably within `fixedBadSeeks_`: https://github.com/videojs/videojs-contrib-hls/blob/2158b756eef8a2441fa580f673e316391d63ada6/src/playback-watcher.js#L157

This would also allow us to re-seek.

src/segment-loader.js
```diff
@@ -1134,15 +1152,22 @@ export default class SegmentLoader extends videojs.EventTarget {
       return;
     }
 
-    if (segmentInfo.timestampOffset !== null &&
-        segmentInfo.timestampOffset !== this.sourceUpdater_.timestampOffset()) {
-      this.sourceUpdater_.timestampOffset(segmentInfo.timestampOffset);
+    const timelineMapping = this.syncController_.mappingForTimeline(segmentInfo.timeline);
+    const absoluteTimestampOffset = segmentInfo.timestampOffset - timelineMapping;
```
gesinger: Since it's all relative to the first timestamp offset we append, wouldn't the numbers come out equitable whether we subtract the starting timestamp offset versus use the actual timestamp offsets provided? I saw you mentioned that it helps handle gaps, but would you be able to provide a case where it happens?

src/segment-loader.js
```diff
@@ -1134,15 +1152,22 @@ export default class SegmentLoader extends videojs.EventTarget {
...
+    // resolve to absolute offset only for live playlists
+    const resolvedTimestampOffset = this.playlist_.endList ? segmentInfo.timestampOffset : absoluteTimestampOffset;
+
+    if (segmentInfo.timestampOffset !== null 
+      && resolvedTimestampOffset !== this.sourceUpdater_.timestampOffset()) {
+
+
+      console.log('setting timestamp offset (resolved to timeline):', absoluteTimestampOffset);
```
gesinger: Extra logging statement.

gesinger (commented):
Thanks for another PR @tchakabam !

I left some comments within. When you get a chance, let me know your thoughts.

forbesjo:

PR #1299, opened 2017-11-23
Support EXT-X-START with an updated m3u8 parser

masterful:
## Description
This implements support for the [EXT-X-START](https://tools.ietf.org/html/draft-pantos-http-live-streaming-18#section-4.3.5.2) tag. This tag indicates a preferred point at which to start playing a Playlist.

I should note that I didn't explicitly add support for the precise attribute, the seek should be as precise as possible with the current implementation, since we don't seek based on segment start time but rather to a time within a particular segment.

## Specific Changes proposed
- update m3u8-parser
    (not sure why the package-lock file exploded, sorry - perhaps it's my version of npm?)
- change setupFirstPlay to support `EXT-X-START` tag

## Requirements Checklist
  - [ ] Docs/guides updated
        (not sure how to approach this one, since it seems outside the scope of the VideoJS docs to explain HLS implementation?)
        (again - not sure this is in scope?)


src/master-playlist-controller.js
```diff
@@ -711,24 +711,38 @@ export class MasterPlaylistController extends videojs.EventTarget {
       return false;
     }
 
-    // when the video is a live stream
-    if (!media.endList) {
+    // when the video is a live stream, or has a specific start time:
```
gesinger: We may want to update the comment above `setupFirstPlay` as well

src/master-playlist-controller.js
```diff
@@ -738,7 +752,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       // trigger firstplay to inform the source handler to ignore the next seek event
       this.trigger('firstplay');
       // seek to the live point
```
gesinger: We may want to update this comment

src/master-playlist-controller.js
```diff
@@ -711,24 +711,38 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
+      if (media.start) {
+        const offset = media.start.timeOffset;
+
+        if (offset < 0) {
+          startTime += offset;
+        } else {
+          startTime = offset;
+        }
+        // cap the startTime to within the seekable range:
+        startTime = Math.max(seekable.start(0), Math.min(seekableTime, startTime));
```
gesinger: By https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.5.2 , I think we need to account for `PRECISE`, or at least mention it in a comment (it's probably OK to ignore using it and just add in a comment for the time being, since this PR just makes things better)

src/master-playlist-controller.js
```diff
@@ -711,24 +711,38 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
       }
 
+      const seekableTime = seekable.end(0);
+      let startTime = seekableTime;
+
+      if (media.start) {
+        const offset = media.start.timeOffset;
+
+        if (offset < 0) {
+          startTime += offset;
```
gesinger: I think `seekable.end(0)` should be the "safe seekable end" meaning 3 segments from the end of the playlist, whereas the spec says the offset should be from the end of the playlist.

Would be good to add a test for that.

src/master-playlist-controller.js
```diff
@@ -711,24 +711,38 @@ export class MasterPlaylistController extends videojs.EventTarget {
...
-        // point
+        // without a seekable range, the player cannot seek to a specific time
         return false;
       }
 
+      const seekableTime = seekable.end(0);
+      let startTime = seekableTime;
+
+      if (media.start) {
+        const offset = media.start.timeOffset;
```
gesinger: It might be clearer (and easier to test independently) to wrap all of this inside of another function. e.g., `startTime = mediaStartTime(media.start.timeOffset, media.start.precise, seekable.end(0))`

utils/manifest/startLiveStream.m3u8
```diff
@@ -0,0 +1,15 @@
+#EXTM3U
+#EXT-X-START:TIME-OFFSET=0
```
gesinger: Might be good to use a positive number since 0 is often an edge case.

gesinger (commented):
Thank you for the PR @masterful ! Looks good, just a few comments.

The package-lock.json is always a fun one. To just have the single change try doing: `npm install --package-lock-only m3u8-parser@4.0.0`

menkveldj:
love this! any chance it will be merged into master soon?

masterful:
gah - I'd completely forgotten about this - is it worth revisiting?

menkveldj:
Yes, i spent hours trying to make my own branch and build this out. It would be a huge help for certain types of projects.

utils/manifest/startLiveStream.m3u8
masterful: I don't have the data to say otherwise, but I'm curious to know how you determined it was an edge case, since my primary use case is to use `0`...

masterful:
I'm not sure I understand where the tests are failing :disappointed:

forbesjo:

Issue #600, opened 2016-03-15
Support for webpack?


hanfeisun:
In the version of 1.x videojs-contrib-hls, I can shim the package like this:

```
import vjs  from 'video.js';
window.videojs = vjs;

require('imports?this=>window!videojs-contrib-hls');

```

However, in 2.0.1 videojs-contrib-hls, this method doesn't work. 

A dependency called `webworkify` will throw an error

```
Uncaught TypeError: Cannot convert undefined or null to object
module.exports  @   index.js:10
VirtualSourceBuffer @   virtual-source-buffer.js:66
addSourceBuffer @   html-media-source.js:168
setupSourceBuffer_  @   videojs-contrib-hls.js?./~/imports-loader?this=>window:621
handleSourceOpen    @   videojs-contrib-hls.js?./~/imports-loader?this=>window:532
data.dispatcher @   video.js:18626
trigger @   video.js:18730
EventTarget.trigger @   video.js:7950

```

Will videojs-contrib-hls plan to support webpack in the future?

dmlap:
I think this was the issue behind https://github.com/videojs/mux.js/pull/53. That was fixed in mux.js >= 2.0.1, which you should use if you build this project yourself (and in our next release when that happens). Do you see the same error if you build contrib-hls yourself?

samward1985:
+1

samward1985:
Eh - no luck... @hanfeisun did you find a solution?

hanfeisun:
@samward1985  No, it doesn't work under Webpack now.. So I am still using `videojs-contrib-hls 1.3.11`..

andrewmnlv:
Hello all!)
It works for me:
   resolve: {
      alias: {
         'videojs-contrib-hls': __dirname + '/node_modules/videojs-contrib-hls/dist/videojs-contrib-hls
      }
   }

videojs-contrib-hls v2.2.0

gkatsev:
I think contrib-hls 2.2.0 won't require any aliases anymore. Though, unfortunately, that version also requires videojs 5.10.1 and up.

andrewmnlv:
@gkatsev i have videojs v5.10.2, but webworkify breaks webpack build because of
[no fallbacks for arguments](https://github.com/substack/webworkify/blob/master/index.js#L3)

gkatsev:
If webworkify is the only problem, we should consider swapping it out for another module with similar functionality that works in both webpack and browserify.

[...]

PR #1307, opened 2017-12-14
add entry of require() for webpack

hxl-dy:
## Description
fix https://github.com/videojs/videojs-contrib-hls/issues/600

## Specific Changes proposed
add entry of require() for webpack

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "5.12.2",
   "description": "Play back HLS with video.js, even where it's not natively supported",
   "main": "es5/videojs-contrib-hls.js",
+  "module": "dist/videojs-contrib-hls.js",
   "engines": {
     "node": ">= 0.10.12"
   },


gkatsev:
I'm not sure this will actually fix it. Not sure about webpack, but browserify doesn't like importing browserified code unless it ran through derequire, so, that could be an issue.
Though, I think the main issue is that webpack users would still need to configure this because our UMD build expects videojs to be available globally: https://github.com/videojs/videojs-contrib-hls/blob/master/package.json#L63-L67

thecotne:
@gkatsev 
> our UMD build expects videojs to be available globally

why is that?

wouldn't UMD package use require/define function to import dependencies ?

and use global variables if require and define are not available 

and list dependencies as peerDependencies 

like this https://github.com/umdjs/umd/blob/master/templates/jqueryPlugin.js

forbesjo:
#600 is fixed, closing

PR #1327, opened 2018-01-26
update url-toolkit to ^2.1.3

tjenkinson:
See https://github.com/videojs/videojs-contrib-hls/pull/1029

I added the `^` given you are using a lockfile. If you want it to be static I can remove that.


forbesjo:
Looks like there are a lot of package-lock updates. When I removed and re-added your module to update/unlock I only noticed `url-toolkit` update (and `karma-detect-browsers` was removed). Could you update the module that way and overwrite this commit?

forbesjo:
Deleted my comment because I ran my commands against videojs-http-streaming and not this project. Currently testing this PR against a couple of sources and will merge shortly.

PR #1328, opened 2018-01-26
simplify resolve-url

tjenkinson:
It is possible to remove some of the checks here, because they are already performed in the library.

If the relative url is actually absolute, it returns it [here](https://github.com/tjenkinson/url-toolkit/blob/71441a287aabe535ffac70a380f546cda6719459/src/url-toolkit.js#L46) after running through one regex (which is cached), so it should also improve performance slightly


gesinger:
Thanks for the PR @tjenkinson !

re: the broken test case. Will the protocol that URLToolkit returns be the same as the page?

forbesjo:

Issue #963, opened 2017-01-19
Uncaught exception at 'addSourceBuffer' when changing source or disposing/re-creating video


carlosllera:
## Description
In our scenario, we need to dynamically create a video, change its source several times within few seconds (user will be previewing videos), and dispose and re-create it at some point.

**Using our video source files**, when in the middle of a video, we change the video source through `videojs().src` function, an _Uncaught DOMException: Failed to execute 'addSourceBuffer' on 'MediaSource': This MediaSource has reached the limit of SourceBuffer objects it can handle._ is received.

Here is a reduced test case on jsbin:
http://jsbin.com/tojewagula/edit?html,console,output

Here are the source files you can download and check (npm install and npm start should be enough):
https://drive.google.com/file/d/0B-Ny3L4qtHdUXzRaZlltWjF0YXM/view?usp=sharing

Here are some example M3U8 URLs we are using:
https://scdn.golden-race.net/hlsvideos/dog6/d6-5-4-3-1437.m3u8?st=JMUY2EVWtAHdK-GDFcNyMw&e=1485420418
https://scdn.golden-race.net/hlsvideos/sp/s4-4-3-054.m3u8?st=BvMPTBAMEamHQICmFvEoJg&e=1485424078
https://scdn.golden-race.net/hlsvideos/mt/mt-5-4-1-0229.m3u8?st=yYu5J6ml4wkTdgSrbAkY9Q&e=1485424498

## Sources
We realized that inserting a debug point in the function `createRealSourceBuffers_()` provoked that this exception is never thrown, so we just inserted a simple _wait_ of  just before the buffer is created with `_this2.mediaSource_.nativeMediaSource_.addSourceBuffer(...)`, and it's working almost all the times (we still get the exception sometimes on extreme conditions).

## Steps to reproduce
Using the given jsbin example:

1. Choose one of our URLs, paste it in the text input, and press Load. In the browser console you can check that the video is created with the given url and it starts playing.
2. After waiting a couple of seconds, press Load again with the same video or choose a different one (may have to do that several times until the error is found).
3. Encounter the error.

Note: Can be also discovered even though video is disposed.

### Expected
Video should be working correctly, even though is Loaded several times and/or disposed.

### Error output
An unexpected error is found and **video looses sound**.

## Additional Information
[...]

PR #1330, opened 2018-01-29
Fix issue 963: error on SourceBuffer creation

tchakabam:
## Description

Fixes https://github.com/videojs/videojs-contrib-hls/issues/963, see history/triage there.

Defers creation of SourceBuffer and retries as necessary (see comments in code).

Trying to deal with apparent uncertainty coming from lower layers (MediaSource wrappers or actually browsers implementations).

TODO:

- [ ] Adapt tests to check for creation of SourceBuffer asynchroneously

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -130,7 +130,7 @@ export default class SourceUpdater {
     } else if (mediaSource.readyState === 'closed') {
       mediaSource.addEventListener('sourceopen', createSourceBufferDeferred);
     } else {
-      throw new Error('MediaSource in illegal ready-state: ' + mediaSource.readyState)
+      throw new Error('MediaSource in illegal ready-state: ' + mediaSource.readyState);
     }
   }
 
--- a/src/source-updater.js
+++ b/src/source-updater.js
@@ -84,12 +84,41 @@ export default class SourceUpdater {
     createSourceBufferDeferred =
       (() => setTimeout(createSourceBuffer, ADD_SOURCE_BUFFER_RETRY_DEFER_MS));
 
+    /**
+     * @private
+     * @member {Array<function>}
+     */
     this.callbacks_ = [];
+
+    /**
+     * @private
+     * @member {function}
+     */
     this.pendingCallback_ = null;
+
+    /**
+     * @private
+     * @member {number}
+     */
     this.timestampOffset_ = 0;
+
+    /**
+     * @public
+     * @member {MediaSource}
+     */
     this.mediaSource = mediaSource;
+
+    /**
+     * @public
+     * @member {boolean}
+     */
     this.processedAppend_ = false;
 
+    /**
+     * @private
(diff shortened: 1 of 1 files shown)


src/source-updater.js
```diff
@@ -41,16 +71,66 @@ export default class SourceUpdater {
...
+    // that sometimes when the `sourceopen` event is triggered, or when
+    // the readyState indicates open, the
+    // MediaSource is actually not ready yet to accept
+    // calls to its `addSourceBuffer` method.
+    // This seems to happen especially for when we need to create SourceBuffers
+    // for seperate audio streams (A/V not muxed into the "main" stream).
+    // However when state is settled on the next tick,
+    // it seems safe to do so.
+    createSourceBufferDeferred =
+      (() => setTimeout(createSourceBuffer, ADD_SOURCE_BUFFER_RETRY_DEFER_MS));
```
tchakabam: running this immediately (zero timeout) doesn't fix the issue. 

we have experienced that this type of "delay" allows for a reliable behavior.

that proves this is a pretty "fuzzy" issue in the lower layers imho.

tchakabam:
Just another thought: This seems to happen actually only when we need to create several SourceBuffer objects. This leads me to the doubt wether somehow browsers implementations have introduced an intermediate state just after creation of SourceBuffers where we need to wait before the next call to addSourceBuffer. Maybe we need to check `updating` property, wait for an `updateend` event? Just checked the spec again, it's not mentioning any of that or giving any clues about repeated calls to `addSourceBuffer`, see https://developer.mozilla.org/en-US/docs/Web/API/MediaSource/addSourceBuffer

gesinger:
Thanks for another contribution and all the detailed notes!

I agree that the timeout solution, while simplest, is not the most pleasant. If the browsers aren't (or aren't consistently) managing the state when dealing with adding multiple source buffers in quick succession, it may be necessary for us to consider wrapping the media source itself (and/or making changes via videojs-contrib-media-sources) and making the functions async to ensure that we are waiting for each `addSourceBuffer` to appropriately finish (as you mentioned, listening for `updating` and `updateend`). That would at least be a little more deterministic than timeouts.

There are a few more details in https://www.w3.org/TR/media-source/#methods , but not complete there either.

tchakabam:
@gesinger Yes totally agree about that. I'd also want the fix/solution to be robust against anything, and for that we need to do it as you describe.

Xesme:
Is there a status update or work around for this merge?

TylerZubke:
I'm also wondering on the status of this. There are certain cases where the vast-vpaid plugin loads a preroll and this error gets thrown.

trailofdad:
This is Issue still alive? I'm trying to solve this...

forbesjo:

PR #1338, opened 2018-02-10
Fix tech.play() throwing unresolved promise errors on Chrome

grobolom:
## Description
This is an enhancement that fixes errors being thrown by videojs-contrib-hls on Chrome. While the error does not impact player functionality, it pollutes the console log, which may result in confusion or time wasted during manual/automated testing. The error is caused by the media `.play()` event returning a Promise in Chrome, which throws the error `The play() request was interrupted by a call to pause()`.

The fix is mirrored from [a similar fix in the video.js core repo](https://github.com/videojs/video.js/pull/3518), which simply wraps the return value of `.play()` in a promise if necessary.

## Specific Changes proposed
This fix proposes to wrap the `this.tech_.play()` call in `handleSourceOpen_()` in the `master-playlist-controller` with a `.then()`, so that Chrome does not show a Promise-based error.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -771,7 +771,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
       // Catch/silence error when a pause interrupts a play request
       // on browsers which return a promise
-      if (playPromise !== undefined && typeof playPromise.then === 'function') {
+      if (typeof playPromise !== 'undefined' && typeof playPromise.then === 'function') {
         playPromise.then(null, (e) => {});
       }
     }
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -767,7 +767,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // code in video.js but is required because play() must be invoked
     // *after* the media source has opened.
     if (this.tech_.autoplay()) {
-      this.tech_.play();
+      const playPromise = this.tech_.play();
+
+      // Catch/silence error when a pause interrupts a play request
+      // on browsers which return a promise
+      if (playPromise !== undefined && typeof playPromise.then === 'function') {
+        playPromise.then(null, (e) => {});
+      }
     }
 
     this.trigger('sourceopen');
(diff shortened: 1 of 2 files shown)


src/master-playlist-controller.js
```diff
@@ -767,7 +767,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // code in video.js but is required because play() must be invoked
     // *after* the media source has opened.
     if (this.tech_.autoplay()) {
-      this.tech_.play();
+      const playPromise = this.tech_.play();
+
+      // Catch/silence error when a pause interrupts a play request
+      // on browsers which return a promise
+      if (playPromise !== undefined && typeof playPromise.then === 'function') {
```
gesinger: I think the linter might warn to use `typeof playPromise !== 'undefined'`

grobolom: @gesinger cool! I'll update later today.

src/master-playlist-controller.js
```diff
@@ -767,7 +767,13 @@ export class MasterPlaylistController extends videojs.EventTarget {
     // code in video.js but is required because play() must be invoked
     // *after* the media source has opened.
     if (this.tech_.autoplay()) {
-      this.tech_.play();
+      const playPromise = this.tech_.play();
+
+      // Catch/silence error when a pause interrupts a play request
+      // on browsers which return a promise
+      if (playPromise !== undefined && typeof playPromise.then !== 'undefined') {
```
gesinger: Thanks for the update @grobolom ! Though in this case the typeof comparing to a function is OK (and preferred), but the `playPromise !== undefined` should probably be `typeof playPromise !== 'undefined'`

grobolom: aaaah. I see. My bad, I'll fix that up!

grobolom:
@gesinger all fixed up 👍

grobolom:
@gesinger hey man, just wondering - is there an approximate timeline for when PRs get merged in these repositories? I understand it's not a high-traffic area, but it would be cool for to try this out in our code (and avoid some errors). No rush though, just asking :)

gesinger:
Hey @grobolom , sorry about the delay. No set timeline, we were just focused on some other issues. That said, this is all set to go except we do a quick round of manual testing before merging it in. We'll try to get around to it today if we can.

grobolom:
@gesinger no sweat - thanks for adding this in! We really appreciate it :)

PR #1341, opened 2018-02-12
Provide previous segment index on exact time match to be conservative

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated


mjneil (approved):
This change is probably better to have than not have, but thorough testing should be done to make sure this isn't a common case that isn't actually problematic, otherwise this may tack on an extra unnecessary segment request frequently.

forbesjo:

PR #1363, opened 2018-03-06
use webwackify for webworkers to support webpack bundle

mjneil:
Regarding the require.resolve:

webpack resolves module names to ids during compile time. One of the issues with webworkify-webpack-dropin was that it used regex to search for the require statements to get the resolved module id so it could properly setup the bootstrapfn to start at the right module. This broke in some cases with minification because the regex would get messed up. By passing the module id directly from require.resolve, you dont have to do any of the regex work, and can just build the entire module list and have the starting id given to you. This problem is described in more detail here https://github.com/videojs/videojs-contrib-hls/issues/600#issuecomment-293550845

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2796,15 +2796,6 @@
         }
       }
     },
-    "babel-runtime": {
-      "version": "6.26.0",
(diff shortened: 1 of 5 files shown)


forbesjo:
This change looks good to me save for the big package-lock changes

PR #1364, opened 2018-03-07
Add greenkeeper ignore section to package.lock

gesinger:
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,6 @@
   "greenkeeper": {
     "ignore": [
       "browserify",
-      "qunitjs",
       "sinon",
       "webpack",
       "uglify-js"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -140,5 +140,14 @@
     "videojs-swf": "^5.2.0",
     "watchify": "^3.6.0",
     "webpack": "^1.13.2"
+  },
+  "greenkeeper": {
+    "ignore": [
+      "browserify",
+      "qunitjs",
+      "sinon",
```
forbesjo: Should we still let GK keep updating the qunit and sinon dependencies?

gesinger: Sinon will break, not sure offhand about qunit.

gkatsev: updating qunit is probably fine.

gkatsev: Video.js couldn't because it had to support IE8 and qunit 2 dropped support for it but contrib-hls never supported IE8.

PR #1374, opened 2018-03-15
Improved seek behaviour

squarebracket:
## Description

This change requires videojs/video.js#5024, which takes care of the UI and API aspects of buffering.

SourceHandlers that use MSE have a problem: if they push a segment into a `SourceBuffer` and then seek close to the end, playback will stall and/or there will be a massive downswitch in quality. The general approach to fixing this that was discussed on slack was by setting the playback rate of the player to zero, buffering all that was required, and then restoring the previous playback rate. In my implementation, I've done this in the source handler.

Additionally, there is a new `seekDeadline` option for the source handler which controls how long a seek should take to resume. When seeking, the master playlist controller calculates the maximum bandwidth playlist for all requests needed to complete a full seek. The calculations are done using `minRebufferMaxBandwidthSelector` (renamed to `maxBandwidthForDeadlineSelector`).

Besides being a generally good change, this was somewhat needed for backwards compatibility with older video.js versions. Because of [this](https://github.com/videojs/video.js/blob/c98912fd92090e81feba3236f0f8d5b4e8c58b03/src/js/player.js#L1295), if you cause playback to stall after seek by seeking too close to the end of a segment, the second segment will have its spinner disappear (since you'll get a `timeupdate` after the `waiting`). This error is currently masked because seek quality is trashed when seeking, so you'll always be able to buffer the second segment before the spinner disappears from the `timeupdate`.

With the seek deadline enforcement, although not a _real_ fix, it is significantly mitigated since you'll be kicked down to a lower rate for seeks which require downloading two segments. So you probably won't end up with any noticeable delay between spinner disappearing and playback resuming. For that reason, I think the "buffer appropriate number of segments" and the "seek deadline enforcement" should be considered a single PR.

When combined with #1239, this makes the seeking behavior vastly improved:
- playback is on highest rate
- seek is requested by client to right before the end of a segment
- playback pauses, spinner appears, `player.seeking()` is now true
- buffer is sufficiently filled at new time in reasonable timeframe, downswitching if necessary
- two segments at lower rate are now in buffer
- playback resumes, spinner disappears, player fires playback events, `player.seeking()` is now false
- next segment download is at highest rate

videojs/videojs-contrib-hls#1239 has been reviewed, but I don't believe it's been tested. Let me know if you want me to stage the changes here so it can all be considered as part of a single PR. 

## Specific Changes proposed
- Added `seekDeadline` option for controlling how long a seek should take to resume
- `seeking()` is deferred from the tech to the SourceHandler
- `setCurrentTime` stores the current playback rate and set the tech's playback rate to zero
- Inside `SegmentLoader#handleUpdateEnd_`, if we have enough buffer that we can make another request in that time, fire `buffered` on `this`
- The `MasterPlaylistController` receives the `buffered` event and restores the previous playback rate
- `minRebufferMaxBandwidthSelector` has been renamed to `maxBandwidthForDeadlineSelector`, and made more generic
  - it now accepts an `extraRequests` parameter which is added to the `numRequests` for the segment
  - instead of returning `rebufferingImpact`, it now just returns `requestTimeEstimate`
- For the deadline passed to `maxBandwidthForDeadlineSelector`, `abortRequestEarly_` uses either the current calculation (when not seeking) or the time elapsed since seeking started
- Math inside `abortRequestEarly_` is now a bit simpler because it's now using the `requestTimeEstimate` instead of `rebufferingImpact`
- The quality used when seeking is determined by how long it will take to get the number of segments needed to fill the buffer after the seek 
- The MPC's `fastQualityChange_` now accepts a `nextPlaylist` option which will be used instead of `this.selectPlaylist()` when given
- The playback watcher has been modified to not bork when the playback rate is zero and nothing is happening

## Requirements Checklist
  - [ ] Docs/guides updated

## Notes
This is implemented assuming that the video segment loader will always finish after the other segment loaders. I also made [an implementation that guarantees all segment loaders have enough data](https://github.com/videojs/videojs-contrib-hls/compare/master...squarebracket:fix/full-buffer-seek-fix), if you want to compare what that would be like. I found this to be much more concise which is why it gets the PR.

Figuring out some of the testing stuff was becoming troublesome. It'll probably take me a while to get through adding all the tests.

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1026,7 +1026,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
   }
 
   seeking() {
-    return this.seeking_;
+    return this.tech_.el_.seeking || this.seeking_;
   }
 
   onSyncInfoUpdate_() {
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -643,6 +643,10 @@ export class MasterPlaylistController extends videojs.EventTarget {
     this.audioSegmentLoader_.on('ended', () => {
       this.onEndOfStream();
     });
+
+    this.mainSegmentLoader_.on('buffered', (event) => {
+      this.tech_.setPlaybackRate(this.playbackRate);
+    });
   }
 
   mediaSecondsLoaded_() {
@@ -953,6 +957,9 @@ export class MasterPlaylistController extends videojs.EventTarget {
     if (buffered && buffered.length && this.mode_ !== 'flash') {
       return currentTime;
     }
+
+    this.playbackRate = this.tech_.playbackRate();
+    this.tech_.setPlaybackRate(0);
 
     // cancel outstanding requests so we begin buffering at the new
     // location
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -1062,7 +1062,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
       deadline: this.seekDeadline_,
       currentTimeline: this.mainSegmentLoader_.currentTimeline_,
       syncController: this.syncController_,
-      extraRequests: 1
+      extraRequests
     });
 
     // cancel outstanding requests so we begin buffering at the new
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -645,10 +645,27 @@ export class MasterPlaylistController extends videojs.EventTarget {
       this.onEndOfStream();
     });
 
-    this.mainSegmentLoader_.on('buffered', (event) => {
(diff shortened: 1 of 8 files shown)


squarebracket:
I think I've managed to nail down all the fiddly bits here.

I had some trouble with the tests, [like `mediachange` not firing for some reason](https://github.com/videojs/videojs-contrib-hls/pull/1374/files#diff-af66169118bc27c2ada066de16503fd7R787), so it'll probably take me a bit to get through making the tests, but the code itself should be feature-complete.

forbesjo:

squarebracket:
For anyone who lands here, I opened PR videojs/http-streaming#370.

PR #1379, opened 2018-03-20
WIP: proper syncing for non-muxed HLS live feeds

squarebracket:
## Description
Corrects a/v desync for non-muxed live feeds by properly syncing the `timestampOffset` of segment loaders when corresponding segments have different starting DTSs. Note that, since VODs don't call `timestampOffset`, this only affects live feeds.

I get sporadic sync problems so I'll still need to look into that.

## Specific Changes proposed
- Use the `timingInfo` of the first finished segment to dispatch a `firststarttime` event to the master playlist controller
- Set the `dtsSync_` of the MPC if it's null (from the event data), then dispatch a `dtsSync` to the calling segment loader with the received `start` and `this.dtsSync_`.
- The `dtsSync` event handler on the segment loader sets the timestamp offset from the reference

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -253,6 +253,7 @@ export class MasterPlaylistController extends videojs.EventTarget {
 
     Hls = externHls;
 
+    this.dtsSync_ = null;
     this.withCredentials = withCredentials;
     this.tech_ = tech;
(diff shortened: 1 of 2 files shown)


forbesjo:

PR #1412, opened 2018-05-03
update some testing configs

squarebracket:
- bump karma browser runners versions
- make detect-browsers' `postDetection` fancier
- don't remove safari from available browsers
- align some karma settings with VHS
- mute playback tests (which means chrome won't block autoplay)
- dockerize it
- use latest firefox
- bump detect-browsers so we can use `preferHeadless: true` option instead of manually rewriting that we want a headless browser
- be fancier with Chrome `customLaunchers`


test/karma.conf.js
```diff
@@ -3,24 +3,30 @@ var isparta = require('isparta');
...
+    preferHeadless: true,
+    // replace chrome headless with one that is suitable for automatic testing
     postDetection: function(availableBrowsers) {
-      var safariIndex = availableBrowsers.indexOf('Safari');
-
-      if(safariIndex !== -1) {
-        availableBrowsers.splice(safariIndex, 1);
-        console.log("Disabled Safari as it was/is not supported");
+      for (let index in availableBrowsers) {
+        var browser = availableBrowsers[index];
```
gesinger: Minor, but linter usually complains if there isn't a space after a variable declaration.

test/karma.conf.js
```diff
@@ -3,24 +3,30 @@ var isparta = require('isparta');
...
-    enabled: false,
     usePhantomJS: false,
+    // use headless mode automatically for browsers that support it
+    preferHeadless: true,
+    // replace chrome headless with one that is suitable for automatic testing
     postDetection: function(availableBrowsers) {
-      var safariIndex = availableBrowsers.indexOf('Safari');
-
-      if(safariIndex !== -1) {
-        availableBrowsers.splice(safariIndex, 1);
```
gesinger: Safari passes?

squarebracket:
Made appropriate changes, plus a couple little other ones. For some reason, on my test mac, but not on other platforms, chrome won't let autoplay happen with

```javascript
this.player = videojs(video, {muted: true});
```

but does let it happen with

```javascript
this.player = videojs(video);
this.player.muted(true);
```

which confuses me greatly, but hey, that it makes it work, so..... :man_shrugging:

test/karma.conf.js
squarebracket: I guess we should probably `QUnit.skip` when running on Safari or something...

squarebracket:
Err, the package-lock wasn't updated, so that will probably have to be done.

Issue #1395, opened 2018-03-31
AES encryption keys are requested from the server for every segment


benwilber:
# Description
I expect the AES encryption key to be fetched from the server (at most) once, but instead it is fetched for every single segment in the playlist.

Given the following playlist:

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-TARGETDURATION:7
#EXT-X-KEY:METHOD=AES-128,URI="/keys/foostream/1522536482595.key",IV=0x0000000000000000000001627E3F0723
#EXTINF:6.848,
1522536482595.ts
#EXTINF:6.000,
1522536489415.ts
#EXTINF:6.600,
1522536495457.ts
#EXTINF:6.167,
1522536502014.ts
#EXTINF:6.000,
1522536508181.ts
#EXTINF:6.000,
1522536514179.ts
#EXTINF:6.000,
1522536520184.ts
#EXTINF:6.000,
1522536526177.ts
#EXTINF:6.000,
1522536532179.ts
#EXTINF:6.500,
1522536538221.ts
```
The HLS specification indicates that the same key should be used for every segment following an `EXT-X-KEY` tag, until the next `EXT-X-KEY`.  So then it is not needed to re-fetch the same key over and over again for every segment.  Especially since acquiring the AES encryption key might be an expensive operation that involves verifying viewer access with database lookups, etc.

EDIT:

I see there is a patch for this here: https://github.com/videojs/videojs-contrib-hls/issues/776

I think deferring to HTTP caching is incorrect in this case and against the HLS specification:

```
Media segments MAY be encrypted.  The EXT-X-KEY tag specifies how to
   decrypt them.  It applies to every media segment that appears between
   it and the next EXT-X-KEY tag in the Playlist file with the same
   KEYFORMAT attribute
```

The URL in the `URI` value of an `EXT-X-KEY` tag may or may not be cache-able upstream -- which is the whole point of defining the interval of segments encrypted per key.  If the player just fetches the key for every single segment then it defeats the whole point of encrypting multiple segments with the same key.

EDIT EDIT:

I can't find a single other HLS media player that re-fetches the encryption keys for every segment.  I've tested Clappr player, Flow player, JW Player, ffplay, QuickTime, and VLC.  It appears this is behavior unique to this Videojs HLS plugin.

benwilber:
[...]

PR #1433, opened 2018-06-28
Cache encryption keys

benwilber:
## Description

See: https://github.com/videojs/videojs-contrib-hls/issues/1395

AES encryption keys are fetched for every segment even if the same key is used to encrypt multiple segments (per HLS specification.)  This causes unnecessary load on the key server if access to encryption keys are guarded by user-specific authentication and authorization mechanisms.

## Specific Changes proposed

Cache encryption keys based on their resolved URI and only fetch keys when they're needed to decrypt the next series of segments.

## Requirements Checklist
- [x ] Feature implemented / Bug fixed
  - [ ] Docs/guides updated

This is a rather crude PR because it doesn't update the tests (they fail now) and I'm not even sure this is the correct way to implement this.  I'm not very experienced with this tech stack (I can't figure out how the testing libraries are supposed to work) so this is more "request for guidance" to implement this bug fix.

All feedback or assistance is welcome!

### DIFF
--- a/src/media-segment-request.js
+++ b/src/media-segment-request.js
@@ -126,6 +126,8 @@ const handleErrors = (error, request) => {
   return null;
 };
 
+const cachedEncryptionKeys_ = {};
+
 /**
  * Handle responses for key data and convert the key data to the correct format
  * for the decryption step later
@@ -160,6 +162,7 @@ const handleKeyResponse = (segment, finishProcessingFn) => (error, request) => {
     view.getUint32(8),
     view.getUint32(12)
   ]);
+  cachedEncryptionKeys_[segment.key.resolvedUri] = segment.key.bytes.slice(0);
   return finishProcessingFn(null, segment);
 };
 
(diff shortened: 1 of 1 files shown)


aaronchenweb:
I got same problem, please help

gesinger (commented):
Thanks for the PR, and sorry for the delay.

To maintain backwards compatibility, what do you think of adding an option that defaults to the original behavior, but if set to true enables caching. Maybe something like `cacheEncryptionKeys`? And maybe a test or two around it.

Thanks again.

forbesjo:
@benwilber are you still interested in this feature?

benwilber:
@forbesjo Yes I am still interested in implementing this fix.  Unfortunately I'm not very familiar with this tech stack and have been having trouble updating the tests so that this case is properly covered.  If anybody could provide any assistance, or point me in the right direction I would appreciate it.

Thanks!

forbesjo:

PR #1436, opened 2018-07-02
ignore: move bot

forbesjo:
## Description
Added a bot for moving issues


gkatsev:
If you're using the default, it could be empty too, fwiw

forbesjo:
I'd like to err on explicit options so a user can see everything at a glance

gkatsev:
Sounds good!

PR #1437, opened 2018-07-13
fix travis config to _actually_ use containerzed infra

squarebracket:
## Description
When I was testing how to get the tests running in the dockerized travis infra, I was running in the travis docker image as root. When moving to the actual travis config, it told me I needed to be root, so without thinking I added `sudo` to the front. I was sure they'd allow package installs, which they do, but only [using the apt addon](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Container-Based-Infrastructure); having `sudo` in the script [would automatically kick you over to the VM-based infra](https://docs.travis-ci.com/user/reference/overview/#For-a-particular-.travis.yml-configuration).

[Until they broke that yesterday.](https://github.com/travis-ci/travis-ci/issues/9875)

Anyway, now we'll actually be using the containerized infra.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,6 @@ before_script:
   - greenkeeper-lockfile-update
   # always install the appropriate version of video.js for the environment
   - npm i "video.js@$VJS";
-  # pulseaudio is needed for dockerized chrome since it has a dummy sink and
-  # the dummy audio ALSA kernel module isn't loaded by Travis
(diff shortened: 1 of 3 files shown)


squarebracket:
Travis is switching to all non-containerized infra, so this PR is no longer relevant.

PR #1099, opened 2017-04-26
Fixed codecs to mimetype conversion to take into account all possible scenarios

imbcmdth:
## Description
There are 4 possible predicates that influence how to generate mimetypes from codec strings:

* Video codec specified
* Audio codec specified
* Are there multiple audio tracks (audio media-groups)
* If so, is the primary media demuxed (lacking audio) 

This PR adds complete unit testing of all 24 possible combinations and fixes issues around the function that caused audio-only streams with multiple audio tracks present.

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [x] If necessary, more likely in a feature request than a bug fix
  - [x] Unit Tests updated or fixed
- [ ] Reviewed by Two Core Contributors

PR #1439, opened 2018-07-28
Use default audio codec if none specified by playlist

arski:
@imbcmdth @gesinger @mjneil In this PR https://github.com/videojs/videojs-contrib-hls/pull/1099/files you put falling back to default audio codec only if multiple audio streams were found https://github.com/videojs/videojs-contrib-hls/pull/1099/files#diff-3f8fe2d6cf175982cf8257063da92a81R206 - up until then the default codec was always used if none was specified by the manifest.

For us this specifically breaks audio playback when we get manifests like https://user-images.githubusercontent.com/904818/37703814-cba60ae2-2cee-11e8-9167-e35a8a3b5734.png that do not explicitly specify audio codecs (but do have audio).

I'll happily update tests or whatever, just wanted to make sure you agree with the fix first.

FYI @lachilles @robinmaben @practual

### DIFF
--- a/src/master-playlist-controller.js
+++ b/src/master-playlist-controller.js
@@ -162,12 +162,9 @@ export const mimeTypesForPlaylist_ = function(master, media) {
     }
   }
 
-  // HLS with multiple-audio tracks must always get an audio codec.
-  // Put another way, there is no way to have a video-only multiple-audio HLS!
-  if (isMaat && !codecInfo.audioProfile) {
-    videojs.log.warn(
-      'Multiple audio tracks present but no audio codec string is specified. ' +
-      'Attempting to use the default audio codec (mp4a.40.2)');
+  // Use default audio profile if none specified by playlist
+  if (!codecInfo.audioProfile) {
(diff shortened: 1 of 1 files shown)


arski:
hello kind people of videojs-contrib-hls :) could you share your thoughts on this please

arski:
@imbcmdth is this library still actively maintained? I see quite a few issues and PRs hanging about :(

gesinger:
While going around the spec can sometimes help problematic streams, in this case, because the spec is pretty clear on `MUST`, the stream could cause unexpected issues on environments where we use native playback.

Also, we may actually have some bugs if we try to work around a mislabeled `CODECS` attribute. For instance, it would set up an audio buffer for streams which should be video only, and thus not allow for video only playback.

Ultimately, we probably shouldn't trust the manifest at all, and delay creation of source buffers until we parse the content, which is something being worked on with "transmux before append." But for the meantime, we can't assume the audio codec based on manifest-only information if it goes against spec.

arski:
fair enough, thanks for your replies. we will work on fixing our manifests instead :)

